PR #4, opened 2014-02-13
Added Bootstrap-Select

suzidao:
Hi Robb,

Sorry about the repository confusion. I've integrated the Bootstrap-Select plugin from http://silviomoreto.github.io/bootstrap-select/ into Patternfly and adjusted it to match RCUE standards. The plugin handles both single- and multiple-selection. The only problem we may have is the lack of submenus/flyout support, because form elements can't be nested like list elements.

Thanks,
Susanne

Commit comments:

Single-select and multi-select dropdown plugin for Bootstrap. Does not
handle submenus/flyouts.


rhamilto:
Susanne,

Thanks for tackling this.  A few notes:

Within bootstrap-select.less, you've not followed the published coding standards (rules are not alphabetized by selector), nor have you taken advantage of nesting your rules (you've got 33 distinct instances of ".bootstrap-select", which is unnecessary and unwieldy).  Please refactor this with those considerations in mind.  

Please let me know if I can help,
Robb.

rhamilto:
So in digging in to this deeper, it appears you copied the CSS from Bootstrap-select component into a LESS file and modified it an needed?  Is that true?  Ideally what would be done is the CSS from the component (bootstrap-select.css) would be imported into PatternFly, and the bootstrap-select.less file would only contain overrides for the default styles in bootstrap-select.css.  This is consistent with the way all the other components are integrated.  So, to patternfly.less, add:

```
/* Bootstrap-select */
@import (less) "../components/bootstrap-select/bootstrap-select.css";
```

on line 9.

Then simply add your customizations to bootstrap-select.less.

Please also update the title element of bootstrap-select.html (it still reads "Dropdowns").  And include the minified version of the js library (bootstrap-select.min.js) instead of the unminified version (bootstrap-select.js).

rhamilto:
Manually merged and tidied up a few things.

PR #6, opened 2014-02-19
Adding Bootstrap-Combobox

suzidao:
Hi Robb,

I've added the completed combobox plugin here. I've also updated the bower package to include it. There's only one major issue with the plugin we chose: for some reason, it doesn't render a disabled select form element into a disabled custom combobox. Currently, a disabled select form element renders an ENABLED combobox. You can see it in action on the test page. It's likely that we'll be able to alter the javascript so that it will render a disabled combobox, but I wanted to check in with you to see if that's a worthwhile path to take, or if we should just look for an alternative combobox plugin.

Please let us know what you think.

Thanks,
Susanne

Commit Comments:

Updated bower file to include bootstrap-combobox in package.
Further refinement of combobox styles


rhamilto:
Hi, Susanne.

Thank you for submitting this.

I don't think we should invest anymore time in this option.  In addition to the issues you noted, it works nothing like the developer's demo at http://dl.dropboxusercontent.com/u/21368/bootstrap-combobox/index.html.  I will work on identifying and implementing a solution that does work the way we want.

Thanks,
Robb.

P. S.  There is a missing comma on line 20 of bower.json.  I was unable to install Bootstrap-Combobox component until I fixed the error.  Please make sure you test your changes when adding Bower components.  A simple `bower update` will verify what you added is correct.

suzidao:
Hi Robb,

Thanks for catching the comma. I'm still relatively new to GitHub and working with bower packages, so this has been a great learning experience for me.

Thanks for bearing with me,
Susanne

PR #8, opened 2014-03-02
Tables and DataTable plugin styles and configuration

joshuakim:
Hi Robb,

This pull request is for bootstrap's "tables" component and Asset Board's Data Table skinning and configuration using DataTables 1.9.x plugin.

NOTE:
IE Issues - table header gradient using filter overwrites borders, so this looked really odd. So instead, for IE8, 9 flat bg colors are used instead of gradients.

DataTable JS: I’ve included the global default setting for all datatables in patternfly.js. If $.fn.dataTableExt object is detected, then the default configuration is applied.
If for some reason certain features are not needed like the Search box or pagination, etc., we could modify the sDom and remove these features when the .datatable({...}) is initialized.

~ JK ~


rhamilto:
Manually merged.  Fixed a couple of bugs.

joshuakim:
Thanks! I just reviewed your changes.

I forgot we were alphabetizing declarations. And as you've probably
noticed, I left some of the class names as is because it was provided by
the DataTables code like .dataTables_header, table.datatable, etc.

PR #17, opened 2014-05-05
adjust @input-height-base to @font-size-base: 12px

pvoborni:
.input-group-btn height is by 1px larger than input's -> input-groups look ugly.

This should make them equal.


rhamilto:
Thanks, Petr!  In digging in to this bug a big deeper, it's actually the result of an incorrectly computed  @line-height-base.  I've fixed the issue with https://github.com/patternfly/patternfly/releases/tag/v0.2.5

Thanks again!

PR #25, opened 2014-06-25
packaging: add make and spec

gregsheremeta:
make is the core of packaging, may be used by different distributions to
avoid logic duplication.

spec is for rpm based systems.

Signed-off-by: Alon Bar-Lev alon.barlev@gmail.com

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,7 @@ node_modules/
 nodevars.bat
 npm
 npm.cmd
+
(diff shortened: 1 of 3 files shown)


rhamilto:
@gregsheremeta I'm in Westford this week at a UXD all hands.  Can we sync up on this on Fr Jun 27?

gregsheremeta:
Monday would be better.

PR #26, opened 2014-06-25
Added 'main' property to bower

jpkrohling:
This helps Wiredep in injecting the right files into an annotated index.html file.


rhamilto:
Thanks @jpkrohling!

PR #28, opened 2014-07-07
added Changelog

gregsheremeta:


rhamilto:
We decided against this in favor of the log at https://github.com/patternfly/patternfly/releases

PR #33, opened 2014-09-22
More work to enable Sass compatibility.

awood:
Sass is pretty picky about variable declaration order and with overrides so I have to import the variables file first thing in the patternfly.scss file.  I can't import it first thing if any of the values aren't defined so the modifications in this patch change some of the references to variables in external components to just their actual values plus there is some reordering so that variables are defined before they are used.


rhamilto:
Once the dust settles, I'd love to see some documentation on the "rules" of variable declaration.  It was obvious when it was alphabetical.  It's not so obvious now.  :)

PR #43, opened 2015-01-06
Vertical Navigation with corrections

cardosogabriel:


cardosogabriel:
That's great Robb. Did you figure out how to display the two new pages in the nav structure?

rhamilto:
@cardosogabriel, Yeah.  We opted to condense all three options in to the existing tests/tab.html page; users can simply remove headings or icons if they're not needed.

PR #56, opened 2015-04-07
Reordered $panel-primary-text in variables.less for SASS compatibility

skateman:
I'm working on the [sass version](https://github.com/patternfly/patternfly-sass) of Patternfly and I have some issues with the variable order. Reordering the variables in the LESS version is an easier solution than doing this algorithmically with a parser. Also Bootstrap uses variables ordered by declarations first.

### DIFF
--- a/less/variables.less
+++ b/less/variables.less
@@ -161,6 +161,7 @@
 @tooltip-bg:                                                        #434343;
 @tooltip-max-width:                                                 220px;
 // References variable declared in blocks above
+@panel-primary-text:                                                @body-bg;
 @alert-danger-bg:                                                   @body-bg;
 @alert-danger-border:                                               @brand-danger;
 @alert-danger-text:                                                 @gray-dark;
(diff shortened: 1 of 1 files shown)


less/variables.less
```diff
@@ -196,11 +196,11 @@
 @padding-small-vertical:                                            @padding-base-vertical;
 @panel-danger-border:                                               @brand-danger;
 @panel-danger-heading-bg:                                           @brand-danger;
+@panel-primary-text:                                                @body-bg;
```
rhamilto: @skateman, Can you move this variable up into the section that precedes the current section (move it to line 149)?  That way it maintains the pseudo organization we've already got in the file.

skateman: sure! :)

less/variables.less
```diff
@@ -146,6 +146,7 @@
 @pagination-border:                                                 #bbbbbb;
 @pagination-hover-bg:                                               #ededed;
 @panel-inner-border:                                                #cecdcd;
+@panel-primary-text:                                                @body-bg;
```
rhamilto: D'oh.  My bad.  This line should actually go to line 164 preceded by  "// References variable declared in block above".

skateman: @rhamilto done :)

PR #78, opened 2015-06-18
project restructure into src and adding jshint

erundle:
I added jshinting and fixed any issues that came up in the report. I also pulled the js files from dist into src and moved them each into their own file. On watch or build these are concatinated into the patternfly.js file in dist. I also moved the less files under src/. I did not change the dist directory paths at all. 

Please take a look at the src changes and let me know if any of these are going to be an issue


erundle:
Will create a new pull request with a subset of these changes.

PR #79, opened 2015-06-19
JSHint added, code style cleanup

akurdyukov:
I added some code style checks and formatted code a bit. Main problem was in `sidebar` variable that failed webpack to build a project with patternfly included.


rhamilto:
Thanks, @akurdyukov!  This is a great addition!  Unfortunately, this change is "breaking" the js transformation of data tables.  See https://rawgit.com/akurdyukov/patternfly/master/tests/datatables.html vs. https://rawgit.com/patternfly/patternfly/master/tests/datatables.html  Mind taking a look?

akurdyukov:
Sure. Bug is fixed now.

rhamilto:
And one more, @akurdyukov.  Because sidebar() is no longer global, calling it via the data table call back [1] on https://rawgit.com/akurdyukov/patternfly/master/tests/tab.html no longer works.  

[1] https://github.com/akurdyukov/patternfly/blob/master/tests/tab.html#L588

akurdyukov:
Alright, I fixed that, but now sidebar looks like jQuery plugin. It's the only effective global scope I have when loading with webpack. I really do not like putting anything into 'window'. What do you think?

rhamilto:
Thank, @akurdyukov.  That seems reasonable to me.  

@erundle, @jeff-phillips-18, @dtaylor113, opinions?

erundle:
Looks good to me!!!

rhamilto:
Well, dang.  One more bug.  "PatternFly" [1] is no longer in the global scope.  Sorry I didn't catch that earlier.

[1] https://github.com/akurdyukov/patternfly/blob/master/tests/popovers.html#L93

akurdyukov:
Updated :)

rhamilto:
Thanks, @akurdyukov!  I made one additional commit to your PR as the test files are actually generated using Jekyll and the corresponding source test files (/tests-src/*.html) needed to be updated as well.

PR #82, opened 2015-07-20
change default text color to black

sorki:
Makes sites much easier to read due to increased
contrast between backround and foreground.

Signed-off-by: Richard Marko rmarko@fedoraproject.org

### DIFF
--- a/components/bootstrap/less/variables.less
+++ b/components/bootstrap/less/variables.less
@@ -28,7 +28,7 @@
 //** Background color for `<body>`.
 @body-bg:               #fff;
 //** Global text color on `<body>`.
-@text-color:            @gray-dark;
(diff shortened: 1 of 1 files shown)


rhamilto:
Thanks, @sorki.  There are varying opinions on this topic, but it seems the general consensus is to not use pure black on a white background, but to use a dark shade of gray as pure black on a white background creates jagged-looking text due to anti-aliasing which is actually worse for readability.  For example, the body copy on http://www.nytimes.com/2015/07/20/world/middleeast/murder-at-sea-captured-on-video-but-killers-go-free.html?hp&action=click&pgtype=Homepage&module=span-ab-top-region&region=top-news&WT.nav=top-news is set in the same color (#333) we're using in PatternFly.

PR #87, opened 2015-08-04
Explicitly set bootstrap-select's variable overrides in our code

skateman:
It's better to understand what is overridden and the CSS file can be used instead of LESS.
It also fixes an issue in patternfly-sass, where only the CSS can be used.

### DIFF
--- a/less/bootstrap-datepicker.less
+++ b/less/bootstrap-datepicker.less
@@ -64,5 +64,8 @@
   }
   .input-group-addon {
     padding: @padding-base-vertical @padding-base-horizontal;
+    line-height: @line-height-base;
+    background-color: @input-group-addon-bg;
+    border-color: @input-group-addon-border-color;
   }
 }


erundle:
I agree it's better for us to explicitly override values and I don't see any issues with your request but I am slightly confused to what issue this fixes.

If you look at the date picker code on github they have 14 properties set on input-group-addon and this still seems to be a subset of them: https://github.com/eternicode/bootstrap-datepicker/blob/1.4/less/datepicker3.less

Can you elaborate a bit more on the patternfly-sass CSS issue so we can prevent these types of issues in the future.

skateman:
Currently the LESS-SASS conversion pipeline doesn't allows me to convert files which were included from bower dependencies. That's why I'm including datepicker into patternfly-sass as a CSS file. I have specified only those variables in the `bootstrap-datepicker.less` which are modified by `variables.less`, because the CSS file has those values "hardcoded".

erundle:
Makes sense thanks.

less/bootstrap-datepicker.less
```diff
@@ -64,5 +64,8 @@
   }
   .input-group-addon {
     padding: @padding-base-vertical @padding-base-horizontal;
+    line-height: @line-height-base;
```
rhamilto, line 67: As a reminder, https://github.com/patternfly/patternfly#coding-style specifies that we alphabetize properties by declaration.

PR #88, opened 2015-08-05
remove bootstrap-select  js initialization

wojtask9:
bootstrap-select >1.6.3 do this automatically.


wojtask9:
done

rhamilto:
Good catch, @wojtask9!  Thanks!

PR #92, opened 2015-08-12
Adding basic aggregate status card

rhamilto:


rhamilto:
@andresgalante or @erundle review and merge, please?

less/variables.less
```diff
@@ -9,6 +9,7 @@
 @btn-default-bg-img-stop:                                           #ededed;
 @btn-xs-font-weight:                                                400;
 @card-pf-accented-border-top-color:                                 #39a5dc;
```
andresgalante, line 11: Isn't this the same blue as brand-primary? Kyle is using way too many colors :)
We could have a colors section on our variables file and reference them, so it is less confusing and we avoid repetition.

andresgalante:
:+1: looks fine to me

less/variables.less
rhamilto, line 11: That's a good point, @andresgalante.  As we've been growing organically we've not gone a great job of keeping inventory of the colors.  Add a card to https://trello.com/b/Hz3Nmwk4/patternfly-reference-implementation?

andresgalante, line 11: @rhamilto I've added to the backlog. I'll send a PR with this as soon as I find a time gap to do it.

PR #94, opened 2015-08-14
cleans up data tables reference html

andresgalante:
@rhamilto I've removed the extra classes on the data tables reference implementation HTML.
No more classes on td `<td class="center">X</td>` or tr `<tr class="gradeA">`


rhamilto:
Thanks, @andresgalante!  I updated https://www.patternfly.org/widgets/#data-tables with this change.

PR #95, opened 2015-08-14
Adding support for bootstrap-datepickers with readonly inputs

rhamilto:
Updates Boostrap-datepicker test page to include examples with readonly inputs.  Adds styling to make
.form-inputs.bootstrap-datepicker[readonly] inputs look like regular .form-inputs.

The recommended PatternFly use case is to not allow manual input on Bootstrap-datepicker inputs to avoid invalid date entry.


rhamilto:
Since I can't assign more than one person to the PR, I'll just tag you both in a comment.  @andresgalante @erundle

rhamilto:
Eek.  Found a problem.  Will resubmit when resolved.

rhamilto:
I had to add the clear button to the picker so that date could be removed via the datepicker widget on readonly inputs.  Otherwise there was no way for a user to remove a date once one was chosen on readonly inputs.  See https://rawgit.com/rhamilto/patternfly/datepicker/tests/bootstrap-datepicker.html for a live example.

less/bootstrap-datepicker.less
```diff
@@ -2,9 +2,29 @@
...
+    }
+  }
+}
+
 .datepicker {
   border-radius: @border-radius-base;
-  .datepicker-switch, tfoot .today {
+  .datepicker-switch, tfoot .clear, tfoot .today {
     font-size: @font-size-large;
     font-weight: 500;
```
andresgalante, line 29: Todo on variables when we do the color guide we should also do a font weight guide. I'll add it to the backlog.

andresgalante:
@rhamilto I've tried to teak it to use `type="date"` instead of `text` but the plugin we are using does not agree with me. 
:+1: to merge this. Nice catch with the clear btn. I wouldn't have think about it.

tests-src/bootstrap-datepicker.html
```diff
@@ -6,24 +6,44 @@
...
+        <input type="text" class="form-control bootstrap-datepicker" name="start">
         <span class="input-group-addon">to</span>
-        <input type="text" class="form-control" name="end">
+        <input type="text" class="form-control bootstrap-datepicker" name="end">
       </div>
       <script>
         // Initialize Boostrap-Datepicker
         $('.bootstrap-datepicker, .input-group.date, .bootstrap-datepicker-inline').datepicker({
           autoclose: true,
+          clearBtn: true,
```
erundle: Do we need to run this by someone from a design perspective? Another option might be just having an "X" button next to the calendar icon that clears the input. Otherwise it LGTM

rhamilto: Good call, @erundle.  @LHinson and @serenamarie125 are reviewing.

less/bootstrap-datepicker.less
rhamilto, line 29: Another great suggestion, @andresgalante!

rhamilto:
After discussion with @serenamarie125 and @LHinson, we've decided to remove the clear button from the recommended Bootstrap-datepicker examples until further testing can be completed.

rhamilto:
@andresgalante or @erundle please merge.

andresgalante:
@rhamilto I'll merge it now.
Just a design suggestion: @LHinson @serenamarie125 For the search box we are using a "x" on the input to clear the form. We can use the same pattern here: https://www.patternfly.org/widgets/#search

PR #96, opened 2015-08-19
Adding note regarding .row-cards-pf usage on nested .row

rhamilto:
@andresgalante got tripped up by the lack of documentation regarding .row-cards-pf usage when nesting .row within .row.row-cards-pf.  This is a stop-gap measure for documenting correct .row-cards-pf usage.  Does it make sense?


andresgalante:
:+1: This is super useful until we have a proper documentation this should do the trick. @erundle if you agree please merge it.

PR #102, opened 2015-08-25
Alternate Header/Footer/Sidebar Layout

erundle:


less/layouts.less
```diff
@@ -0,0 +1,47 @@
+.container-pf-scroll-body {
+  background: @container-nav-pf-vertical-alt-colored-bg;
+  overflow: hidden;
+}
+
+.container-pf-sticky-footer > .container-pf-scroll {
```
rhamilto: Would be good to nest this rule on line 13 with selector '.container-pf-sticky-footer > & {'

less/layouts.less
```diff
@@ -0,0 +1,47 @@
...
+.container-pf-scroll-content-area {
+  padding: 20px 0;
+}
+
+.container-pf-nav-pf-vertical-alt {
+  margin-left: (@nav-pf-vertical-alt-width + 5px);
+  transition: @flyout-transition-pf;
+}
+
+.container-pf-nav-pf-vertical-alt.collapsed-nav {
```
rhamilto: Would be good to nest this rule on line 22 with selector '&.collapsed-nav {'

less/layouts.less
```diff
@@ -0,0 +1,47 @@
...
+  margin-left: (@nav-pf-vertical-alt-collapsed-width + 5px);
+}
+
+.container-pf-nav-pf-vertical-alt.hidden-nav {
```
rhamilto: Would be good to nest this rule on line 25 with selector '&.hidden-nav {'

less/nav-vertical-alt.less
```diff
@@ -0,0 +1,137 @@
+.container-pf-sticky-footer > .nav-pf-vertical-alt {
```
rhamilto: Would be good to nest this rule using selector '.container-pf-sticky-footer > &'

less/nav-vertical-alt.less
```diff
@@ -0,0 +1,137 @@
...
+  float: left;
+  height: @container-nav-pf-vertical-alt-height-calc;
+  overflow-x: hidden;
+  overflow-y: auto;
+  transition: @flyout-transition-pf;
+  width: @nav-pf-vertical-alt-width;
+  z-index: 1000;
+
+  .list-group {
+    margin-bottom: 0px;
```
rhamilto: No need to specify a unit.  Zero is always zero.  :)

less/nav-vertical-alt.less
```diff
@@ -0,0 +1,137 @@
...
+      &:focus {
+        color: @gray-dark;
+        text-decoration: none;
+      }
+
+      &:hover {
+        color: @nav-pf-vertical-alt-active-color;
+        text-decoration: none;
+      }
+	}
```
rhamilto: This file contains tabs in addition to spaces.  The first tab is here.  Please only use spaces.

less/nav-vertical-alt.less
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  .list-group-item {
+    padding: 0;
+
+    a {
+      color: @gray-dark;
+      display: block;
+      font-size: 14px;
+      height: 63px;
```
rhamilto: An explicit height feels really hack-y.  Can the contents not determine the height?

less/nav-vertical-alt.less
```diff
@@ -0,0 +1,137 @@
...
+      padding: 17px 20px 17px 25px;
```
rhamilto: Variables?

less/nav-vertical-alt.less
```diff
@@ -0,0 +1,137 @@
...
+	.badge {
+	  background: @gray-dark;
+	  border: 1px solid @nav-pf-vertical-alt-background-color;
+	  border-radius: 3px;
+	  color: @nav-pf-vertical-alt-badge-color;
+	  font-weight: 700;
+	  float: right;
+	  font-size: 9px;
+	  min-width: 20px;
+	  padding: 5px 5px;
```
rhamilto: Could just be '5px'.

less/nav-vertical-alt.less
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	.badge {
+	  background: @gray-dark;
+	  border: 1px solid @nav-pf-vertical-alt-background-color;
+	  border-radius: 3px;
+	  color: @nav-pf-vertical-alt-badge-color;
+	  font-weight: 700;
+	  float: right;
+	  font-size: 9px;
```
rhamilto: Use variable expression, please.  @font-size-base - 3 // 9px

less/nav-vertical-alt.less
erundle: From what I recall I put in the explicit height to handle the wrapping text case and to force an ellipsis instead but I will double check.

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
+.navbar-pf-header-alt {
+  background: @navbar-pf-header-alt-bg-color url('../img/nav-header-banner.svg') top left no-repeat;
+  border: none;
+  border-radius: 0px;
```
rhamilto: No need for a unit.

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
...
+  border-top: 2px solid @navbar-pf-border-color;
+  height: @navbar-pf-header-alt-height;
+  margin-bottom: 0px;
+  padding-top: 3px;
```
rhamilto: Please eliminate padding on this element and let the contents determine the height.

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
+.navbar-pf-header-alt {
+  background: @navbar-pf-header-alt-bg-color url('../img/nav-header-banner.svg') top left no-repeat;
+  border: none;
+  border-radius: 0px;
+  border-top: 2px solid @navbar-pf-border-color;
+  height: @navbar-pf-header-alt-height;
+  margin-bottom: 0px;
```
rhamilto: No need for a unit.

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
+.navbar-pf-header-alt {
+  background: @navbar-pf-header-alt-bg-color url('../img/nav-header-banner.svg') top left no-repeat;
+  border: none;
+  border-radius: 0px;
+  border-top: 2px solid @navbar-pf-border-color;
+  height: @navbar-pf-header-alt-height;
```
rhamilto, line 6: Contents should determine the height of the bar since the images are going to be different sizes.

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
...
+        margin-top: -5px;
+        z-index: 2000;
+      }
+      
+      .notification-icon-count {
+        background-color: @notification-badge-color;
+        border-radius: 50%;
+        color: @navbar-pf-header-alt-active-color;
+        cursor: pointer;
+        font-size: 9px;
```
rhamilto: Please use a variable.

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
...
+        font-weight: 700;
+        height: 15px;
+        line-height: 15px;
+        padding-left: 0px;
```
rhamilto: No need for a unit.

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
...
+        .caret, .fa, .pf-icon {
+          color: @navbar-pf-header-alt-active-color;
+        }
+      }
+    }
+  }
+
+  .navbar-brand {
+    line-height: @navbar-pf-header-alt-brand-line-height;
+    padding: 8px 20px 15px 0px;
```
rhamilto: Should use variables so that they can be easily overridden since the images are going to change on a per instance basis?

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
...
+
+    .navbar-brand-name {
+      display: inline;
+      height: @navbar-pf-header-alt-brand-name-height;
+    }
+
+    .navbar-brand-icon {
+      display: inline;
+      height: @navbar-pf-header-alt-brand-icon-height;
```
rhamilto: Since this image is going to change on a per instance basis, we don't want CSS sizing it.

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+
+  .navbar-brand {
+    line-height: @navbar-pf-header-alt-brand-line-height;
+    padding: 8px 20px 15px 0px;
+
+    .navbar-brand-name {
+      display: inline;
+      height: @navbar-pf-header-alt-brand-name-height;
```
rhamilto: Since this image is going to change on a per instance basis, we don't want CSS sizing it.

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
...
+
+    .navbar-brand-icon {
+      display: inline;
+      height: @navbar-pf-header-alt-brand-icon-height;
+      margin-right: 15px;
+    }
+  }
+
+  //This logo is too long. Need to hide it
+  @media (max-width: 460px) {
```
rhamilto, line 103: This breakpoint is based on what exactly?

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
...
+    .navbar-brand .navbar-brand-name{
```
rhamilto: Need a space between 'e' and '{'.

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
...
+      display: none;
+    }
+  }
+
+  .navbar-right {
+    margin-right: 0px;
```
rhamilto: No need for a unit.

less/navbar-header-alt.less
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  .navbar-toggle {
+    float: left;
+    margin-left: 14px;
+    margin-right: 20px;
+    display: block;
```
rhamilto: Alpha.

less/navbar-header-alt.less
erundle: Would you like the sizing inline in the HTML? It needs to be sized somewhere.

less/variables.less
```diff
@@ -15,13 +15,19 @@
 @card-pf-border-top-color:                                          transparent;
 @card-pf-container-bg-color:                                        #f5f5f5;
 @card-pf-footer-bg-color:                                           #fafafa;
+@container-nav-pf-vertical-alt-colored-bg: 							#f5f5f5;
```
rhamilto: Your additions use tabs for spacing instead of spaces.  Please only use spaces.

less/navbar-header-alt.less
erundle, line 6: We need to set an explicit height since we need to calculate the remaining container scroll size. If someone modifies this in their product they will need to modify the container height as well.

less/variables.less
```diff
@@ -96,6 +102,27 @@
 // References variable declared in block above
 @navbar-pf-navbar-primary-context-active-border-right-color:        @navbar-pf-navbar-primary-context-active-border-color;
 
+// Top Nav Bar Styles
+@navbar-pf-header-alt-bg-color:										#030303;
```
rhamilto: Please fix the alphabetization on all these new navbar and nav vars.  

And all these vars should be integrated in the first section of vars (around line 50).  The lower blocks reference a variable(s) defined above.  This matters for the less --> sass conversion script used in patternfly-sass.

less/variables.less
```diff
@@ -96,6 +102,27 @@
...
+@nav-pf-vertical-alt-color:         								#464952;
+@nav-pf-vertical-alt-background-color: 								#fff;
+@nav-pf-vertical-alt-active-color:     								#39a5dc;
+@nav-pf-vertical-alt-width:         								235px;
+@nav-pf-vertical-alt-collapsed-width:         						70px;
+@nav-pf-vertical-alt-badge-color: 						  			#fff;
+@nav-pf-vertical-alt-separator:										#f3f3f3;
+
+// Notification Styles
+@notification-badge-color: 						  					#cc0000;
```
rhamilto: should use @brand-danger as value and move to roughly line 252 (block of variables at the end of the file).

tests-src/_includes/nav-vertical-alt.html
```diff
@@ -0,0 +1,75 @@
+<div class="nav-pf-vertical-alt">
+  <ul class="list-group">
+    <li class="list-group-item">
+      <a href="#">
+      <i class="fa fa-dashboard" title="Dashboard" data-toggle="tooltip" data-placement="right"></i>
```
rhamilto: &lt;i&gt; is the "standard" for icons, but we've been using &lt;span&gt; elsehwhere in PatternFly.  Would be good to be consistent.  Please make that change across all your markup.

tests-src/_includes/navbar-header-alt.html
```diff
@@ -0,0 +1,59 @@
+<header class="navbar navbar-pf-header-alt" role="navigation">
+  <div class="navbar-header">
+    <button type="button" class="navbar-toggle">
+    <span class="sr-only">Toggle navigation</span>
```
rhamilto: Indent this and the next three lines.

tests-src/nav-vertical-alt-header.html
```diff
@@ -0,0 +1,25 @@
+---
+categories: [Layouts]
+css-extra: false
+layout: container-scroll-layout
+resource: true
+title: Alternate Header w&sol; Left Hand Navigation
```
rhamilto: &sol;?

tests-src/nav-vertical-alt-header-footer.html
```diff
@@ -0,0 +1,28 @@
+---
+categories: [Layouts]
+css-extra: false
+layout: container-scroll-layout
+resource: true
+title: Alternate Header w/ Left Hand Navigation &amp; Footer
+weight: 8
```
rhamilto: Change value to '10'

tests-src/nav-vertical-alt-header.html
```diff
@@ -0,0 +1,25 @@
...
+weight: 8
```
rhamilto: Change value to 9.

tests-src/nav-vertical-alt-header.html
rhamilto: Title is too long as it gets output in the existing layout navbar.  Maybe we need to make a new layouts section and only put these two examples in?  See https://rawgit.com/erundle/patternfly/master/tests/index.html and https://rawgit.com/erundle/patternfly/master/tests/basic.html

tests-src/nav-vertical-alt-header-footer.html
```diff
@@ -0,0 +1,28 @@
+---
+categories: [Layouts]
+css-extra: false
+layout: container-scroll-layout
+resource: true
+title: Alternate Header w/ Left Hand Navigation &amp; Footer
```
rhamilto: Title is too long as it gets output in the existing layout navbar. Maybe we need to make a new layouts section and only put these two examples in? See https://rawgit.com/erundle/patternfly/master/tests/index.html and https://rawgit.com/erundle/patternfly/master/tests/basic.html

less/navbar-header-alt.less
rhamilto, line 6: So wouldn't it be better to do all that in javascript so it's robust and independent of the display?

erundle, line 6: The issue is that forces us to load some kind of global function to do that calculation which isn't ideal either. Angular applications also shouldn't look at the dom to inspect/set height so I personally believe this approach is the better of two evils.

dist/js/patternfly.js
```diff
@@ -212,3 +212,117 @@
     });
   }
 }(jQuery));
+
+// Util: PatternFly Collapable Left Hand Navigation
```
rhamilto: Should this go in its own file since it's specific to the new layout?

erundle: We have so little custom JS putting this in its own file seems like overkill. If we take that approach the CSS should go in its own file as well. As of right now this JS will not be used by any application since the angular version of this rewrites this logic entirely

less/navbar-header-alt.less
erundle, line 103: Nothing besides the fact that we have too much spacing in the design and need to break before the bootstrap default.

tests-src/_includes/nav-vertical-alt.html
erundle: Sorry, can you please be more specific here. I don't know what is inconsistent with PF.

tests-src/nav-vertical-alt-header.html
erundle: &sol; = "/" I used the "/" since the title was long and I figured you would want me to escape it.

tests-src/_includes/nav-vertical-alt.html
rhamilto: Sorry.  Forgot to escape my markup.  Should make sense now.  :)

tests-src/nav-vertical-alt-header.html
rhamilto: Ok.  Was confused because you didn't escape the "/" in the other page.  These titles should probably be truncated a bit anyhow.

less/navbar-header-alt.less
rhamilto: Probably need to simply make the image the correct size.  I realize this means multiple images, but didn't consider the issue adding CSS sizing presents.  Sorry about that.

less/nav-vertical-alt.less
erundle: Looked at this again. I set the explicit height since I am fading in/out the text associated with the menu item on expand collapse. All I am changing is opacity to get the desired effect. This causes the element to still have height even if it isn't visible to the user so the set height makes sure the link doesn't grow. Without this it causes the expand/collapse to be jumpy even if I try setting the height/width to zero and also position. You can't animate display without keyframes/JS as far as I am aware. For now I am going to suggest pulling this out the a variable and we can address if it becomes an issue.

tests-src/_includes/nav-vertical-alt.html
erundle: I can appreciate the consistency concern but shouldn't we be coding new features towards industry standards? This is pretty much the norm and anyone who uses font-icons tags within their application are going to use "i" instead of "span". I would think it would make sense for us to start using "i" in new features and in a future release we can change existing PF markup to use "i" instead of "span". Thoughts?

rhamilto: I am firmly in the camp of it's better to be consistently wrong than inconsistently right.  I'm ok with switching all icons to &lt;i&gt;, but not ok with shipping both.

erundle: OK, I changed my markup from an "i" to a "span". We can discuss making this change throughout PF when we are on a less tight deadline.

erundle:
I added two new images to address the image height issue. I also created a new layout section on the test page for these pages since that seemed to be more appropriate. I believe I addressed everything that I could so please take a look at let me know if you have any concerns. For now I left the reviews as different commits but I can squash them later if you would like.

erundle:
@rhamilto can you take a look at this again when you get a chance and let me know if this is good to go?

rhamilto:
@erundle Sure thing.  I also want @andresgalante to review as well, so we'll need to wait on this until next week I imagine.

less/nav-vertical-alt.less
```diff
@@ -0,0 +1,137 @@
...
+    &.active {
+      background-color: @nav-pf-vertical-alt-background-color;
+      border-color: @list-group-border;
+
+      a {
+        color: @nav-pf-vertical-alt-active-color;
+      }
+    }
+
+    &.active:before {
```
rhamilto: Could be nested on line 48 like:

```
      &:before {
        background: @nav-pf-vertical-alt-active-color;
        content: " ";
        display: block;
        height: 100%;
        left: 0;
        position: absolute;
        top: 0;
        width: 5px;
      }
```

less/nav-vertical-alt.less
```diff
@@ -0,0 +1,137 @@
...
+      min-width: 20px;
+      padding: 5px;
+      position: absolute;
+      right: 15px;
+      text-align: center;
+      top: 21px;
+      transition: @flyout-transition-pf;
+    }
+
+    .badge.notifications {
```
rhamilto: Could be nested on line 79 like:

```

      &.notifications {
        background: @notification-badge-color;
      }
```

tests-src/_includes/nav-vertical-alt.html
```diff
@@ -0,0 +1,69 @@
...
+      <a href="#">
+        <span class="fa fa-dashboard" title="Dashboard" data-toggle="tooltip" data-placement="right"></span>
+        <span class="list-group-item-value">Dashboard</span>
+      </a>
+    </li>
+    <li class="list-group-item">
+      <a href="#">
+        <span class="fa fa-file-o" title="My Services" data-toggle="tooltip" data-placement="right"></span>
+        <span class="list-group-item-value">My Services</span>
+        <span ng-if="item.count" class="badge">12</span>
```
rhamilto: Please remove all ng-if attributes.

tests-src/_includes/navbar-header-alt.html
```diff
@@ -0,0 +1,59 @@
...
+  <!-- Collect the nav links, forms, and other content for toggling -->
+  <nav class="collapse navbar-collapse">
+    <ul class="nav navbar-nav">
+      <li><a href="#" target="_blank" class="action"><span title="Open" class="pf-icon pficon-export fa-rotate-270"></span></a></li>
+    </ul>
+    <ul class="nav navbar-nav navbar-right">
+      <li class="dropdown">
+        <a class="dropdown-toggle action" id="notifications" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
+          <span title="Notifications" class="fa pficon-flag"></span>
+          <span class="notification-icon-count">2</span>
```
andresgalante: Can we use a bootstrap "badge" instead of creating a new component?

tests-src/_layouts/container-scroll-layout.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<!--[if IE 9]><html class="ie9"><![endif]-->
+<!--[if gt IE 9]><!-->
+<html class="height-100">
```
andresgalante: Maybe this is just nonsense, but can we choose a better name for this class? or add a -pf- for namespace.

erundle: changed to height-pf-100

tests-src/_includes/nav-vertical-alt.html
erundle: done

less/nav-vertical-alt.less
erundle: done

less/nav-vertical-alt.less
erundle: done

tests-src/_includes/navbar-header-alt.html
erundle: Changed to badge

erundle:
I squashed the code review commits. Is this good to go?

andresgalante:
@erundle I know we've talk about this a lot, but I am still not comfortable with the way we handle the scroll bar when there is a fixed header out of the box solution from bootstrap. The fact that the scroll bar goes under the navbar is a bug on our POC that doesn't exists on bootstrap http://getbootstrap.com/examples/navbar-fixed-top/

I also haven't been able to test any of this on IE9. Do you guys have a windows box to make sure it works?

erundle:
Yes this is a screenshot using Windows 8.1 and IE11. I also changed developer mode to IE9 and I didn't see any issues. This is the implementation that the designers and unfortunately we can't just use out of box bootstrap to do this. I have built similar implementations to this many times and I am comfortable that this will not be an issue
screenshot 1

rhamilto:
I agree with Andres.  I am not comfortable with not only the way the scroll bar is implemented, but also the fragility of the layout and the lack of modularity (I checked out the branch and see many areas for improvement).  I have begun a refactor on the code to make it more robust as well as improve modularity.  I will submit a PR once it's ready for comment.  I imagine it will take me a couple more days to complete.  This work will _not_ impact the v2.2.0 release deadline.

The layout is, in fact, broken in IE9 (see attached screenshot).  The problem is "height-pf-100" is missing on line 2.  Note:  Microsoft no longer recommends using developer tools to emulate older versions of IE.

screen shot 2015-08-31 at 10 38 33 am

erundle:
OK I added the height-pf-100 to the IE9 conditional and that addressed the issue. Do we have any idea it that is still needed or can the conditional itself be removed? I am not using it so I don't know if it is a legacy PF thing or not.

I am going to be honest that I am not comfortable with a refactor at this point. This work has been ongoing for 3+ weeks and has gone through multiple code reviews and has been refactored multiple times based on your feedback.

erundle:
Besides the scrolling can you please articulate what you don't like/what doesn't work for you?

rhamilto:
Yes, the conditional comment is needed.

I understand your concern.  But given that neither @andresgalante nor I are comfortable with the current state of the code, a refactor is required.  And, at this point, it's easier for me to refactor directly than try to talk through all the areas I'd like to see improved given the pending v2.2.0 deadline.  @serenamarie125 and @kybaker are aware of this plan and support it.

erundle:
What do you want? JS height calculation instead of CSS? What is insufficient?

jeff-phillips-18:
@andresgalante What scroll bar issue are you speaking of? In the screen shots here I don't see the scrollbar going under the navbar. Does that issue still exist?

I know bootstrap has the scrollbar on the outside but that seems like the result of limitations to me. Theoretically, the scrollbar should only be drawn in the area that actually scrolls which is how the screen shots here appear.

andresgalante:
@jeff-phillips-18 When @erundle shown us the sticky header with bootstrap class `navbar-fixed-top` there was a bug with the scroll bar going underneath the navigation bar. That was a bug on the implementation since on bootstrap the scroll goes over the navbar.

We should keep components and modular as possible, and since we are using bootstrap as based, we should try to take advantage of as many bootstrap components as we can. To build a new one means  not only maintaining it, but will also make it harder to upgrade to new bootstrap versions.

What you mention is right, theoretically it is nicer to have the scroll on the part thats scrolls, but I think that the cost of implementing that is not worth it.

As a rule, I think that we should always stick to bootstrap when the component already exists on bootstrap. And only build new components when it doesn't exists on bootstrap. If we don't follow this rule and we starting changing a thing here and there, we would be heading to a big mess.

erundle:
When we showed the options to @kybaker he said "Do whatever you need to do to make the scrolling happen within the content area" which is the current behavior of implementation. That was the design direction that I was given so I went ahead and made that happen. I don't really see the overhead that is associated with this implementation as being an issue since it is just a couple of extra classes. Yes this is no longer using default bootstrap but we aren't bootstrap.

andresgalante:
@kybaker is not a developer and although he is big and loud he can be reasoning. It is our job to explain what the consequences are of reproducing the behaviour he wants.
And that is exactly what I am against. A "couple of extra classes" for a behaviour that already exists on bootstrap is something that I think we should never do. If we do that with every component it'll lead to a nightmare. And this would be the first step.
But again, I am not part of the patternfly team, so whatever you decide will be what people like me will consume.

rhamilto:
I suggest we talk about this at our 3pm today.  I agree with Andres; modularity is the key, and we should build PatternFly in a way that ensures both simplicity and future flexibility.  Today we're saying we want the alt layout to have a fixed navbar, side nav, and footer, but what happens in X months if we change our minds?  We can build PatternFly in a way that ensures it's future-proof _and_ meets our needs today (which is what I'm working on).  This may mean we trade off the exact location of where the scrollbar appears, but we have a robust solution that can adapt and change as designs do.  I strongly believe ignoring the precedent set by Bootstrap is short sided.

rhamilto:
Replaced by https://github.com/patternfly/patternfly/pull/109

PR #107, opened 2015-09-02
Toolbar

andresgalante:
It adds a tool with filter and view selection options.
It kinda sucks install screens.  @serenamarie125 and @kybaker are aware of it and will come up with a solution to adapt this component to mobile environment.


less/toolbar.less
```diff
@@ -0,0 +1,61 @@
+//
+// Toolbar (Filter and view options)
+// --------------------------------------------------
+.toolbar-pf{
```
rhamilto: Please add a space before each "{"

less/toolbar.less
```diff
@@ -0,0 +1,61 @@
...
+    cursor: default;
+  }
+  a{
+    color: @gray-darker;
+    &:hover{
+      color: @link-color;
+    }
+  }
+}
+.toolbar-pf-results{
```
rhamilto: This rule should come before ".toolbar-pf-view-selector" on line 17.

less/toolbar.less
```diff
@@ -0,0 +1,61 @@
...
+  border-top: 1px solid @sidebar-pf-border-color;
+  margin-top: @grid-gutter-width/4;
+  ul, h5, p{
```
rhamilto: Alpha fix:  put "h5, p" first

tests-src/toolbar.html
```diff
@@ -0,0 +1,155 @@
+---
+alert-message: ', <a href="http://getbootstrap.com" class="alert-link">http://getbootstrap.com</a>, and <a href="http://silviomoreto.github.io/bootstrap-select/">http://silviomoreto.github.io/bootstrap-select/</a>.'
```
rhamilto: You can remove all references to Boostrap select since there aren't any on the page.

tests-src/toolbar.html
```diff
@@ -0,0 +1,155 @@
...
+categories: [Widgets]
+layout: page
+title: Toolbar
+resource: true
+url-js-extra: '!URL_COMPONENTS!bootstrap-select/dist/js/bootstrap-select.min.js'
```
rhamilto: Not needed.

tests-src/toolbar.html
```diff
@@ -0,0 +1,155 @@
...
+---
+      <h2>Toolbar</h2> 
+ 
+      <nav class="navbar navbar-default navbar-pf" role="navigation">
```
rhamilto: role="navigation" is not needed.

rhamilto:
Great start, @andresgalante!  Fix the things I commented on, rebase and squash?

less/toolbar.less
rhamilto: A blank line here would be ok.  :)

PR #109, opened 2015-09-03
Alt layout with fixed header, sidebar, and optional footer

rhamilto:


rhamilto:
@andresgalante, you should probably review and merge this one since @erundle and I both wrote it.  Sorry, it's a big one.

less/nav-vertical-alt.less
```diff
@@ -0,0 +1,167 @@
...
+      width: 18px;
+    }
+
+    .list-group-item-value {
+      display: inline-block;
+      line-height: 30px;
+      opacity: 1;
+      overflow: hidden;
+      text-overflow: ellipsis;
+
```
andresgalante: @rhamilto we could remove this space

PR #114, opened 2015-09-11
Changes to align with new code guide

rhamilto:

I forked http://codeguide.co/ and revised the section on declaration order to match our existing standard (alpha) and published to http://patternfly.github.io/patternfly-code-guide/ (with the corresponding repo at https://github.com/patternfly/patternfly-code-guide).  In reviewing the code guide, I discovered there were some areas of misalignment within PatternFly that this PR covers.


erundle:
Having a code guide is good but honestly I believe most of this is preference and shouldn't be enforced. Spaces vs Tabs don't impact the end user quality of the CSS neither does variable order or less nesting. It is also a hindrance to other people contributing. Granted I know this is what you have been using but these are all your preferences and other people may prefer other ways to write code. I personally prefer to group my variables in logical groupings (ex: all position styles in close proximity since I believe it is easier to grasp what is going on). The same goes for less nesting sometime it is easier to read things one way vs another. I am OK with a code guide I would like to just decide as a team what we feel is important vs. publishing something.

rhamilto:
@dtaylor113 and @jeff-phillips-18, seeing that we appear to have reached consensus on this, please review and merge.

andresgalante:
:+1:

PR #124, opened 2015-09-28
Adding Bootstrap Switch

rhamilto:
New widget for binary form input (e.g., on/off).  


tests-src/bootstrap-switch.html
```diff
@@ -0,0 +1,227 @@
+---
+alert-message: ', <a href="http://getbootstrap.com" class="alert-link">http://getbootstrap.com</a>, and <a href="http://www.bootstrap-switch.org/">http://www.bootstrap-switch.org/</a>.'
+categories: [Widgets]
+layout: page
+title: Bootstrap Switch
```
erundle, line 5: Should we call this toggle switch?

andresgalante:
Although it is a toggle switch, we are using a component name "Bootstrap switch". I think it is a good idea to keep the original name, so it's easy to understand where we pick it from.
Plus this is the way we've been doing things so far, for example date pickers, data tables, etc.

tests/bootstrap-switch.js
```diff
@@ -0,0 +1,61 @@
+(function() {
+  var $confirm;
+
```
erundle, line 3: So I know you just copied this from the example page for the project itself but do want to create any JS in our dist code base to support some of these actions?

rhamilto, line 3: I don't think there is a need; these actions are just for the purposes of demo-ing the widget.

PR #125, opened 2015-09-28
Adding Bootstrap button.js examples to buttons test page

rhamilto:
The Bootstrap button.js examples weren't part of the test page, so this adds them since we'll be promoting their use in the near future.

### DIFF
--- a/tests-src/buttons.html
+++ b/tests-src/buttons.html
@@ -485,7 +485,7 @@ resource: true
           </ul>
         </div><!-- /btn-group -->
       </div><!-- /btn-toolbar -->
-      <h2>buttons.js</h2>
+      <h2>Advanced Buttons - requires bootstrap.js</h2>
       <h3>Single toggle</h3>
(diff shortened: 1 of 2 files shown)


tests-src/buttons.html
```diff
@@ -484,4 +484,32 @@
             <li><a href="#">Separated link</a></li>
           </ul>
         </div><!-- /btn-group -->
-      </div><!-- /btn-toolbar -->
\ No newline at end of file
+      </div><!-- /btn-toolbar -->
+      <h2>buttons.js</h2>
```
erundle: I know this is what bootstrap calls its example but can we find a better name for this? "Advanced Buttons - requires bootstrap.js"

andresgalante:
I don't have any preference between "Buttons" or "Advanced Buttons". The rest :+1:

rhamilto:
Changed heading to @erundle's suggestion.

erundle:
Thanks. Will merge

PR #133, opened 2015-10-07
Dashboard Cards - Updated Ref. Impl. HTML/CSS

dtaylor113:
Now with filter dropdown in header or footer.


less/cards.less
```diff
@@ -49,6 +45,10 @@
   }
 }
 
+.cards-pf {
```
rhamilto: I'm not clear why you've moved this here as it introduces a mis-alphabetization.

dtaylor113: Sorry, my bad, fixed

dtaylor113:
Applied instructions for PR #134 "npm-shrinkwrap.json"

tests-src/_includes/cards.html
```diff
@@ -76,6 +76,77 @@ <h2 class="card-pf-title">
...
+                    <ul class="dropdown-menu">
+                      <li><a role="menuitem">Filter 1</a></li>
+                      <li><a role="menuitem">Filter 2</a></li>
+                      <li><a role="menuitem">Filter 3</a></li>
+                    </ul>
+                  </div>
+                </div>
+              </div>
+            </div>
+            <span>card sub-title</span>
```
rhamilto, line 103: To improve html semantics, it'd be better to write this as &lt;h3 class="card-pf-subtitle"&gt;

erundle:
Should we close this?

dtaylor113:
Not yet, I need to meet with Robb and clean up my css

erundle:
Yea sorry I was asking if this was still relevant. It sounds like it is.

PR #138, opened 2015-10-14
Increased the size of the donut chart custom example

dtaylor113:
Increased width & height to match pf donut pattern example

@rhamilto please review and merge if happy with changes.  
Needed for Leslie.
Thanks


rhamilto:
We should probably all discuss this.  Out of the box, C3 charts will assume the size of their parent container; therefore, we probably don't want to be setting an explicit height/width on the chart itself, but rather on the container as PatternFly is a mobile-first, responsive framework and the size of containers (i.e., grid columns) changes based on the size of the viewport, which in turn should determine the size of the chart.

cc: @kybaker, @LHinson

dtaylor113:
Ok, I'll setup a meeting

PR #142, opened 2015-10-16
Removing data-original-title from tooltip and popover examples

rhamilto:
Bug fix:  instances of "data-original-title" attribute should be "title" in the markup ("data-original-title" is the javascript-generated attribute).

@andresgalante, @erundle: review and merge, please.


andresgalante:
:+1:

PR #145, opened 2015-10-21
Revising alerts to meet new design specifications

rhamilto:


erundle:
Maybe people will actually see this now:
screen shot 2015-10-21 at 1 30 33 pm

+1 Nice Work!

PR #147, opened 2015-10-22
Revising Custom Donut Chart to better match visual comp and adding Utilization Card

rhamilto:
@andresgalante or @erundle, review and merge, please.


dtaylor113:
For the Utilization Card, if you change the unit label to of the middle Quotas chart to "115 of 460 MHz Used" it introduces the truncation marks "...".  IMO the unit label max-width is too small to accommodate the default unit label we are suggesting.

rhamilto:
@dtaylor113, you're referring to the Utilization Bar Card, which is not part of this PR.

dtaylor113:
> @dtaylor113, you're referring to the Utilization Bar Card, which is not part of this PR.

My apologies, got mixed up ;-)

less/cards.less
```diff
@@ -143,6 +165,34 @@
...
+  width: 100%;
+  .card-pf-utilization-card-details-count,
+  .card-pf-utilization-card-details-description {
+    display: table-cell;
+    float: left;
+    line-height: 1;
+    vertical-align: middle;
+  }
+  .card-pf-utilization-card-details-count {
+    font-size: 26px;
```
erundle: variable for font-size or is it not necessary at this size?

rhamilto: Good catch.  That should at least be a calculation, which I changed it to.

erundle:
LGTM

PR #148, opened 2015-10-23
Removing package.json and Gruntfile.js from Bower.json ignore

rhamilto:
So that they are included in the package distributed via Bower

@erundle, review and merge, please?


erundle:
LGTM

PR #150, opened 2015-11-02
Removing unnecessary inline styles from card examples

rhamilto:
Some unnecessary inline styles snuck in to the card examples.

@andresgalante or @erundle, review and merge, please?


erundle:
LGTM

PR #152, opened 2015-11-03
Fix for PTNFLY-801: Donut chart does not render label in IE

dtaylor113:


rhamilto:
Can you also update the 4 examples on https://github.com/patternfly/patternfly/blob/master/tests-src/_includes/cards.html, too, please?

dtaylor113:
Ok, updated Card Donut examples

erundle:
LGTM

PR #159, opened 2015-11-11
Correcting custom donut chart label unit

rhamilto:
@andresgalante, this one is easy.  :)  Too bad I didn't notice these instances the first time I attempted to fix this bug.  :)


dtaylor113:
That's my bad, thanks for fixing.  LGTM

PR #163, opened 2015-12-15
Count Remaining Characters Widget - Test Page

dtaylor113:
image


dist/css/patternfly.css
```diff
@@ -9035,6 +9035,14 @@ fieldset[disabled] .form-control {
 label {
   font-weight: 600;
 }
+.chars-remaining-pf span {
+  vertical-align: baseline;
```
rhamilto: I don't think this is needed as the default value is already baseline.  https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align

dist/css/patternfly.css
```diff
@@ -9035,6 +9035,14 @@ fieldset[disabled] .form-control {
...
+  padding-right: 5px;
+  font-weight: 600;
+}
+.chars-warn-remaining-pf {
+  color: red;
```
rhamilto: This should use a variable.  Most likely @brand-danger.

less/forms.less
```diff
@@ -31,3 +31,13 @@
 label {
   font-weight: 600;
 }
+
+.chars-remaining-pf span {
+  vertical-align: baseline;
```
rhamilto: I don't think this is needed as the default value is already baseline. https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align

less/forms.less
```diff
@@ -31,3 +31,13 @@
...
+  padding-right: 5px;
+  font-weight: 600;
+}
+
+.chars-warn-remaining-pf {
+  color: red;
```
rhamilto: This should use a variable. Most likely @brand-danger.

less/forms.less
```diff
@@ -31,3 +31,13 @@
 label {
   font-weight: 600;
 }
+
+.chars-remaining-pf span {
```
rhamilto: These rules should go first in the file (alphabetize rules by selector).

tests-src/forms.html
```diff
@@ -24,6 +24,38 @@
           </div>
         </div>
       </form>
+      <h2>Count Remaining Characters Control</h2>
+      <form>
+        <div class="form-group">
+          <label for="messageArea"></label>
+          <textarea class="form-control" id="messageArea" name="text" placeholder="Type in your message" rows="5"></textarea>
```
rhamilto, line 31: Please change all camel case instances to follow http://codeguide.patternfly.org/#css-classes

less/forms.less
```diff
@@ -31,3 +31,13 @@
...
+  vertical-align: baseline;
+  padding-right: 5px;
```
rhamilto: Please alphabetize per http://codeguide.patternfly.org/#css-declaration-order

tests-src/forms.html
dtaylor113, line 31: Hi, while I agree that class names should use dashes, I don't think ids or variables should use them; they should be camel case.  I followed the already existing precedence shown in dist/patternfly.js for ids and variable names.

less/forms.less
dtaylor113: fixed

less/forms.less
dtaylor113: fixed

less/forms.less
dtaylor113: fixed

less/forms.less
dtaylor113: fixed

tests-src/forms.html
dtaylor113, line 31: Note that the PatternFly Examples section directly above this new section uses camel case for IDs.

less/forms.less
```diff
@@ -1,6 +1,14 @@
 //
 // Forms
 // --------------------------------------------------
+.chars-remaining-pf span {
+  padding-right: 5px;
+  font-weight: 600;
```
rhamilto, line 5: Alpha fix, please.  (font-weight should come before padding-right per http://codeguide.patternfly.org/#css-declaration-order)

dtaylor113, line 5: fixed

PR #165, opened 2015-12-18
Fix c3ChartsDefaults function references.

jeff-phillips-18:
Using ‘this’ to access the inner functions in the overall Config
function causes issues when calling the function from within an object.
Update the function to be self contained.


rhamilto:
Ok by me.

dtaylor113:
LGTM

PR #171, opened 2016-01-18
Don't mark files as executable unnecesarily

stefwalter:
These files are erroneusly marked as executable. This is
anoying when trying to use them with bower in other projects
where other files become executable or are distributed as
executable.


jeff-phillips-18:
LGTM

PR #172, opened 2016-01-18
Wrapping any math in parenthesis so strict-math can be enabled

### LINKED ISSUE #168: Patternfly is not compatible with less --strict-math=on
variables.less from Patternfly 2.8.0 fails to build:

```
ArgumentError: error evaluating function `ceil`: argument must be a number in /data/src/cockpit/pkg/registry/bower_components/patternfly/less/variables.less on line 273, column 69:
272 @dropdown-link-hover-color:                                         @gray-pf;
273 @font-size-h1:                                                      ceil(@font-size-base * 2); // ~24px
274 @font-size-h2:                                                      ceil(@font-size-base * 1.8333); // ~22px
```

We see this problem in Cockpit. We're using less version `2.5.3`

rhamilto:
Resolves #168


dtaylor113:
LGTM

jeff-phillips-18:
LGTM

PR #182, opened 2016-01-27
Adds Drop down and up kebab variation

andresgalante:
@rhamilto @jeff-phillips-18 @dtaylor113 In order to finish list view I've created the kebab menu based on the drop menu.

This widget will also be applied to the toolbar and toast notification to keep consistency.

Can you take a look at it please?


dtaylor113:
LGTM

less/dropdowns.less
```diff
@@ -146,3 +146,68 @@
...
+    }
+  }
+  .dropdown-menu {
+    left: -15px;
+    margin-top: 11px;
+    &:before {
+      border-bottom: 11px solid @dropdown-border;
+      border-left: 11px solid transparent;
+      border-right: 11px solid transparent;
+      top: -11px;
```
rhamilto: Alpha correction of declarations needed within this rule.

rhamilto:
The arrow looks a little funky.  Maybe the white is off by a pixel?

And perhaps the gray arrow should be a tad darker so it more closely matches the straight border that isn't anti-aliased?

screen shot 2016-01-27 at 11 47 07 am

less/dropdowns.less
```diff
@@ -146,3 +146,68 @@
...
+    padding-bottom: 0;
+    &:active,
+    &:focus,
+    &:hover {
+      color: @link-color;
+    }
+  }
+  .dropdown-menu {
+    left: -15px;
+    margin-top: 11px;
```
jeff-phillips-18, line 166: Should these values for top/margin-top, bottom/margin-bottom, etc be variables?

andresgalante, line 166: Bootstrap has `dropdown` and `dropup` and it also has a class `dropdown-menu-right`. By default its on the left, if you add that class it goes to the right. Up and down are defined by `dropup` or `dropdown` classes.

The only new class I am introducing to the component is `.dropdown-kebab-pf`

jeff-phillips-18:
LGTM

less/dropdowns.less
```diff
@@ -146,3 +146,68 @@
...
+      border-bottom: 10px solid @dropdown-border;
+      border-left: 10px solid transparent;
+      border-right: 10px solid transparent;
+      content: "";
+      display: inline-block;
+      left: 13px;
+      position: absolute;
+      top: -11px;
+    }
+    &:after {
```
rhamilto, line 189: Alpha fix.  Rule should come before &:before.

less/dropdowns.less
```diff
@@ -146,3 +146,68 @@
...
+    &:active,
+    &:focus,
+    &:hover {
+      color: @link-color;
+    }
+  }
+  .dropdown-menu {
+    left: -15px;
+    margin-top: 11px;
+    &:before {
```
rhamilto, line 177: You could DRY &:after and &:before up a bit (combine shared declarations).

less/dropdowns.less
```diff
@@ -146,3 +146,68 @@
...
+      &:after {
+        left: auto;
+        right: 13px;
+      }
+    }
+  }
+  &.dropup .dropdown-menu {
+    margin-bottom: 11px;
+    margin-top: 0;
+    &:before {
```
rhamilto, line 198: You could DRY &:after and &:before up a bit (combine shared declarations).

Alpha fix.  &:after should come before &:before.

tests-src/dropdowns.html
```diff
@@ -29,3 +29,39 @@
           <li><a href="#">Separated link</a></li>
         </ul>
       </div>
+      <h2>Dropdown Kebab</h2>
+      <div class="dropdown dropdown-kebab-pf">
```
rhamilto: It seems like there is a lot of repeated markup.  What if it all went in an include and variables added the addtional needed classes, etc.  See https://github.com/patternfly/patternfly/blob/master/tests-src/_includes/cards.html#L348 for an example.

tests-src/dropdowns.html
```diff
@@ -29,3 +29,39 @@
...
+          <li><a href="#">Action</a></li>
+          <li><a href="#">Another action</a></li>
+          <li><a href="#">Something else here</a></li>
+          <li><a href="#">Separated link</a></li>
+        </ul>
+      </div>
+      {% include widgets/kebab.html %}
+      <h2>Dropup Kebab</h2>
+      <div class="dropup dropdown-kebab-pf">
+        <button class="btn btn-link dropdown-toggle" type="button" id="dropdownKebab" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
```
rhamilto: Repeated id attribute.  Valid HTML requires id attributes are unique.

less/dropdowns.less
```diff
@@ -146,3 +146,66 @@
...
+      color: @link-color;
+    }
+  }
+  .dropdown-menu {
+    left: -15px;
+    margin-top: 11px;
+    &.dropdown-menu-right {
+      left: auto;
+      right: -15px;
+      &:before,
```
rhamilto: Ahem.  "a" comes before "b".  ;-)  Two additional instances below.

andresgalante:
Now it should be ok :)

andresgalante:
@rhamilto @dtaylor113 @jeff-phillips-18 Can we close without merging? I'll send send a new PR today with the kebab plus listview, toolbar and toast.

PR #185, opened 2016-01-28
Fix-bug: paging of datatables doesn't work well when there are multiple tables

dabeng:
For example, if user creates two datatables on the same page, when he clicks the "next page" button of one table, the other table also turn to next page at the same time.


jeff-phillips-18:
Please include the patternfly.min.js file generated by the grunt build.

dtaylor113:
Hi, could you please rebase and we will merge this fix into master.
Thank you!

dabeng:
Hi, @dtaylor113 , please merge it.

PR #200, opened 2016-02-25
Fix for donut text sizes in list view.

jeff-phillips-18:
Do not set all donut chart text sizes to be the large font, only set
the large font when setting the title via the PF chart utilities.


dtaylor113:
LGTM

PR #205, opened 2016-03-01
Add less variables for Patternfly Color Palette colors

jeff-phillips-18:


dtaylor113:
LGTM

andresgalante:
@jeff-phillips-18 I think we should remove the color without number, since they are the same as `-400`.

jeff-phillips-18:
@andresgalante I believe we want both so that we have the ability to modify if necessary.

andresgalante:
@jeff-phillips-18 I don't agree, if we want to modify blue-400 then we modify it. duplicating them will only generate problems since if blue-400 changes blue will also have to change. Its not DRY.

jeff-phillips-18:
@andresgalante OK, how about we change it to be
@color-pf-blue:  @color-pf-blue-400  ?

The point being we should have a standard blue.

andresgalante:
@jeff-phillips-18 that is a better solution, but I wouldn't duplicate the names. Default is 400. I guess I am biased based on material design where default is 500, there is no default name without a saturation value.

jeff-phillips-18:
@andresgalante Updated.

PR #209, opened 2016-03-03
Simple jQuery function to display the current PatternFly version as re…

dtaylor113:
…corded in the bower.json file

Simple function to retrieve the current version stored in the PF bower.json file.  Version will be set to any HTML element with id of 'pfversion'.

This function will be used to display the current PF version number on the main page of the PF.org site.  It is a good way to verify that the PF.org site is using the latest PF base ref version.


dlabrecq:
Looks good.

dist/js/patternfly.js
```diff
@@ -1373,3 +1373,13 @@
   };
 }(jQuery));
 
+// Util: Retrieve Current PatternFly Version Number
+(function ($) {
+  'use strict';
+
+  $.fn.getPatternFlyVersion = function () {
+    $.getJSON("../../bower.json", function (data) {
```
jeff-phillips-18, line 1381: This function doesn't 'get' any thing. It updates any components of the particular id to the version string. I would rename it appropriately and it would be nice to allow the selector to be passed in rather than forcing a particular id on the component to be updated.

jeff-phillips-18:
Should this really be part of the patternfly library? Couldn't this script just be added to patternfly.org site?

dtaylor113:
Ok, closing this as I see where I can add it in the patternfly-site repo.

PR #217, opened 2016-03-08
Fix missing instructions

rafabene:
NPM instruction doesn't contain the style installation. This was added in this commit


jeff-phillips-18:
:+1:

PR #219, opened 2016-03-14
Adding individual Bootstrap table examples

rhamilto:
@ajacobs21e are working on new table styling for https://github.com/openshift/origin.  In reviewing the existing tables test page, I noticed it doesn't include all the individual Bootstrap options.  If I remember correctly, that was done because the official PatternFly table should include the striped, bordered, hover, and responsive options.  That said, it's challenging to see the effects of PatternFly on these individual Bootstrap options if they're not individually included on the test page, so I'm adding them to solve that limitation.

@andresgalante, @dtaylor113, @jeff-phillips-18, PTAL.

### DIFF
--- a/tests-src/tables.html
+++ b/tests-src/tables.html
@@ -4,62 +4,7 @@ layout: page
 title: Tables
 resource: true
 ---
-      <h2>Contextual classes</h2>
-      <table class="table table-bordered">
-        <thead>
-          <tr>
-            <th>#</th>
-            <th>Column heading</th>
-            <th>Column heading</th>
-            <th>Column heading</th>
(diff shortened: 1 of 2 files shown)


dtaylor113:
LGTM

jeff-phillips-18:
Can we put the recommended tables at the top of the page and the bootstrap examples after? I'd prefer the first thing folks see is the recommended usage.

rhamilto:
@jeff-phillips-18, yep.  I've changed it so the Bootstrap examples come at the end of the page.

jeff-phillips-18:
LGTM

PR #222, opened 2016-03-16
PTNFLY-1309

cardosogabriel:
Task accomplished, but I would like to verify some points with someone more experienced:
1. I got the svgs in dropbox links. Did not figure how to save them, so copied the svg code and saved them using my IDE. 
2. Opened them in illustrator, and deleted the blue grid lines in the background. 
3. Reduced the images / SVG canvas to 16px (which is the default size) and saved.

Is this procedure ok? For the next one, maybe we could have the icons in files already without the background, so not necessary to save again?
Thanks


bdellasc:
Those blue grid lines shouldn't be visible when used (that was my understanding). We have them setup as larger files with a grid because that's how It is recommended the source files be built. Very large (multiple of 14), with a grid as a reference. This gives us maximum flexibility on resizing them to whatever we need.

The default size is actually 14px. This decision was driven by Font Awesome's size, since we were utilizing a number of their icons and we wanted the grid sizes to match between the font awesome icons and patternfly icons.

cardosogabriel:
Thanks for the explanation @bdellasc. I updated the PR with the icons without modifying them in Illustrator. It looks good.

dist/fonts/PatternFlyIcons-webfont.svg
```diff
@@ -43,7 +43,7 @@
...
-<glyph unicode="&#xe611;" glyph-name="remove2" data-tags="remove, delete, trashcan, recycle bin, bin, dispose" horiz-adv-x="878" d="M146.286-73.143h585.143l73.143 658.286h-731.429l73.143-658.286zM585.143 804.855v146.002h-292.571v-146.002l-292.571-0.283v-219.429l73.001 73.143h731.57l73.143-73.143v219.712h-292.571zM511.010 804.855h-145.296v73.001h145.296v-73.001z" />
+<glyph unicode="&#xe611;" glyph-name="delete" data-tags="remove, delete, trashcan, recycle bin, bin, dispose" horiz-adv-x="878" d="M146.286-73.143h585.143l73.143 658.286h-731.429l73.143-658.286zM585.143 804.855v146.002h-292.571v-146.002l-292.571-0.283v-219.429l73.001 73.143h731.57l73.143-73.143v219.712h-292.571zM511.010 804.855h-145.296v73.001h145.296v-73.001z" />
```
jeff-phillips-18: Is this change intentional?

cardosogabriel: Yes. I realized that the names here follow the classes used for them, and in this case it is pficon-delete. 
But now looking better I see that some don't really follow.
U want to revert the name?

jeff-phillips-18: Yeah, I would revert it just so there are no unexpected repercussions

cardosogabriel: Reverted. Thanks

rhamilto:
@cardosogabriel, it looks like PatternFlyIcons-webfont.json didn't update properly as icons from the Icomoon - Free set are missing from the PatternFlyIcons-webfont.json (as is the project metadata).  Did you import the existing https://github.com/patternfly/patternfly/blob/master/PatternFlyIcons-webfont.json as a project, load the project, make changes, and then return to Manage Projects to Download?

cc: @LHinson

cardosogabriel:
@rhamilto, I see what you are saying. I did as u said until "make changes", but instead of returning to manage projects to download, I selected Generate Font. Now I redid the procedure as you described and it looks ok. Thanks for the review!

rhamilto:
:+1:

bdellasc:
This may be something worth documenting so if the responsibility
surrounding the font changes hands down the line, it's available.

jeff-phillips-18:
LGTM

PR #232, opened 2016-03-21
Bumped version to 3.3.1

dlabrecq:


dtaylor113:
Hi, LGTM.

I would add to the release notes something about "regenerated npm-shrinkwrap" because we did work with someone who did download 3.3.0 and tried to do a NPM-Only install, which didn't work for that release and they maybe waiting to see it resolved in 3.3.1  -thanks

dlabrecq:
Regenerating npm-shrinkwrap is part of the release process notes now. Will we need to add release notes about that for each release?

dtaylor113:
No, only when we forget to do it for a release like we did for 3.3.0 :-)

jeff-phillips-18:
LGTM

PR #237, opened 2016-03-24
PTNFLY-1142 Attributing same width to create column layout.

cardosogabriel:
My approach to the solution was to add a class 'list-view-pf-additional-info-column' to the elements 'list-view-pf-additional-info' when you want to have this column view.
In the implementation, you can see this working on the 2nd and 3rd last lines of the list.
Instead of setting a fixed width that is vulnerable, I'm measuring the elements and applying the biggest width to all.


cardosogabriel:
@andresgalante , @jeff-phillips-18 , can u review?

andresgalante:
@cardosogabriel I don't quite get it. I see that this [solution works](https://cdn.rawgit.com/cardosogabriel/patternfly/PTNFLY-1142/tests/list-view.html), but there is no less file to review.
Also I would avoid adding a `list-view-pf-additional-info-column` to each element and go for a modifier class on `list-view`

cardosogabriel:
The changes were in the html file only.
We can have a modifier in list-view, but with this approach the user can select list lines that will have this alignment. I probably did this way because of the patternfly list that is present, with different entries. If u think I should have per full list I can change.
In tests-src/list-view.html you can see the javascript I wrote.

andresgalante:
@cardosogabriel thanks or the explanation. I'll test it.

andresgalante:
@dlabrecq @dtaylor113 @jeff-phillips-18 Can you please review the javascript on tests-src/list-view.html ?

dtaylor113:
Hi, can you please elaborate on what issue you are trying to address?  Perhaps with a screenshot showing the column sizes in question?  -thanks

andresgalante:
@dtaylor113 its [here](https://cdn.rawgit.com/cardosogabriel/patternfly/PTNFLY-1142/tests/list-view.html).
screen shot 2016-05-13 at 2 56 50 pm

tests/list-view.html
```diff
@@ -515,6 +515,17 @@
...
 </div>
+<script>
+  // Fixed column width
+  $(document).ready(function () {
+    var setWidth = 80;
+    $('.list-view-pf-fixed-column .list-view-pf-additional-info-item').each(function(){
+      var elementWidth = $(this).width();
+      if (elementWidth > setWidth) setWidth = elementWidth;
+    });
+    $('.list-view-pf-fixed-column .list-view-pf-additional-info-item').css('width', setWidth + 'px');
```
dlabrecq: You're looking for the widest possible element, so you have to iterate over each and every one?

Seems odd to set width like this, but it appears that's what we do with jQuery and matchHeight.

dlabrecq: Here is a slightly cleaner way to write the JavaScript.

http://robido.com/jquery/use-jquery-to-match-height-or-width-of-elements-equalize-to-make-them-all-the-same-uniform-height-or-width/

andresgalante:
@LHinson this is the only PR we have open ATM. What do you guys want to do with it?

jeff-phillips-18:
LGTM

LHinson:
@andresgalante what do you think? I'm not familiar with this problem/solution to be able to say...

cardosogabriel:
The code is delivering what as asked. Only thing is that Dan commented that
there is a better way to write the javascript. I don't know much of jquery,
so I can give a try on Monday. If in trouble, perhaps that could provide
the proper code.

andresgalante:
thanks @cardosogabriel if you want to try it, go ahead. If you find it too challenging you can ask for guidance or ask @LHinson to schedule it for a js dev.

jeff-phillips-18:
@dlabrecq Could you implement the change you would like to see and send it along to @cardosogabriel ? Thanks.

cardosogabriel:
Applied the code suggested by Dan. It works and has a better jquery. Thanks @dlabrecq +1

andresgalante:
@cardosogabriel thanks! 

@dlabrecq if this is ok, please merge it 🚀

jeff-phillips-18:
Change looks good!

dlabrecq:
LGTM

Issue #233, opened 2016-03-22
Overly specific bower dependencies


himdel:
Depending on a specific (patch) version of a bower package makes it impossible to use external libraries that depend on one of the patternfly-required libraries unless the version matches exactly.

There is no need for that, for packages that do follow semver, just make the dependency `>~`.

My specific problem is moment - `"moment": "2.11.2"` which should probably depend on `>~ 2.11` (or `>= 2.11.0 && < 2.13`).

dtaylor113:
Hi @dlabrecq,
Can you please look at making this adjustment so it will be included in our next patternfly release?
Thanks,
- Dave

dlabrecq:
I'll plan on changing this to ~2.11

ceefour:
Please either update moment to `~2.12.0` or lax the version dependency as @himdel suggested: >= `2.11.0 && < 2.13`.

What's unfortunate is that https://github.com/Eonasdan/bootstrap-datetimepicker 4.15.35 actually depended on moment `~2.8` :( Latest version is 4.17.37 which [is not yet available in `npm`](https://github.com/Eonasdan/bootstrap-datetimepicker/issues/1522) :(

dlabrecq:
This issue was addressed with PR #248.

PR #239, opened 2016-03-24
Overly specific bower dependencies #233

dlabrecq:
Modified moment dependency to be ~2.11


dtaylor113:
This PR title is "Overly specific bower dependencies" which seem to imply multiple dependencies updated, whereas it was just for 'moment'.   Should we change all dependencies to have a "~" prefix on major release number?
Ex: 
~1.2.0: must be at least as high as 1.2.0 but less than the next major version

dlabrecq:
I can make that change. Although, some dependencies use the caret character (e.g., "grunt-contrib-copy": "^0.8.1"). Probably shouldn't change those.

dtaylor113:
Yes, I believe carrot has similar behavior

dlabrecq:
Can this be merged?

jeff-phillips-18:
LGTM

Issue #236, opened 2016-03-24
bootstrap-treeview depends on bower and phantomjs


jmikulka:
I'm using npm managed version of PatternFly to avoid need of installing bower. However one of the dependencies (`bootstrap-treeview`) has bower as a dependency. Besides that it also depends on phantomjs which seems to me like overkill. This is present since `patternfly@3.1.0`.

Would it be possible make all dependencies optional (https://docs.npmjs.com/files/package.json#optionaldependencies) and to ship only `dist`?

dtaylor113:
Hi,
We can look into moving bootstrap-treeview into 'optionalDependencies' in the package.json file.

Can you please elaborate on what you mean by: "...ship only dist"? 

We used to ship all the dependencies in a '/components' directory, however this changed in v3.0.0 due to a number of patternfly users having issues with the dependencies we shipped and those same dependencies in their individual applications.  Uses requested we stop shipping patternfly dependencies and have users rely on the NPM and/or Bower to install the dependencies.

Thanks,
- Dave Taylor

dtaylor113:
Hi @dlabrecq,
Can you please look at making this adjustment so it will be included in our next patternfly release?
Thanks,
- Dave

dlabrecq:
I have some questions. Can we talk, @dtaylor113 ?

jmikulka:
Hi @dtaylor113,

correct me if I am wrong but I assume that all necessary PatternFly files (js, css, fonts and images) are installed into the `dist/` (`node_modules/patternfly/dist/` and `bower_components/patternfly/dist/`). If I decide to use datatables or bootstrap-select in my project, I should explicitly install these extra packages alongside PatternFly to avoid installing lots of unused packages.

Hope that clarifies my concerns,
Jiri

jmikulka:
Hi @dlabrecq,

thanks for moving bootstrap-treeview from dependencies to optionalDependencies. I was just wondering if it would make sense to move all but really essential (I assume the only crucial dependencies are jquery and bootstrap) dependencies to optional? 

In my use case I'm using npm to install PatternFly, jQuery, Bootstrap and datatables only. In my html I include only jquery, bootstrap, datatables and patternfly (in this order). Everything works like a charm which means that none of the current PatternFly dependencies is a real dependency but is an optional dependency (except jquery and bootstrap).

Thanks,
Jiri

dlabrecq:
[...]

PR #240, opened 2016-03-24
PTNFLY-236: make all dependencies (but jquery and bootstrap) optional

jmikulka:
Changes that I have suggested in #236.

This solution allows one to install PatternFly without any extra dependencies using `npm install --no-optional` command.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,22 +5,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "bootstrap": "3.3.5",
-    "bootstrap-datepicker": "1.4.0",
-    "bootstrap-select": "1.7.3",
-    "bootstrap-switch": "3.3.2",
-    "bootstrap-touchspin": "3.1.1",
-    "bootstrap-treeview": "1.2.0",
-    "c3": "0.4.10",
-    "datatables": "1.10.9",
-    "datatables.net-colreorder": "1.3.1",
-    "drmonty-datatables-colvis": "1.1.2",
-    "font-awesome": "4.3.0",
-    "google-code-prettify": "1.0.1",
-    "jquery": "2.1.4",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -16,26 +33,11 @@
     "drmonty-datatables-colvis": "1.1.2",
     "font-awesome": "4.3.0",
```
jeff-phillips-18, line 34: I would leave font-awesome as a dependency. There are too many places it is used to consider it optional.

jeff-phillips-18:
@rhamilto If you have a moment to look at this, your insights would be appreciated.

dlabrecq:
I should mention that we've recently added the '~' char to dependencies -- see PR #239.

rhamilto:
@jeff-phillips-18, this should be fine.

@dlabrecq, I had specifically locked the dependencies to specific versions because we were seeing CSS changes resulting from differing patch versions of dependencies being installed depending on when `bower install` was run; in hindsight, only the offending dependencies should have been locked.  In looking at https://github.com/patternfly/patternfly/pull/239, it appears that only Boostrap-Select needs to be locked.

dlabrecq:
Thanks @rhamilto. I'll fix the Boostrap-Select dependency.

dlabrecq:
Discussed with @jeff-phillips-18 and @dtaylor113 and agreed to making the dependencies optional. @jmikulka do you prefer to make the changes discussed in the code review or have us do it?

jmikulka:
@dlabrecq I'd actually prefer you do it because I can see that you are planning to do more changes than I initially planned. Is that ok? Should you need any help, just give me a shout! Thanks.

dlabrecq:
Ok, I just created a PR. Closing in favor of PR #248.

PR #241, opened 2016-03-30
Locking version for Boostrap-Select

dlabrecq:
"I had specifically locked the dependencies to specific versions because we were seeing CSS changes resulting from differing patch versions of dependencies being installed depending on when bower install was run; in hindsight, only the offending dependencies should have been locked. In looking at #239, it appears that only Boostrap-Select needs to be locked." -- rhamilto


dlabrecq:
Closing because this has been fixed when bumping version number to 3.3.2.

PR #242, opened 2016-03-30
Bumped version number to 3.3.2

dlabrecq:


dtaylor113:
LGTM

jeff-phillips-18:
Should this be 3.4.0 or 3.3.2? Is this a bug fix release?

jeff-phillips-18:
I believe we also need to update the version number in bower.json

dlabrecq:
I only see a couple bug fixes in the commits, so I chose v3.3.2. The version number in both bower.json and package.json have been updated?

jeff-phillips-18:
OK, I missed the bower.json change my bad.

Just wanted to make sure the version number change was intentional, I agree with your assessment.

LGTM

PR #247, opened 2016-03-31
PTNFLY-1184 Migrate all PF "Cards" Content over to the site

dlabrecq:
This creates templates for use by both patternfly tests and patternfly-toolkit pages.
@dtaylor113 and @jeff-phillips-18 please review.


jeff-phillips-18:
Why the changes to bootstrap less files?

tests-src/_includes/widgets/dashboard.html
```diff
@@ -230,12 +230,16 @@ <h2 class="card-pf-title">
...
       <div class="row row-cards-pf">
-      <!-- Important:  if you need to nest additional .row within a .row.row-cards-pf, do *not* use .row-cards-pf on the nested .row  -->
-{% include widgets/card-pf-utilization-single-metric.html chart1="chart-pf-donut-4" chart2="chart-pf-sparkline-4" %}
+        <div class="col-xs-12 col-sm-4 col-md-4">
+          <!-- Important:  if you need to nest additional .row within a .row.row-cards-pf, do *not* use .row-cards-pf on the nested .row  -->
+          {% include widgets/card-pf-utilization-trend-single-metric.html chart1="chart-pf-donut-4" chart2="chart-pf-sparkline-4" %}
+        </div>
       </div><!-- /row -->
       <div class="row row-cards-pf">
       <!-- Important:  if you need to nest additional .row within a .row.row-cards-pf, do *not* use .row-cards-pf on the nested .row  -->
```
jeff-phillips-18, line 245: The dashboard widget should use the new card widgets rather than re-creating them.

jeff-phillips-18:
At this point, the widgets directory is getting a bit unruly. Maybe we should have a widgets/cards directory and a widgets/nav directory?

dlabrecq:
I didn't change any less files, the build did that. I only created the new widget templates.

tests-src/_includes/widgets/dashboard.html
dlabrecq, line 245: I don't understand "re-creating them" as this is one of the new templates that Robb previously created. I'm not recreating anything here. I simply renamed the template to make it more clear that it contains a utilization trend card.

jeff-phillips-18:
The bootstrap less files are copied from the bower install area (components). You have a different version of bootstrap-select in your area. Please make sure this is the correct version we want (1.7.3). Either yours is not or the previous commit was not.

dlabrecq:
I'll move the card widgets/templates to their own directory per your suggestion. I'll take a look at navigation seperately.

dlabrecq:
Patternfly's bower.json and package.json files have bootstrap-select 1.7.3. I've nuked components and node_modules to be certain my workspace has the right version. 

NPM installs bootstrap-select 1.7.3. However, the components/bootstrap-select/bower.json file says 1.7.2.  The .bower.json (same directory) shows version 1.7.3. Not sure what I need do here?

tests-src/_includes/widgets/dashboard.html
jeff-phillips-18, line 245: The dashboard.html widget defines a bunch of cards to show on the dashboard. Rather than include the html for each of those cards in dashboard.html, replace that html with includes of the new card widgets you have created.

jeff-phillips-18:
OK, then it could have been the previous commit that was not up to date.

dlabrecq:
Sorry, I think nuking the components and node_modules directories did the trick. I've rebased and pushed those changes.

dlabrecq:
Regarding the dashboard, I have been replacing HTML with any widget/templates I create. However, I'm looking at the dashboard and don't see anything else that can replaced with the templates I created from the cards test page. That is, nothing that hasn't been already replaced by Robb. There are far more complex examples at the bottom of the dashboard page and don't want to blow that away.

jeff-phillips-18:
That's fine. That can be done later.

LGTM

dtaylor113:
Ok, cloned and reviewed cards & dashboard test pages.  LGTM

dlabrecq:
Thanks!

PR #250, opened 2016-04-07
PTNFLY-1333 Adapting List View widgets/page to fit the new website needs.

cardosogabriel:


dlabrecq:
LGTM

jeff-phillips-18:
Please use jekyll 2.5.3 to build to remove the unintentional changes to the tests files.

jeff-phillips-18:
Please install bootstrap 3.3.6 and rebuild rather than 3.3.5 to remove the unintended changes to the bootstrap less files.

jeff-phillips-18:
Please install font-awesome 4.5.0 and rebuild to remove the unintended changes to the font-awesome less files.

tests-src/_includes/widgets/list-view/list-view-variation-1.html
```diff
@@ -0,0 +1,121 @@
...
+    order: null
+  };
+  c3.generate(customDonutChartConfig);
+  donutChartTitle = d3.select("#chart2").select('text.c3-chart-arcs-title');
+  donutChartTitle.text("");
+  donutChartTitle.insert('tspan').text("21%").classed('list-view-donut-title', true);
+
+  customDonutChartConfig = c3ChartDefaults.getDefaultDonutConfig('14%');
+  customDonutChartConfig.size.height = 71;
+  customDonutChartConfig.donut.width = 5;
```
jeff-phillips-18, line 84: This should probably be included in the c3ChartDefaults. For instance c3ChartDefaults.getDefaultInLineDonutConfig

cardosogabriel:
Rebuilt with the versions @jeff-phillips-18 specified. Looks better now.

tests-src/_includes/widgets/list-view/list-view-variation-1.html
jeff-phillips-18, line 84: I didn't realize this existed already and was simply moved from one file to another. You can skip this for now and we can address it in the future.

jeff-phillips-18:
LGTM, please rebase and squash.

jeff-phillips-18:
After the rebase, it looks like the wrong version of jeykll was used to build.

cardosogabriel:
I apologize. When you left I started to work on patternfly-toolkit, and somehow it only works for me when I do bundle install, which brings jekyll 3.1.2 :(
It would be great to find a solution for this.

LHinson:
@dlabrecq seems like if patternfly and pf-toolkit were using the same version, this wouldn't be as painful. Instead of taking all migration stories next sprint, how about a story to look into this problem?

dlabrecq:
Curious what "brings jekyll 3.1.2" as I had to install Jekyll myself.

Also, what do you mean by "do bundle install"? Typically, the steps to install/build patternfly-toolkit are:

npm install
bower install
gem install jekyll -v 2.5.3 (If needed per the patternfly readme)
grunt build.

jeff-phillips-18:
looks good @cardosogabriel

cardosogabriel:
@dlabrecq , I got many errors trying to run grunt with jekyll 2.5.3. It would always say that something is missing, and I would install the missing thing, and it would ask for another one. In other shots, it would give an error because of kramdown (or something like that). So I spent a lot of time researching and found that bundle install solve these issues.
As I said, when u run it, it does solve to issues, but install jekyll 3.1.2.
For u to see, I runned the commands you said and copyed the log. Again, got error in grunt build.
https://gist.github.com/cardosogabriel/bc14913d13e1abb42e8808ea908568a4
Please let me know if there is another way to solve it.

LHinson:
@dlabrecq I just created a bug here: https://patternfly.atlassian.net/browse/PTNFLY-1361 regarding this issue. Hoping you can pull this in this sprint to address.

dlabrecq:
I don't believe this is a version issue between PF and PatternFly toolkit. Both builds use the version of whatever Jekyll has been installed on your system.

This looks more like @cardosogabriel may have a permissions problem? When I install Jekyll, I don't install it as root (i.e., using sudo). 

I fixed a similar problem for @LHinson, during the UXD meetup. The only thing you should need to install as root is bower because that's installed globally.

PR #251, opened 2016-04-08
PTNFLY-1337: Donut charts are supposed to be centered

jeff-phillips-18:


dtaylor113:
LGTM. +1 for comments

PR #277, opened 2016-05-06
Mini donut chart example needs slight height increase

dlabrecq:
Mini donut chart example needs slight height increase to avoid legend wrapping on Firefox.
This issue is best seen when the template is included on the new patternfly-org site.


dtaylor113:
LGTM

jeff-phillips-18:
LGTM

PR #282, opened 2016-05-11
PTNFLY-1395 Notifications Drawer

cardosogabriel:
Initial implementation for notifications drawer. Tested in all required browsers and it works :)
- Fixed accordion "dancing" title when expanding it (panels.less)
- Current implementation of the drawer only works with Vertical Navigation Bar

@jeff-phillips-18 , I need some of your magic to:
- Close the drawer when clicking outside of it
- Infinite scroll. Style is there, but not the functionality.
- Whenever scrolling inside the drawer, make it _not_ scroll the page when the inside scroll finishes. I saw that in your fixed height accordion, but could not make it work here.
- Improvement: Make the whole accordion area clickable to expand. That would require changes in the accordion component.

Not sure why some c3 css changes came in the push. Please let me know if I rebased wrongly.


less/notifications-drawer.less
```diff
@@ -0,0 +1,102 @@
...
+// --------------------------------------------------
+
+.drawer-pf {
+  background-color: @color-pf-black-100;
+  border: 1px solid @card-pf-border-color;
+  .box-shadow(0 6px 12px rgba(0,0,0,.175));
+  overflow-y: scroll;
+  position: absolute;
+  right: 0;
+  width: (@grid-gutter-width * 8);
```
andresgalante: @cardosogabriel It seems strange to me to measure it with the gutter width, why not doing something like:

``` less
max-width: 31.1% ; // sets the size to 4 cols;
```

for `md` viewport

less/notifications-drawer.less
```diff
@@ -0,0 +1,102 @@
...
+  .panel {
+    border-left: none;
+    border-right: none;
+  }
+  .panel-body { padding: 0; }
+  .panel-counter {
+    display: block;
+    font-style: italic;
+    line-height: 1.2;
```
andresgalante, line 26: @cardosogabriel Can you make it based on `@line-height-base` ?

less/notifications-drawer.less
```diff
@@ -0,0 +1,102 @@
+//
+// Notifications Drawer
+// --------------------------------------------------
+
+.drawer-pf {
+  background-color: @color-pf-black-100;
+  border: 1px solid @card-pf-border-color;
+  .box-shadow(0 6px 12px rgba(0,0,0,.175));
+  overflow-y: scroll;
```
andresgalante: Change this to "auto" to avoid this:
screen shot 2016-05-13 at 11 25 02 am

less/notifications-drawer.less
```diff
@@ -0,0 +1,102 @@
...
+  .btn {
+    padding: 15px 0;
+    width: 318px;
+  }
+}
+
+.drawer-pf-loading {
+  color: @gray-pf;
+  font-size: (@font-size-base + 2);
+  text-align: center;
```
andresgalante: Maybe use `.text-center` instead?

less/notifications-drawer.less
```diff
@@ -0,0 +1,102 @@
...
+.drawer-pf-title {
+  background-color: @color-pf-black-100;
+  border-bottom: 1px solid @card-pf-border-color;
+  position: fixed;
+  width: 318px;
+  h3 {
+    font-size: @font-size-base;
+    margin: 0;
+    padding: 6px 15px;
+    text-align: center;
```
andresgalante: same here, we might use `.text-center`

tests-src/_includes/widgets/layouts/nav-vertical-notification-drawer.html
```diff
@@ -0,0 +1,119 @@
...
+        <button class="btn btn-link">Mark All Read</button>
+      </div>
+    </div>
+  </nav>
+</nav> <!--/.navbar-->
+<script>
+  $(document).ready(function() {
+    // Show/Hide Notifications Drawer
+    $('.drawer-pf-trigger').click(function() {
+      $(this).toggleClass('open');
```
andresgalante, line 114: @cardosogabriel @jeff-phillips-18 as you guys know I don't know js, but I see that bootstrap uses things like `aria-hidden="true"` anyway, do whatever you think its best. Also I don't like the classes `open` and `hide` it should be something like open-close or show-hide.

andresgalante:
@jeff-phillips-18 can you pick @cardosogabriel branch and send him a PR with the js, se we can merge all tother in this PR please? is it a good idea?
 btw, Gabril is aware that he has conflicts, he wants to rebase and solve them after he has the js :)

andresgalante:
@cardosogabriel the background of the accordion is white. Can you please add a bg there please?
Thanks!

jeff-phillips-18:
My understanding is that this PR is to get the base css/html in to pattternfly. There will be a follow-up PR to add the javascript functionality. I would suggest we continue with this PR without the auto-close or changes to the fixed accordion for now. I will add the javascript functionality in th follow-up PR. @LHinson  Do you agree?

tests-src/_includes/widgets/layouts/nav-vertical-notification-drawer.html
dlabrecq, line 114: The aria attributes are for accessibility, which we should be doing as well.

less/notifications-drawer.less
dlabrecq: Are you suggesting the example use div class="drawer-pf-title text-center"> instead?

andresgalante:
@jeff-phillips-18 ok, sounds good to me. @cardosogabriel can you rebase and solve conflicts please?

less/notifications-drawer.less
andresgalante: @dlabrecq nope, I think I'd keep it in the less.

tests-src/_includes/widgets/layouts/nav-vertical-notification-drawer.html
andresgalante, line 114: @dlabrecq you are right, sorry, I was talking about the `data-toggle=""`attribute and not `aria`

less/notifications-drawer.less
dlabrecq: I'm curious, why add the extra complexity? For the sake of maintainability, .text-center doesn't seem to add any value here.

less/notifications-drawer.less
```diff
@@ -0,0 +1,102 @@
...
+      border-left: 1px solid @navbar-pf-navbar-utility-border-color;
+      border-right: 1px solid @navbar-pf-navbar-utility-border-color;
+      padding-left: 15px;
+      padding-right: 15px;
+    }
+    &.open .drawer-pf-trigger-icon {
+      background-color: @navbar-pf-navbar-utility-hover-bg-color;
+    }
+  }
+}
```
dlabrecq, line 116: Seeing a lot of nesting and curious if this is necessary? Own coding guide at http://codeguide.patternfly.org/#css-nesting reads:

"Avoid unnecessary nesting. Just because you can nest, doesn't mean you always should. Consider nesting only if you must scope styles to a parent and if there are multiple elements to be nested."

less/notifications-drawer.less
cardosogabriel: We agreed to have it fixed at 320px width.

less/notifications-drawer.less
cardosogabriel, line 26: Patternfly line-height-base is 1.667. To achieve the visual design wanted, I need to use 1.2 here.

tests-src/_includes/widgets/layouts/nav-vertical-notification-drawer.html
cardosogabriel, line 114: @andresgalante , I used the class open because it is the class bootstrap add in the dropdown component. I leave to the js guys to update it if they find something else better.

less/notifications-drawer.less
cardosogabriel, line 116: @andresgalante Yes, will rebase. Am fixing some last things based on comments.
Regarding the white color, inside the accordion, when expanded is white. So I'm putting the notification items as white, but keep the drawer bg in the current gray. Also better for contrast with the page in the background.

cardosogabriel, line 116: @dlabrecq , this is because I need to overwrite a declaration from patternfly: .navbar-pf-vertical .nav .nav-item-iconic, which defines the padding.

andresgalante, line 116: @dlabrecq good point. It needs the nesting though to add more specificity to the selectors.

tests-src/notifications-drawer.html
```diff
@@ -0,0 +1,28 @@
...
+    $(".row-cards-pf > [class*='col'] > .card-pf > .card-pf-body").matchHeight();
+    $(".row-cards-pf > [class*='col'] > .card-pf > .card-pf-footer").matchHeight();
+    $(".row-cards-pf > [class*='col'] > .card-pf").matchHeight();
+
+    // initialize tooltips
+    $('[data-toggle="tooltip"]').tooltip();
+
+    // Initialize the vertical navigation
+    $().setupVerticalNavigation(true);
+  });
```
jeff-phillips-18, line 27: I believe you also need:

  $(document).ready(function() {
    $('#fixed-accordion').initCollapseHeights();
  });

in order to get the body of the accordion to scroll rather than the drawer.

cardosogabriel, line 27: I did it, but it doesn't work well, since the height is fixed in this component. It doesn't provide a good experience.
Thing a good approach would be to use the full available height to allow scroll. That would require some javascript to calculate it based on the user's display height and items of the drawer itself. I was expecting that you could do something like that, since you are the JS expert ;)

jeff-phillips-18, line 27: Then I'm OK with adding it in a follow-up PR.

andresgalante, line 27: @cardosogabriel @jeff-phillips-18 lets think about performance when we do it. Measuring heights with js tend to be poor.

less/notifications-drawer.less
Rohoover, line 26: FYI - At some time in the future I would expect the rows within the notification drawer to be clickable. This will require them to have a hover state. This is specific to the content, not the accordion tabs. 

I'm uploading the original conceptual design as reference. Please DO NOT follow this document, as it is old and the additional drawer functionality was scoped out until a later (undetermined) time.

[PTNFLY 640 - Event_Task Drawerv2.pdf](https://github.com/patternfly/patternfly/files/266368/PTNFLY.640.-.Event_Task.Drawerv2.pdf)

Also here is a visual representation of the hover:
patternfly_toastnotificationtray_hover

andresgalante, line 26: @cardosogabriel I know this is not a line comment, but can you please add the hover color for rows like @Rohoover suggested?

cardosogabriel, line 26: @Rohoover , do you want the hover state to be added now or in the future?

Rohoover, line 26: @cardosogabriel @andresgalante We might want to wait until the rows are clickable before we add hover. Thoughts?

andresgalante, line 26: @Rohoover 👍 
@LHinson please note that we are missing the hover styles.

Rohoover, line 26: Additional note: The Loading More belongs within the Tab's child area, not at the bottom of the drawer. It appears when a user scrolls down, it notifies the user that more content is being fetched (if applicable).

cardosogabriel, line 26: @Rohoover , yes, this will be update in my next push. Thanks!

cardosogabriel, line 26: @Rohoover , @andresgalante , regarding the hover, for now I added a modifier in the html where the user can choose if he wants hover or not.
div class="drawer-pf drawer-pf-notifications-clickable
This class "drawer-pf-notifications-clickable" will make the hover appear.

cardosogabriel:
@andresgalante, add hover and rebased!

less/notifications-drawer.less
```diff
@@ -0,0 +1,113 @@
...
+
+.drawer-pf-notification-info,
+.drawer-pf-notification-message {
+  display: block;
+  padding-left: 27px;
+  padding-right: 19px;
+}
+
+.drawer-pf-notifications-clickable .drawer-pf-notification:hover {
+  cursor: pointer;
```
andresgalante: @cardosogabriel the `drawer-pf-notifications-clickable` to optin hovers is a great idea!

But the default should be with hovers. Can you make it the other way around, so that the default is with hovers, but the user can optout using them please?

Also remove the cursor pointer in case the text of the notification has links.

Thanks! this is awesome.

andresgalante:
@jeff-phillips-18 When I open the last kebab drop menu triggers the scroll in a very strange way. Do you think we can solve it? If this is not related to js let me know and I can work with @cardosogabriel to solve it.
screen shot 2016-05-20 at 9 32 38 am

andresgalante:
@cardosogabriel: @serenamarie125 noticed that the icons are smaller than it should. I've measure them and checked the code and I can't see anywhere changing the size of the icons for the drawer. Maybe I am missing something or we should do something on the icons in general to force them to the correct size. 
What do you think?
screen shot 2016-05-20 at 9 40 04 am

screen shot 2016-05-20 at 9 37 00 am

less/notifications-drawer.less
cardosogabriel: @andresgalante: Thanks. I'm making it the opposite.

andresgalante:
@cardosogabriel @jeff-phillips-18 thanks guys, awesome work.

PR #287, opened 2016-05-16
Modified vert-nav test so code does is not duplicated

dlabrecq:
Modified vert-nav test so code does is not duplicated in multiple places.
For example, both vert-nav examples, as well as the overview/code tabs of patternfly-org.


jeff-phillips-18:
LGTM

andresgalante:
@dlabrecq 👍  thanks!

PR #310, opened 2016-06-08
Add input-border-focus to match border shadow color.

dverbeek84:
This merge will fix the focus color of the form inputs.
Otherwise the bootstrap color is used.


andresgalante:
@dverbeek84 Thanks for the fix.

@jeff-phillips-18 @dabeng Can you please review this one?

jeff-phillips-18:
LGTM

PR #315, opened 2016-06-16
Add option to disable links

infraredgirl:


less/links.less
```diff
@@ -0,0 +1,4 @@
+a.disabled {
+  pointer-events: none;
+  color: @gray;
```
andresgalante: Thanks for the PR!
Can you please put the declarations in alphabetical order? `color` first, then `pointer` and also change the color to `@color-pf-black-500`

andresgalante:
What do you think if we also add a `cursor: not-allowed` ?

less/links.less
```diff
@@ -0,0 +1,4 @@
+a.disabled {
```
andresgalante, line 1: ok, this is a question to you: what if we make it just `.disabled` without the anchor, so we can use the same modifier in other elements? does it makes sense?

andresgalante:
@jeff-phillips-18 @srambach Can you guys please review this one?

less/links.less
jeff-phillips-18: Bootstrap also sets the cursor to 'not-allowed', do we want to do that here as well?

less/links.less
infraredgirl, line 1: I think most other elements (buttons, etc) already have a mechanism for disabling them, links are an exception here rather than a rule. Furthermore, I'm not sure it makes sense to make it completely general, not all elements can be disabled. Also the implementation of the class is specific to links (pointer-events, etc) so it may not make sense in case of some elements.

I would suggest to leave the class specific to links for now, we can always refactor later if the need arises.

andresgalante, line 1: That makes sense @infraredgirl

andresgalante:
@infraredgirl This is great, thanks a lot 😄 Just one more thing, can you please add this class to the card layout in any of the links, so we have a way to demo it?

@LHinson FYI we are adding a disable state for links

infraredgirl:
Done

srambach:
Looks good, thanks!

tests/cards.html
```diff
@@ -808,7 +808,7 @@ <h2 class="card-pf-title">
                 </ul>
               </div>
               <p>
-                <a href="#" class="card-pf-link-with-icon">
+                <a href="#" class="card-pf-link-with-icon disabled">
                   <span class="pficon pficon-flag"></span>View CPU Events
                 </a>
               </p>
```
jeff-phillips-18, line 814: Actually, this file is generated. You can find the source files for the cards html in tests-src/_includes/widgets/cards

If you are looking for this particular card, it is in tests-src/_includes/widgets/layouts/cards.html

jeff-phillips-18:
Please include the built tests files as well. Thanks!

jeff-phillips-18:
LGTM. Thanks @infraredgirl !

less/links.less
```diff
@@ -0,0 +1,5 @@
...
+  color: @color-pf-black-500;
+  cursor: not-allowed;
```
jtomasek: You can use cursor: @cursor-disabled;

PR #318, opened 2016-06-17
Add option to disable links

infraredgirl:


less/links.less
```diff
@@ -0,0 +1,5 @@
+a.disabled {
```
jtomasek, line 1: using
`a.disabled, a[disabled] {`
gives a nice alternative of using an disabled attribute to set the anchor disabled. Although doing so is not html valid, but anyway...:)

dist/css/patternfly-additions.css
```diff
@@ -3325,6 +3325,11 @@ table.datatable th:active {
 .layout-pf-alt.layout-pf-alt-fixed-with-footer body {
   padding-bottom: 37px;
 }
+a.disabled {
+  color: #8b8d8f;
+  cursor: not-allowed;
+  text-decoration: none;
+}
```
jeff-phillips-18: I believe we should also turn off the text decorations on hover so the underline is not shown.

jeff-phillips-18:
LGTM

less/links.less
andresgalante, line 1: @jtomasek good point but there is no disabled attribute on anchors on html, I'll merge it and we can iterate on it later.

@akurdyukov thanks a lot!

PR #320, opened 2016-06-21
Fix-bug: datatables' global search based on regexp doesn't work

dabeng:
In datatables of patternfly, regular expression global search feature isn't enabled.


andresgalante:
@jeff-phillips-18 can you review this one please?

dist/js/patternfly.js
```diff
@@ -89,14 +89,19 @@
...
             "<li>Clear the search field, then click Search to return to the 20 most recent records.</li>" +
           "</ul>",
         "sSearch": ""
       },
-      "sPaginationType": "bootstrap_input"
+      "sPaginationType": "bootstrap_input",
+      "oSearch": {
+        "sSearch": "",
+        "bRegex": true,
+        "bSmart": false
```
dtaylor113, line 103: Why is bSmart set to false?  "When "bSmart" DataTables will use it's smart filtering methods (to word match at any point in the data), when false this will not be done."

jeff-phillips-18:
Please squash your commits.

dist/js/patternfly.js
dabeng, line 103: Hi @dtaylor113 , To avoid confliction between smart search and regex search, I set bSmart to false. For more details, please refer to https://datatables.net/reference/api/search()

dabeng:
Hi @jeff-phillips-18 , I have squashed the commits into single.

dist/js/patternfly.js
dabeng, line 103: Hi @dtaylor113 , you could also comprehend the above settings after you read the description  - DataTables' built-in filtering is "smart" in that it breaks the user's input into individual words and then matches those words in any position and in any order in the table (rather than simple doing a simple string compare).

Although this can significantly enhance usability of the filtering feature, it uses a complex regular expression to perform this task, and as such it can interfere with a custom regular expression input if you enable that option (search.regex). As such, this option is provided to disable this smart filtering ability on this page(https://datatables.net/reference/option/search.smart)

andresgalante:
@dabeng I am sorry it took me so long to get back to this PR, Can you solve the conflicts please?

PR #322, opened 2016-06-22
PTNFLY-1009 About Modal.

cardosogabriel:


andresgalante:
Here is a link to rawgit: https://rawgit.com/cardosogabriel/patternfly/PTNFLY-1309/tests/about-modal.html

tests-src/about-modal.html
```diff
@@ -0,0 +1,45 @@
+---
+categories: [Widgets]
+layout: page
+title: About Modal
+resource: true
+---
+      <button class="btn btn-default" data-toggle="modal" data-target="#myModal">Launch about modal</button>
+      <div class="modal fade" id="myModal" tabindex="-1" role="dialog" aria-labelledby="myModalLabel" aria-hidden="true">
+        <div class="modal-dialog">
+          <div class="modal-content pf-about-modal">
```
andresgalante: @cardosogabriel we don't prefix class, we postfix them.

tests-src/about-modal.html
```diff
@@ -0,0 +1,45 @@
...
+            <div class="modal-header">
+              <button type="button" class="close" data-dismiss="modal" aria-hidden="true">
+                <span class="pficon pficon-close"></span>
+              </button>
+            </div>
+            <div class="modal-body">
+              <h1>Product Title</h1>
+              <div class="pf-product-versions">
```
andresgalante: @cardosogabriel same here, `class-name-pf`

dist/css/patternfly-additions.css
```diff
@@ -5764,4 +5764,43 @@ a.disabled {
 .force-hide-secondary-nav-pf .persistent-secondary .nav-pf-persistent-secondary {
   display: none !important;
 }
+.pf-about-modal {
+  background-color: #292e34;
+  background-image: url(../img/about-modal-bg.png);
+  background-position: right bottom;
+  background-repeat: no-repeat;
+  background-size: 216px auto;
```
andresgalante: Can't we just make the image the correct size instead of resizing it?

less/about-modal.less
```diff
@@ -0,0 +1,42 @@
+//
+// About Modal
+// --------------------------------------------------
+
+.pf-about-modal {
+  background-color: @color-pf-black-900;
+  background-image: url(../img/about-modal-bg.png);
+  background-position: right bottom;
+  background-repeat: no-repeat;
+  background-size: 216px auto;
```
andresgalante, line 10: Can't we just make the image the correct size instead of resizing it?

less/about-modal.less
```diff
@@ -0,0 +1,42 @@
...
+  @media (min-width: @screen-sm-min) {
+    background-size: auto;
+  }
+  .modal-body {
+    color: @color-pf-white;
+    padding-bottom: 16px;
```
srambach, line 16: Can the padding be a relative measure?

less/about-modal.less
```diff
@@ -0,0 +1,42 @@
...
+    padding-left: @grid-gutter-width;
+    @media (min-width: @screen-sm-min) {
```
andresgalante, line 19: Let's make it mobile first.

less/about-modal.less
```diff
@@ -0,0 +1,42 @@
...
+  background-position: right bottom;
+  background-repeat: no-repeat;
+  background-size: 216px auto;
+  @media (min-width: @screen-sm-min) {
+    background-size: auto;
+  }
+  .modal-body {
+    color: @color-pf-white;
+    padding-bottom: 16px;
+    padding-left: @grid-gutter-width;
```
andresgalante, line 17: I'd also change the padding on the right in case the text is large.

less/about-modal.less
```diff
@@ -0,0 +1,42 @@
...
+}
+
+.pf-product-version .label {
+  padding-left: 0;
+  padding-right: (@grid-gutter-width / 4);
+}
+
+.pf-product-versions {
+  margin-bottom: 30px;
+  margin-top: 30px;
```
srambach, line 34: Also here, can the margin be a relative measurement? (2em, for instance?)

less/patternfly-additions.less
```diff
@@ -54,3 +54,4 @@
 @import "toast.less";
 @import "toolbar.less";
 @import "vertical-nav.less";
+@import "about-modal.less";
```
andresgalante: please try to keep it alpha order

tests-src/about-modal.html
andresgalante: @cardosogabriel What do you think if we use a horizontal definition list? http://getbootstrap.com/css/#horizontal-description

less/about-modal.less
cardosogabriel, line 10: It is the correct size for bigger screens. I resized down for mobile.

less/about-modal.less
cardosogabriel, line 16: As far as I know we are not using relative measures in patternfly: http://codeguide.patternfly.org/#css-syntax
Probably because it is based on Bootstrap 3, which uses absolute unities.
Bootstrap 4 will support rem and ems though: http://blog.getbootstrap.com/2015/08/19/bootstrap-4-alpha/

less/about-modal.less
cardosogabriel, line 19: Isn't it mobile first? First declare for mobile, and then for higher screens (screen-sm-min) ?

less/about-modal.less
cardosogabriel, line 34: Same response as above.

less/patternfly-additions.less
cardosogabriel: Yes, I got confused with this order:
@import "variables.less";
@import "mixins.less";
@import "blank-slate.less";

tests-src/about-modal.html
cardosogabriel: My bad 👎 
Fixed.

tests-src/about-modal.html
cardosogabriel: It generates more code, since I need to overwrite the default visual presentation, that is different. It is more consistent with bootstrap way of writing though.

less/about-modal.less
cardosogabriel, line 17: Done!

less/about-modal.less
srambach, line 16: OK, I understand. I'm new to Bootstrap and PF both so I will wait for Bootstrap 4. :-)
Also, I don't see anything in the code guide specifically talking about units; am I missing it?

cardosogabriel, line 16: It is not explicit in this document, but I confirmed with Andres. Only the exemple shows that we use px:
margin-bottom: 15px;
padding: 15px;

tests-src/about-modal.html
```diff
@@ -0,0 +1,39 @@
...
+        <div class="modal-dialog">
+          <div class="modal-content about-modal-pf">
+            <div class="modal-header">
+              <button type="button" class="close" data-dismiss="modal" aria-hidden="true">
+                <span class="pficon pficon-close"></span>
+              </button>
+            </div>
+            <div class="modal-body">
+              <h1>Product Title</h1>
+              <div class="product-versions-pf">
```
andresgalante, line 18: @cardosogabriel Is this div necessary? maybe we can just add a modifier to the `<dl>`

less/about-modal.less
```diff
@@ -0,0 +1,50 @@
...
+
+.product-versions-pf {
+  margin-bottom: 30px;
+  margin-top: 30px;
+  .dl-horizontal {
+    dd {
+      margin-left: 0;
+    }
+    dt {
+      float: left;
```
andresgalante: the `.dl-horizontal` is already floating left.

less/about-modal.less
```diff
@@ -0,0 +1,50 @@
...
+    color: @color-pf-white;
+  }
+}
+
+.product-versions-pf {
+  margin-bottom: 30px;
+  margin-top: 30px;
+  .dl-horizontal {
+    dd {
+      margin-left: 0;
```
andresgalante: @cardosogabriel there is a better way to overwrite it. if you do this you don't allow to have more than 2 lines of text on the dd, it will lose alignment. Ping me and lets have a chat to see how to better approach this.

less/about-modal.less
```diff
@@ -0,0 +1,44 @@
+//
+// About Modal
+// --------------------------------------------------
+
+.about-modal-pf {
+  background-color: @color-pf-black-900;
+  background-image: url(../img/about-modal-bg.png);
```
andresgalante: @cardosogabriel just one more thing. Can we make this image a variable on variables.less so its easier to change?

tests-src/about-modal.html
```diff
@@ -0,0 +1,37 @@
...
+                <ul class="list-unstyled">
+                  <li><strong>Label</strong> Version</li>
```
srambach, line 20: If this were a `<dl>`, then the labels could be `<dt>s` and the versions  `<dd>s` and it would be more semantic. You might also want to think about making the versions line up even if the labels are different widths.

cardosogabriel, line 20: Interesting. That was the previous version that we implemented. @andresgalante suggested this current code, perhaps he can clarify with you, but from my understanding the current approach is more generic and requires less css code.
Regarding the alignment, it was specified by design a margin after the label, and not alignment. I believe this could be revisited in the future if we feel the need.

andresgalante, line 20: @cardosogabriel sorry took me so long to come back to this PR. I'll merge it now.

@srambach I'll merge it for now. We can iterate on the list later. `<dl>` would be more semantic but we try them and when end up floating to generate spaces and brake lines. Having fixed width is also a good idea. Let's merge it for now and iterate on it, and send a new PR to improve it.

@srambach @cardosogabriel thanks a lot!!

PR #328, opened 2016-07-01
Notification Drawer for Horizontal Nav.

cardosogabriel:


dlabrecq:
I'm looking at the notification drawer example on rawgit and see a problem with the kebabs. When I click on a kebab for the last notification, in any section, the kebab is truncated/hidden from view. For some notifications, the kebab menu items are barely visible. Seems this needs to be on its own layer or should be scrolled into view?

Note that I've already scrolled to the bottom of the notification. After clicking the kebab, the user is forced to scroll, once again, to move the menu items into view.

Also, when clicking "Mark All Read", I'm expecting the "5 New Events" label to change?

andresgalante:
Adding link to rawgit: https://rawgit.com/cardosogabriel/patternfly/PTNFLY-1460/tests/notification-drawer-horizontal-nav.html

cardosogabriel:
@dlabrecq : Yes, this Kebab behavior is an issue we are aware. A fix for this would be to show the kebab above, but z-index didn't work. Another approach would be to scroll down when the kebab is opened. This might require some jquery I guess. IMO this is not an impediment to release though.
Regarding the "mark all read", this will be added in the angular.

dlabrecq:
Developers will be looking at this for examples, especially if this feature shows up on the new web site. If we cannot figure it out, and provide a working solution, how can we expect others will?

IMO, adding the "mark all read" functionality in Angular only isn't a good solution. (Not everyone uses Angular.) I'm certain we can use JavaScript to update/hide the "5 New Events" label.

cardosogabriel:
Since I'm not a JavaScript expert, could you provide such solution?
Regarding the Kebab issue, wdyt @andresgalante @LHinson ?

dlabrecq:
Who wrote the jQuery below? Found this in nav-horizontal-notification-drawer.html.

// Mark All Read
$('.panel-collapse').each(function (index, panel) {
  var $panel = $(panel);
  $panel.on('click', '.drawer-pf-action .btn', function() {
    $panel.find('.unread').removeClass('unread');
  });
});

I'm not able to test this modification, but you might consider changing the code above to this:

$('.panel-default').each(function (index, panel) { // Get parents with this class.
  var $panel = $(panel);
  $panel.on('click', '.drawer-pf-action .btn', function() {
    $panel.find('.unread').removeClass('unread');
    $panel.find('.panel-counter').addClass('hide'); // Hide "5 New Events" label
  });
});

If that doesn't work, this may be an alternative. Again, I haven't tested it.

$('.panel-collapse').each(function (index, panel) {
  var $panel = $(panel);
  var $parent = $(panel).parentElement; // Get parent element
  $panel.on('click', '.drawer-pf-action .btn', function() {
    $panel.find('.unread').removeClass('unread');
    $parent.find('.panel-counter').addClass('hide'); // Hide "5 New Events" label
  });
});

andresgalante:
Hi @dlabrecq I think @jeff-phillips-18 did, but I am not sure.

andresgalante:
@dlabrecq @jeff-phillips-18 @priley86 we need your help to review and improve the js on this one so @cardosogabriel can move on

jeff-phillips-18:
@dlabrecq 's comment is simply to add the functionality to remove the "5 Unread Events" label. I'm not sure that needs to be done here, different applications may want to say "0 Unread Events" rather than hide the label. That label would be driven by backend data that would get updated after the Clear All button is pressed. I'm OK with this the way it is without the change.

LGTM

dlabrecq:
I was also thinking "0 Unread Events" would be better Vs hiding. @cardosogabriel did you try the code?

If this were only a test page, I'd be fine with it as is. However, I think including this on pf.org will look like a bug. I'll defer to @LHinson here.

jeff-phillips-18:
@cardosogabriel  Can you please update the javascript to include the bolded line below?

```
// Mark All Read
$('.panel-collapse').each(function (index, panel) {
  var $panel = $(panel);
  $panel.on('click', '.drawer-pf-action .btn', function() {
    $panel.find('.unread').removeClass('unread');
    **$(panel.parentElement).find('.panel-counter').text('0 New Events');**
  });
});
```

I've tested this and it works fine. 

edit: bold doesn't seem to work, the line with *\* around it ;)

cardosogabriel:
@dlabrecq , @jeff-phillips-18 , thanks, the code works. Rebased the branch.

I agree that "0 New Events" is good when the user has "5 New Events" and click on Mark all as read. 
However, from usability perspective, if some of the tabs initially has no new events, not displaying the label is better, so there is a visual difference between the tabs that has new events vs the ones with no new content. But I believe this is more a recommendation that should be in the "Design" tab of the site. WDYT @LHinson ?

Can you review and merge if we are good?

LHinson:
@cardosogabriel the rawgit link above doesn't work anymore. Will you provide another link? Want to look at it before commenting.

jeff-phillips-18:
@LHinson try https://rawgit.com/cardosogabriel/patternfly/PTNFLY-1460/dist/tests/notification-drawer-horizontal-nav.html

jeff-phillips-18:
I believe the notification drawer is not supposed to scroll with the page, it should be fixed at the top and bottom.

Rohoover:
@cardosogabriel 
@jeff-phillips-18 

Regarding "0 New Notifications" - I think that is a good solution.

LHinson:
I've also confirmed with @Rohoover that the notification drawer should stay attached to the icon so the way it is seen today is the intended design. @jeff-phillips-18

LHinson:
Alright, eating my words here a bit. I understand what @jeff-phillips-18 is saying now. The notification drawer for vertical nav (https://rawgit.com/patternfly/patternfly/master/dist/tests/notifications-drawer.html) is inconsistent with this one. 

I think the approach should be consistent. In the version for vertical nav, the masthead is sticky. So it makes sense that the drawer sticks too. The user can easily collapse or expand the drawer no matter where they are on the page. 

@Rohoover what are your thoughts re: making the masthead and horizontal nav stick here too so that the notification drawer can stick.

Rohoover:
That sounds good @LHinson 

The drawer is attached to that icon, in that it allows the toggle of the drawer. I think making sure those two are within proximity is the more important theme.

LHinson:
After reviewing at the PatternFly meeting, we are in a agreement that both the masthead and the horizontal nav should stick therefore so does the notification drawer

cardosogabriel:
Currently the behaviour of our Horizontal Navigation is to not stick: https://www.patternfly.org/pattern-library/navigation/horizontal-navigation/#_overview

I built the Notification Drawer for Horizontal Navigation based on the Horizontal Navigation layout page: https://www.patternfly.org/pattern-library/navigation/horizontal-navigation/horizontal-navigation.html#_

@LHinson , just to clarify, are you saying that the horizontal navigation should always stick? If so, I suggest to create a new story to update the horizontal navigation pattern and make the horizontal nav bar stick. We probably need to update the "design" tab of the website also explaining this. It is better to fix this component first and then update the Notification Drawer, since it is based on that one.

jeff-phillips-18:
@cardosogabriel @LHinson  I'm OK with adding this as is and adding an issue/story to update all horizontal nav pages to use sticky masthead and navigation bars.

cardosogabriel:
Yes, I this is a better approach. Also easier in terms of development.

LHinson:
Agreed! Thank you @cardosogabriel @jeff-phillips-18  

I've made a new epic for these changes - https://patternfly.atlassian.net/browse/PTNFLY-1646

jeff-phillips-18:
LGTM

jeff-phillips-18:
@dtaylor113 @dlabrecq Can we get this reviewed and merged?

dlabrecq:
LGTM

dtaylor113:
LGTM

cardosogabriel:
Great, thanks all for all the help!

PR #334, opened 2016-07-06
Adds Wizard component

srambach:
This adds the Wizard component. 
https://rawgit.com/srambach/patternfly/wizard-1243/dist/tests/wizard.html [fixed url 7/18/16]
The wizard is always open now to make it easy to compare the versions, but I will remove the debug code to make them pop up from the buttons before we close this PR.
Could you please review? @andresgalante @jeff-phillips-18 @dlabrecq @LHinson @mwmahlberg Thanks!


srambach:
@sjcox-rh, please also review this. Thanks!

dtaylor113:
LGTM

dlabrecq:
1. We need to create templates for these examples, so they may be included in the patternfly web site. I can point you to examples in the existing test pages.
2. Should we also provide functionality to navigate between pages? Developers are going to look for examples, especially on the patternfly web site, so it would be great to see a fully working wizard example that users can navigate through.

andresgalante:
@dlabrecq I agre that it would be good to have the functionality, but the idea for this first implementation was to have at least the html/css on base patternfly so @jeff-phillips-18 can use it on his angular implementation.
There are other components without functionality like the pagination, where we show all the different states but we don't make it work.
Do you think we can leave the functionality for later?

dlabrecq:
Will this example be added to the patternfly web site? If not, I'm fine with adding functionality later.

It would still be good to create templates for the patternfly web site, though. If this ends up on the web site, we will need this and the example should be functional.

mwmahlberg:
I wholeheartedly agree – nothing more frustrating than „This is what you could do with it – if you knew how. We just do not explain it (yet).“

LHinson:
I'll add another story to the backlog to address @dlabrecq 's point about #1 and #2. That way this PR can be merged (assuming there are no other concerns) and @jeff-phillips-18 can run with the html/css for angular-pf. We won't add the code to the site until we have addressed the templates and added functionality. Any concerns from anyone re: this approach?

srambach:
I like that solution (adding a story). I agree that it can be frustrating
to find the design but not everything you need to plug it in.

LHinson:
Check out https://patternfly.atlassian.net/browse/PTNFLY-1568 and let me know if you have comments/edits to the new story @srambach @dlabrecq

sjcox-rh:
@srambach This looks great!  A couple small things.  There isn't supposed to be a green outline around the primary steps and the Back button on the Initialization page and first step should be disabled.  Other than that, good to go!

srambach:
@sjcox-rh You are absolutely correct about the green outline - that is not on my local copy so I believe I have a problem in pushing the css file. I will investigate this.
Also, I will correct the button disabling.
I don't think the design specified the amount of white space as padding around the main content area. Would you like to give me a specific amount? I believe I have set it at 2ems temporarily.

less/wizard.less
```diff
@@ -0,0 +1,330 @@
+//
+// Wizard
+// --------------------------------------------------
+
+// Wizard header
+// Top section of the wizard w/ title and dismiss
+
+// variables
+@wz-default-bg-color: @color-pf-white;
```
andresgalante: @srambach Why did you created new variables instead of just using the original ones?

srambach: @andresgalante I had no especially good reason for the new variables other than that some had already been defined hard coding the color values, so I changed them to use the predefined colors and added others as they came up. I am happy to remove the extra variables.

sjcox-rh:
@srambach Thanks!!

As for padding, the padding around the review step works better.  Is that 2ems?

The padding for the first step is not enough, makes the wizard look cluttered.

less/wizard.less
```diff
@@ -0,0 +1,305 @@
...
+    }
+
+    a {
+      color: @color-pf-black;
+      font-size: 16px;
+      cursor: pointer;
+      margin-left:1em;
+      margin-right:1em;
+      text-decoration: none;
+      transition: 0.25s;
```
andresgalante: We don't have a standard for animations yet, till we do I'd day we should remove animations, or add them everywhere 😄  consistency!

less/wizard.less
```diff
@@ -0,0 +1,305 @@
...
+
+      /* draw the step number in the circle */
+      &:before {
+        background-color: @color-pf-white;
+        border-radius: 100%;
```
andresgalante: a circle is 50%

less/wizard.less
```diff
@@ -0,0 +1,305 @@
...
+      &:hover {
+        &:before {
+          background-color: @wz-hover-bg-color;
+          border-color: @wz-hover-border-color;
+          color: @wz-hover-color;
+        }
+      }
+    }
+  }
+  li.current a {
```
andresgalante: Can we avoid adding the `li` on the selector? Also you can use an ampersand less selector like this on the previews declaration:

```
a{
  .current &{}
}
```

less/wizard.less
```diff
@@ -0,0 +1,305 @@
...
+.wizard-pf-review-page {
+  .list-group, .list-group-item {
+    border: none;
+    margin-bottom: 0;
+  }
+  ul {
+    margin-left: 24px;
+  }
+}
+.wizard-pf-review-step {
```
andresgalante: Why duplicate the selector from line 241?

tests-src/wizard.html
```diff
@@ -0,0 +1,448 @@
...
+      <div class="modal-content">
+        <div class="modal-header">
+          <button type="button" class="close" data-dismiss="modal" aria-label="Close"><span
+            aria-hidden="true">&times;</span></button>
+          <dt class="modal-title">Wizard Title</dt>
+        </div>
+        <div class="modal-body wizard-pf-body clearfix">
+          <div class="wizard-pf-steps">
+            <ul class="wizard-pf-steps-indicator">
+              <li class="viewed">
```
andresgalante: We need to think of a better name for the `viewed` class. If it's a patternfly class name then it needs a `-pf` posfix, and its it's a modifier of something we needs a very long and ugly name like `wizard-pf-steps-indicator-viewed` 💩 
What do you think?

tests-src/wizard.html
```diff
@@ -0,0 +1,448 @@
...
+                  <span class="wizard-pf-substep-number">3B.</span>
+                  <span class="wizard-pf-substep-title">Progress</span>
+                </a>
+              </li>
+            </ul>
+          </div><!-- /.wizard-pf-sidebar -->
+
+          <div class="wizard-pf-main">
+            <div>
+              <div class="wizard-pf-review-page">
```
andresgalante: Does it make sense to use an accordion here? https://rawgit.com/patternfly/patternfly/master/dist/tests/accordions.html

LHinson:
This looks good @srambach. I'll let you and @andresgalante work out the technical bits :) I did have one question... when I click the "Launch wizard...." button, my screen fades but no modal is actually launched. Not sure if anyone else i having this problem?

srambach:
@LHinson That behavior is an artifact of the modals being displayed inline for now, just for ease of reviewing them - before the PR is closed, I will turn that off so that they are not displayed all the time, but pop up when the button is clicked.

less/wizard.less
srambach: I will remove the li; also on the li.viewed. I'm not sure I understand what you are saying about the ampersand selector, however - which selector do you mean, and is there an advantage to doing it that way?

less/wizard.less
srambach: Do you mean .wizard-pf-review-**page** and .wizard-pf-review-**step**? Or that a .list-group-item may also be a .wizard-pf-review-step? (It may also be a .wizard-pf-review-substep)

tests-src/wizard.html
srambach: Long and ugly, but more descriptive. However, we also use .current right there...is that ok because it's the common convention?

tests-src/wizard.html
srambach: Yes, possibly - it bugged me that I felt I was reinventing a wheel that should have been there (an expanding/collapsing list). However, I don't think it's quite an accordion, because I think more than one could be open at a time, whereas an accordion only has one item open at a time.

less/wizard.less
srambach: Removed!

less/wizard.less
srambach: Fixed.

LHinson:
oh ok, got it! Thanks @srambach

srambach:
I've redone the review list to use a treeview. @sjcox-rh if you have time, could you please check the design of the review step? I was unable to reduce the font size for substeps. Also, there is now a hover highlight by default. I'm not sure if I can turn that off, but I will try if you don't want it there.
Also, @andresgalante @LHinson please take a look? Thanks!

srambach:
@sjcox-rh This is working now, so it's ready to review if you have time. Thanks!

tests-src/wizard.html
srambach: I changed this to a treeview to get the expanding/collapsing behavior.

sjcox-rh:
@srambach I reviewed the treeview.  It looks fine given the limitations you came across.  Would prefer the substep text to be small but if it's too difficult, it's fine the way it is.  Also, the hover states for the main steps are fine, but if you could turn off the hover state for the substeps, that would be great.  Again, if you can't, then it's okay the way it is.  However, if substeps are clickable, then keep the hover state.

Thanks!

srambach:
@sjcox-rh Unfortunately, for the same reason I can't make the text smaller on the sub-steps, I also can't turn off the hover on them.

srambach:
@andresgalante I have made the changes we talked about this morning. I have left the positioning on the circles in px because it opens a big can of works for IE9 to calculate it.

tests-src/wizard.html
```diff
@@ -0,0 +1,447 @@
...
+          <button type="button" class="close" data-dismiss="modal" aria-label="Close">
+            <span aria-hidden="true">&times;</span>
+          </button>
+          <dt class="modal-title">Wizard Title</dt>
+        </div>
+        <div class="modal-body wizard-pf-body clearfix">
+          <div class="wizard-pf-steps hidden">
+            <ul class="wizard-pf-steps-indicator">
+              <li>
+                <a><span class="number-pf">1</span><span class="title-pf"First Step</span></a>
```
jawnsy: Missing: >

tests-src/wizard.html
```diff
@@ -0,0 +1,447 @@
...
+              </li>
+              <li>
+                <a><span class="number-pf">2</span><span class="title-pf">Second Step</span></a>
```
andresgalante: @srambach sorry, but I dont really like`title-pf` and `number-pf` class names. 
Can we change them to `wizard-pf-steps-title` and `wizard-pf-steps-number` please?

less/wizard.less
```diff
@@ -0,0 +1,248 @@
...
+        font-size: 14px;
+        font-weight: 700;
+        height: 50px;
+        outline: 0;
+        padding-top: 11px;
+        padding-left: 20px;
+        position: relative;
+        white-space: nowrap;
+        width: 18em;
+
```
andresgalante: @srambach can we remove this blank line?

andresgalante:
@srambach Just point out 2 small changes the same look awesome, thanks a lot. Can you remove the conflicts so we can merge it please?

jeff-phillips-18:
Please squash the commits and can we see the final version of the test page without the inline wizard pages?

andresgalante:
@dabeng @jeff-phillips-18 the only thing missing is to move the code to the include in order to use it on the website. But since that's going to happen after we add the js, do you think we can merge it as it is?

@srambach In order to consume the snippet on the pf.org we need to move the html to the `_include/` folder and generate an include on the test page. The steps to achieve this are:
1.  Create a new folder folder in [/_includes/widgets](https://github.com/patternfly/patternfly/tree/master/tests/pages/_includes/widgets) called `wizard`
2. Add the html snippets, you have plenty of examples there
3. Go back to wizard.html on `pages/` and replace the html for the includes. This is an example of how a form example was included: https://github.com/patternfly/patternfly/blob/master/tests/pages/forms.html#L426

This is a step we should take every time in order to easily update pf.org. If you need help or you don't have time to do it, please let me know.

jeff-phillips-18:
My only concern is using Treeview even though it's limitations prevent us from obtaining the styling we believe should be on the nodes (substeps in particular). I'm not convinced that the small savings in code warrants these limitations.

srambach:
Treeview does have limitations, but I am not excited about reinventing a collapsing list for this wizard.  I really think Patternfly should have its own collapsing list because it's a pretty common element. What about leaving it here for now, since we already have the dependency of having Treeview in Patternfly, but try to get a replacement into the backlog soon? @jeff-phillips-18 @LHinson @andresgalante

serenamarie125:
@srambach, the text in the Review page does look way too large.  I'm assuming this is because of the treeview. @jeff-phillips-18 I wonder what the OpenShift Deployment Wizard would look like when switching to the PF version

jawnsy:
Yay! :tada: Thanks for the merge :+1:

What release version of PatternFly can we expect this in? And when will that be released?

andresgalante:
@jawnsy probably 3.7.0 and probably tomorrow or Wednesday 😄

PR #341, opened 2016-07-07
Moved the js src out of the dist folder into a new src folder

bleathem:
This PR introduces a `src` folder and moves the `patternfly.js` source into that folder.  I introduced a grunt task to copy it back from the `src` folder to the `dist` folder, in case we have people consuming the `patternfly.js` file (from the `dist` folder) instead of the `patternfly.min.js` file.  Since we don't yet provide js source maps, this is probably quite common.

@andresgalante pointed out that folks are likely consuming the _less_ files in downstream projects, so I refrained from moving the less files into the `src` folder.  We should track this in a jira and co-ordinate with an appropriate version bump (ie. either minor or major).

The test files should also be moved into the `dist` folder, to highlight that they are generated.  Given the impact that change will have on the patternfly.org site, I suggest we tackle that move in a separate PR.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -63,6 +63,12 @@ module.exports = function (grunt) {
           {expand: true, cwd: 'components/c3/', src: ['c3.css'], dest: 'less/lib/c3/'},
         ],
       },
+      js: {
+        files: [
+          // copy js src file
+          {expand: true, cwd: 'src/js/', src: ['patternfly.js'], dest: 'dist/js/'}
+        ]
+      }
     },
     csscount: {
       production: {
@@ -101,7 +107,7 @@ module.exports = function (grunt) {
     jslint: {
       client: {
         src: [
-          'dist/js/patternfly.js'
+          'src/js/patternfly.js'
         ],
         directives: {
           // node environment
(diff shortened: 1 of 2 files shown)


dlabrecq:
LGTM. If you also want to move the test files, I can easily update the web site.

dtaylor113:
LGTM...although I deleted the dist directory, expecting a grunt build to recreate all the folders in dist, however only css and js were created -no dist/fonts or dist/imgs.  There are grunt tasks for copying the fonts (from bootstrap and font-awesome dependencies), not sure why grunt build didn't execute the font grunt task.  I do not see any grunt tasks for dist/imgs, so are these imgs/svgs directly checked into dist?   If so, sounds like we should have a resources dir which we copy to dist.

bleathem:
+1 @dtaylor113.  I suggest we address that in a subsequent PR.  Let's get this merged so we can build off of it in https://github.com/patternfly/patternfly/pull/342

dtaylor113:
Added JIRA story to capture remaining work:

PatternFly base ref. GitHub repo. source code re-org - Dev
https://patternfly.atlassian.net/browse/PTNFLY-1576

PR #347, opened 2016-07-11
Move Jekyll test templates to tests->pages

priley86:
*updates gruntfile
*adds redirect for anyone with old link bookmarked
*also fixes issue with unit tests compiling to dist (they are not needed there)


dlabrecq:
The files under _includes are what we have been referring to as our "templates". I don't know if "templates" is the right choice for a directory name for all tests?

I had thought we planned to place everything under src, similar to angular-patternfly?

Can we move this to src or src/tests, instead? We can still create a redirect, but it may be best if both repos were similar. Folks continuously ask where the source is, so it may become clear if it were actually under src?

dtaylor113:
Unfortunately, the test _includes/templates are NOT source code, we have specific warnings on those pages saying "These examples are included for development testing purposes", so I think it is appropriate that they live in a root test dir.   The only "source" code we distribute from base PF repo is patternfly.js, and .css files.
As for whether to call them "tests/_includes" or "tests/templates", I'd do whatever is standard in Jekyll.

dtaylor113:
My vote to be crystal clear would be to rename tests/templates to tests/pages, since everyone refers to them as 'test pages'

bleathem:
I feel there is merit in keeping the test source distinct from the "project source", similar to how the jQuery repo is structured.  An alternative is to adopt a repo structure like what bootstrap is using, where everything is at the top level, but I don't think this maps well to the complexity we have in patternfly.  As such I recommend we keep the `/src` and `/tests` folders distinct.

Instead of `/tests/templates`, how about `/tests/pages`?

priley86:
@dtaylor113 @bleathem I like the suggestion for `/tests/pages` vs `tests/templates`...updating my PR

bleathem:
Looking at the folder structure (https://github.com/priley86/patternfly/tree/tests-templates/tests), everything seems good.

dlabrecq:
LGTM.

Issue #333, opened 2016-07-06
I wanna fix the bug in datepicker of patternfly but where is its source code


dabeng:
Hi @andresgalante .I wonder which place I can commit my pull request about datepicker.

andresgalante:
I am a bit lost here, whats the issue you want to fix?

dabeng:
![screenshot from 2016-07-07 08-05-24](https://cloud.githubusercontent.com/assets/5689075/16638456/0dc5a400-43ad-11e6-8848-514cc034cc3c.png)
![screenshot from 2016-07-07 08-06-25](https://cloud.githubusercontent.com/assets/5689075/16638460/12ec4402-43ad-11e6-91b5-621d5b7b82b3.png)

You could get the above two snapshots easily if you clicks calendar icon firstly and then clicks input box. I think it's a intolerable bug for users.

dlabrecq:
This particular example is included via a template from patternfly. The source lives in tests-src/widgets/forms/bootstrap-datepicker.html.

I noticed that the input element is not closed in the template, but I also don't see this issue with the patternfly test pages (i.e., https://rawgit.com/patternfly/patternfly/master/tests/bootstrap-datepicker.html).

dabeng:
![screenshot from 2016-07-07 09-44-23](https://cloud.githubusercontent.com/assets/5689075/16640080/cac773f0-43ba-11e6-97f6-4f6b90db93f0.png)

Dear @dlabrecq , could you find that the shadow becomes double dark.

dabeng:
Redundant dropdown menu of datepicker, it does indeed exist.

andresgalante:
@dabeng thanks for the clarification and the detail information. We do have a bug and a strange implementation on the website.
If this is a bug in our implementation, do you want to send a PR to fix it or want us to add it to our back log?

And if it is a bug in the component itself, maybe we can send a PR to the upstream project and update our dependency when they merge it. What do you think?

dlabrecq:
Yes, I see the highlighting darkening as well.

dabeng:
Hi @andresgalante , as a matter of course, I'm glad to send  PR for the bug I found.

dabeng:
Hi @dlabrecq . Could you tell me how to find the folder " tests-src" and I want to know why I got the unrendered page as shown below  when I access test page directly on rawgit.com.

[...]

PR #348, opened 2016-07-12
Fix-bug: there are repeated dropdown menus of datepicker

dabeng:
The [issue](https://github.com/patternfly/patternfly/issues/333) is caused by repeated instantiation of datepicker controls. Now the datepicker with icon no longer pops up two dropdown menus.

### DIFF
--- a/dist/tests/bootstrap-datepicker.html
+++ b/dist/tests/bootstrap-datepicker.html
@@ -87,21 +87,21 @@
       <h3>Validation States</h3>
       <div class="form-group has-success">
         <label class="control-label" for="boostrap-datepicker-success-readonly">With success</label>
-        <input type="text" class="form-control bootstrap-datepicker" id="boostrap-datepicker-success-readonly" readonly>
+        <input type="text" class="form-control bootstrap-datepicker input-datepicker" id="boostrap-datepicker-success-readonly" readonly>
       </div>
       <div class="form-group has-warning">
         <label class="control-label" for="boostrap-datepicker-warning-readonly">With warning</label>
-        <input type="text" class="form-control bootstrap-datepicker" id="boostrap-datepicker-warning-readonly" readonly>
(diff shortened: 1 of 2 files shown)


andresgalante:
👍  awesome, thanks @dabeng 

@srambach if you have a min, can you review this one please?

jeff-phillips-18:
I believe we should set an ID on the date picker component and use the ID in the JQuery rather than removing the CSS class. The CSS class is used for styling and is necessary.

dabeng:
@jeff-phillips-18 , thank you very much, I realized bootstrap-datepicker is useful. I updated PR according to your tip.

dabeng:
Can any body help review, it's so easy code modification while it fixed so conspicuous bug.

andresgalante:
@bleathem @dlabrecq @priley86  can you please review this one and merge it if its ok?

dist/tests/bootstrap-datepicker.html
```diff
@@ -129,7 +129,7 @@
       </div>
       <script>
         // Initialize Boostrap-Datepicker
-        $('.bootstrap-datepicker, .input-group.date, .bootstrap-datepicker-inline').datepicker({
+        $('.bootstrap-datepicker, .input-group.date, .bootstrap-datepicker-inline').not($('.input-group.date').find('.bootstrap-datepicker')).datepicker({
```
jeff-phillips-18: This JQuery statement is complex and hard to understand. I would rather see one div around the non-initialized datepickers (not the first three that are includes since they are already initialized) and use that div to help select the datepickers that need to be initialized.

dabeng: I got your point, new commit is coming.

jeff-phillips-18:
LGTM

dtaylor113:
LGTM

dtaylor113:
LGTM

PR #353, opened 2016-07-13
changes to bower

priley86:
Allow `tests/pages` to be included in bower distribution.

*Updates to README `test-src`  to `tests/pages`

Note: it would be ideal to only include `tests/pages/_includes` with a glob pattern like this:

```
"tests/pages",
"!tests/pages/_includes",
```

However it looks like bower currently does not support negated global patterns on directories and only individual files. 
https://github.com/bower/bower/issues/614 

@dlabrecq and I have decided to leave it like this for now.

@bleathem @andresgalante @dtaylor113 let us know if this works


dtaylor113:
Hi, should this PR include the new files under dist/tests/pages?
Or are you saying tests/pages will now be included when doing "bower install patternfly"?   If so, I don't think pf.org pulls from 'test/pages/_includes' -think it's expecting the files to be in '/dist/tests/pages/_includes'  @dlabrecq is this correct?  
Per @bleathem 's point, maybe we don't want to confuse our dist files with 'tests', so perhaps we should just update pf.org to pull from 'tests/pages/_includes"?

priley86:
@dtaylor113 i think @dlabrecq is planning to update pf.org to point at `test/pages/_includes`

dlabrecq:
pf.org had pulled from tests-src/_includes, so I'm expecting to pull from tests/pages/_includes. It can pull from either directory, but no need to be in dist. If/when we move to github subtrees, we can ignore tests in bower.json.

bleathem:
I've created a jira to track the pf.org move to submodules/subtree. patternfly.atlassian.net/browse/PTNFLY-1601

What about using npm now to pull the patternfly files into the site, rather than bower? That way we could leave the test artifacts hidden in the bower distribution.

dlabrecq:
If anyone else wants to implement another approach, that's fine. However, I'd rather not introduce npm; especially, if we plan on using github subtrees. There isn't much time left in this sprint to modify pf.org, retest Travis, etc.

This had been working with bower and want to get that going ASAP -- we will be cutting new releases soon. If there is time in this sprint, I may look at subtrees; although, I still have other stories to work on.

bleathem:
I wasn't implying neither when nor whom should introduce npm, merely proposing that we could consider it as an alternate means to achieve our goals.  Sounds like the bower solution is working well though, and we can leave it at that.

jeff-phillips-18:
LGTM

dlabrecq:
@bleathem Don't get me wrong, it's a good idea. I suppose whether it's npm or subtrees, we will want only one copy of patternfly in pf.org. (Not only does pf.org depend on patternfly itself, but angular-patternfly pulls it in as well.) I imagine changing all the various instances in pf.org that point to source/components/patternfly/*, for example. Just want to ensure we have adequate time to test that properly. I'm thinking we would also introduce a subtree to angular-patternfly in order to remove the extra patternfly dependency in pf.org?

bleathem:
Having angular-patternfly depend on patternfly (core) is a clean and reasonable dependency, and can be well managed via bower/npm. The patternfly.org site however isn't a library that we distribute, and so I think we can get away with simplifying the dependency on patternfly (core) via the git subtree.  It's at least worth exploring.  I've tracked it via [PTNFLY-1601](https://patternfly.atlassian.net/browse/PTNFLY-1601).

PR #354, opened 2016-07-14
Patternfly smoke tests

priley86:
- introduce Patternfly js libraries to Karma
- provide smoke test examples for running unit tests on Patternfly test fixtures
- removes any inline scripts from fixtures during Karma testing, and sources Karma scripts instead
- adds grunt-karma
- adds automated test examples for the following components:
  - About Modal
  - Accordion
  - Alerts
  - Area Charts
  - Badges
  - Bar Charts
  - Bootstrap Combobox
  - Bootstrap Datepicker
  - Bootstrap Select
  - Bootstrap Switch
  - Bootstrap Treeview
  - Datatables

### DIFF
--- a/tests/unit/bootstrap-datepicker.spec.js
+++ b/tests/unit/bootstrap-datepicker.spec.js
@@ -1,0 +1,32 @@
+describe("bootstrap datepicker test suite", function () {
+
+  beforeEach(function () {
+    globals.readFixture(globals.fixturePath + 'bootstrap-datepicker.html');
+  });
+
+  it('should open the first datepicker after show', function (done) {
+
+    var datePicker1 = $('#datepicker1');
+
+    datePicker1.datepicker({
+      autoclose: true,
+      orientation: "top auto",
+      todayBtn: "linked",
+      todayHighlight: true
+    });
+
+    datePicker1.datepicker('show');
+
+    setTimeout(function () {
+      expect($('.datepicker-dropdown')).toExist();
+
+      //close datepicker1
+      datePicker1.datepicker('hide');
+      expect($('.datepicker-dropdown')).not.toExist();
+
+      done();
+    }, globals.wait);
+
+  });
+
+});
--- a/tests/unit/bootstrap-treeview.spec.js
+++ b/tests/unit/bootstrap-treeview.spec.js
@@ -1,0 +1,53 @@
+describe("bootstrap-treeview test suite", function () {
+
+  beforeEach(function () {
+    globals.readFixture(globals.fixturePath + 'bootstrap-treeview.html');
+  });
+
+  it('should render a tree view with three parents and one child node', function (done) {
+
+    var defaultData = [
+      {
+        text: 'Parent 1',
+        href: '#parent1',
+        tags: ['1'],
+        nodes: [
+          {
+            text: 'Child 1',
+            href: '#child1',
+            icon: 'fa fa-file-o',
+            tags: ['0']
+          }
+        ]
+      },
+      {
+        text: 'Parent 2',
+        href: '#parent2',
+        tags: ['0']
+      },
+      {
+        text: 'Parent 3',
+        href: '#parent3',
+        tags: ['0']
+      }
+    ];
+
+    $('#treeview1').treeview({
+      collapseIcon: "fa fa-angle-down",
+      data: defaultData,
+      expandIcon: "fa fa-angle-right",
+      nodeIcon: "fa fa-folder",
+      showBorder: false
+    });
+
+    setTimeout(function () {
+      var nodes = $('#treeview1 ul li');
+      var indent = nodes.find('.indent');
+
+      expect(nodes).toHaveLength(4);
+      expect(indent).toHaveLength(1);
+      done();
+    }, globals.wait);
+  });
+
+});
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -194,6 +194,11 @@ module.exports = function (grunt) {
       options: {
         livereload: true
       }
+    },
+    karma: {
+      unit: {
+        configFile: 'karma.conf.js'
+      }
     }
   });
 
--- a/README.md
+++ b/README.md
@@ -142,7 +142,7 @@ The HTML pages in `tests/` are generated using Jekyll.  Do *not* edit these file
 Unit tests are written for [Karma test server] (https://karma-runner.github.io/1.0/index.html) with [Jasmine](http://jasmine.github.io/)
 
 ```
-npm test
+grunt karma
 ```
 ## Release
 
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -17,6 +17,12 @@ module.exports = function(config) {
     files: [
       'components/jquery/dist/jquery.js',
       'components/bootstrap/dist/js/bootstrap.js',
+      'components/bootstrap-combobox/js/bootstrap-combobox.js',
+      'components/bootstrap-datepicker/dist/js/bootstrap-datepicker.js',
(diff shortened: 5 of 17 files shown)


priley86:
@bleathem @dtaylor113 @dlabrecq @jeff-phillips-18  would love any feedback

priley86:
also - this links to [PTNFLY-1550](https://patternfly.atlassian.net/browse/PTNFLY-1550)

tests/unit/bootstrap-treeview.spec.js
```diff
@@ -0,0 +1,53 @@
...
+        tags: ['0']
+      },
+      {
+        text: 'Parent 3',
+        href: '#parent3',
+        tags: ['0']
+      }
+    ];
+
+    $('#treeview1').treeview({
```
dlabrecq, line 35: Curious about this test. There is already a treeview1 rendered in the page with it's own data. Do we want to test what's we've added to the page (i.e., in bootstrap-treeview.html) or render a completely new widget?

I see a mix of both between tests, so just wondering what's the recommended approach? (For example, we don't need to do create the widget in the badges test.) If we're rendering a completely new widget, is there any need to load individual pages like bootstrap-treeview.html?

tests/unit/bootstrap-datepicker.spec.js
```diff
@@ -0,0 +1,32 @@
...
+    var datePicker1 = $('#datepicker1');
+
+    datePicker1.datepicker({
+      autoclose: true,
+      orientation: "top auto",
+      todayBtn: "linked",
+      todayHighlight: true
+    });
+
+    datePicker1.datepicker('show');
```
dlabrecq: Similar question to treeview. There is already a datepicker1 rendered in the page with it's own data. Do we want to test what's we've added to the page (i.e., in bootstrap-datepicker-text-input.html) or render a completely new widget?

tests/unit/bootstrap-treeview.spec.js
priley86, line 35: @dlabrecq So I had a bit of trouble reusing our current test templates/html due to the fact that jquery and other plugins were loading twice. Because of this, I stripped out any page scripts in the html and simply used the html (see `tests/unit/globals.js`). Therefore all JS is in Karma or the unit test (so I essentially recreate different scenarios on the existing templates with new JS). 

I can certainly see it being feasible to create custom templates for various tests (and I would certainly be inclined to do this), but for now, I attempted to reuse existing templates. It seems to be working OK in this format now and I wanted to start our Karma testing with some templates we are already comfortable with.

priley86, line 35: @dlabrecq re-reading your comment, it looks like I didn't quite address it. I am OK with changing the treeview js test to match what's in `dist/tests/bootstrap-treeview.html` to be consistent. Let me know.

dlabrecq, line 35: Too bad we have to duplicate the JS. (Would prefer to test what's already in the page.) It may be more clear using custom templates so folks don't confuse what we're testing? Although, not necessary right now.

dlabrecq, line 35: It doesn't necessarily need to match. I simply didn't want to duplicate the JS, but test the widget rendered in the page. Doesn't sound like we can do that.

priley86, line 35: @dlabrecq I think the idea in Karma/Jasmine is you have the option to create scenarios as you see fit. I am OK with recreating the JS as it allows you to try different approaches and still test the same plugin with your framework. If someone were to change the underlying framework, introduce a conflicting framework, or accidentally remove it, the test would indicate this (even though it is not the same use case). I definitely see it worthwhile to test more use cases though.

With a large project such as this, more unit tests will really help give confidence when maintenance becomes less frequent, or when we try to integrate different approaches.

priley86, line 35: Another thing to note, `jasmine.readFixture` is a custom method I've added and is of course optional. You can always use [jasmine-jquery's](https://github.com/velesin/jasmine-jquery) `loadFixture` and it will retain any scripts.

dlabrecq, line 35: I'm ok with creating JS for different scenarios, just disappointed it's required. (It was more confusion on my part as to what we're testing.) I was hoping we could say to the community that the examples included in pf.org are unit tested. All this testing is valuable, but we're technically not testing the page examples -- not what's being included in pf.org.

priley86, line 35: @dlabrecq thanks for the feedback here...will wait until others weigh in too.

bleathem:
We should probably have a `grunt test` task defined to run our tests.  Npm run scripts are great, but so far this project is using grunt as a build tool - we should stay consistent.

bleathem:
`grunt build; npm test` gives me the error: https://gist.github.com/bleathem/6f2942db908029dc6f42dbd86cad34eb

fixed with a `bower install` as per @priley86 's suggestion

bleathem:
I'm not a fan of adding methods to external globals as we've done with `jasmine.readFixture`.
It's fragile; upstream, or some other 3rd party could break our impl.

We should consider introducing a new global to hold any helper methods we need.

bleathem:
These tests look great.  Well written and quite readable.  I like the approach taken of being able to test different scenarios with the same template file - less code to maintain.

If we want to test the exact code distributed `/dists/tests` then we could consider a dedicated set of tests to do so.

With the addition of a grunt task, and moving the `readFixture` method to a custom global I am heppy to see this PR merged.

priley86:
@bleathem @dlabrecq this PR is updated with changes. Let me know if you'd like anything else done here before merge.

bleathem:
+1 the changes look great.  This PR is ready to merge from my perspective.

andresgalante:
Can we merge this one?

priley86:
@andresgalante yep - ready for merge.

jeff-phillips-18:
+1

andresgalante:
@priley86 thanks a lot! 😄

dlabrecq:
I really wish we would have found a way to test what's already in the page. I still believe we need to understand issue regarding "query and other plugins were loading twice". This sets a precedent that other contributors will follow.

I'm ok with creating JS for different scenarios, but we're duplicating code that already exists in the page. We may be unit testing the component, but not what we've created in the test pages. I don't know that we can tell the community that the examples in pf.org have been unit tested.

priley86:
@dlabrecq i think there will be a lot of opportunity with e2e tests to test the live pages. I see jasmine/karma more as a unit test framework for incorporating specific scenarios (so it leaves the window open for lots of options). In most projects I have worked before, we used new templates strictly for testing, but in this case, wanted to make use of the ones created. I would say it is fine to create custom templates/fixtures for your tests if you think this is a good pattern to share. I can provide an example if it helps and you think it would be useful.

dlabrecq:
Yes, I'm looking for at least one, really extensive test example. Not only to prove this is the right framework for patternfly, but as a template for folks to create new tests.

If we cannot test what's already in our test pages, due to jquery loading twice or whatever reason, then perhaps another framework might work better? It doesn't seem right to load these test pages, duplicate JavaScript, and not be able to test the components already rendered in the page.

Perhaps we should create new templates so it's clear this framework can only unit test new component instances and not the components already rendered in the test pages?

bleathem:
@dlabrecq can you create a new story in jira to track these improvements you are suggesting?  I think it's great we have some automated tests in place (thanks @priley86), but let's be sure to iteratively improve on them.  We can schedule these iterative improvements in jira accordingly.

PR #360, opened 2016-07-20
Small text change to be more concise

LHinson:


jeff-phillips-18:
LGTM

dlabrecq:
LGTM.

FYI, I believe the preferred approach is to make changes in your own fork of the repo. Typically, you'll only see release branches here.

bleathem:
@dlabrecq I'd argue that's not necessarily the _preferred_ approach.  I was advocating the opposite just yesterday in fact.  We should all feel comfortable opening feature branches in the main repo, particularly if we need to collaborate on developing a feature.  Perhaps a discussion we should take up on the Patternfly ML.

In this case the branch was crated using github's web tooling for editing files and submitting those results as a PR.  As we want to encourage more non-technical folks to contribute to patternfly repos, and working from a fork demonstrably complicates the process.

bleathem:
Thanks for the PR @LHinson :+1:

dlabrecq:
Yes, we should have a discussion. At a min., folks should clean up (delete) old branches.

PR #363, opened 2016-07-21
Fix @navbar-pf-alt-bg-img variable

jtomasek:
This change makes the @navbar-pf-alt-bg-img align to other image variables. Without this change, the url to this image is incorrect when @img-path is overriden.


jeff-phillips-18:
Good catch. 👍

jeff-phillips-18:
@dtaylor113 @dlabrecq

dtaylor113:
LGTM

dlabrecq:
LGTM

PR #248, opened 2016-04-02
Make dependencies (except bootstrap, jquery and font-awesome) optional.

dlabrecq:
Changes are based on review comments from PR #240.
@dtaylor113 @jeff-phillips-18 please review.

PR #365, opened 2016-07-21
Update bower and npm dependencies

priley86:
- Updates bower and npm dependencies and devDependencies
- Should address dependency warnings currently showing in gemnasium
  - https://gemnasium.com/github.com/patternfly/patternfly
- Noted change needed to grunt-contrib-connect to pass options.base
  - https://github.com/gruntjs/grunt-contrib-connect#base


priley86:
@bleathem @dlabrecq @andresgalante @dtaylor113  would really appreciate your review here. One thing to note in this PR is the less files have changed in `less/lib` due to dependency upgrades. This is because our grunt copy task copies in library less from npm here. This will affect the compiled css as well, so some elements will be affected by css changes.

I have done a cursory review of each test page and haven't found anything significant. Was primarily checking for js errors but didn't see any.

dtaylor113:
Cloned this dev branch, npm install, bower install, grunt server -test pages look good.
loaded dev branch in browser side-by-side to an earlier version.  Visually compared a few test pages, inspected CSS -same in both browser windows
LGTM

priley86:
Linking in JIRA: [PTNFLY-1583](https://patternfly.atlassian.net/browse/PTNFLY-1583)

priley86:
@dtaylor113 thanks!

bower.json
```diff
@@ -21,21 +21,21 @@
   ],
   "dependencies": {
     "bootstrap": "~3.3.6",
-    "bootstrap-combobox": "~1.1.6",
-    "bootstrap-datepicker": "~1.4.0",
-    "bootstrap-select": "1.7.3",
```
dlabrecq, line 26: There was a reason why we had a specific version for bootstrap-select, so we should check with @rhamilto?

"I had specifically locked the dependencies to specific versions because we were seeing CSS changes resulting from differing patch versions of dependencies being installed depending on when bower install was run; in hindsight, only the offending dependencies should have been locked. In looking at #239, it appears that only Boostrap-Select needs to be locked."

Gruntfile.js
```diff
@@ -1,9 +1,4 @@
 /*global module,require*/
-var lrSnippet = require('connect-livereload')();
-var mountFolder = function (connect, dir) {
-  return connect.static(require('path').resolve(dir));
-};
-
```
dlabrecq, line 6: Would you please explain why this is changing?

bower.json
priley86, line 26: @dlabrecq is there a reason to lock the css? I guess I would like further explanation too

Gruntfile.js
priley86, line 6: `connect.static` is no longer in the api. They have added `options.base` to allow the serving of multiple directories. Please test this out and ensure it - on my machine livereload worked and changes to `tests/pages` triggered a jekyll build.

bower.json
dlabrecq, line 26: The quote was a comment in regard to #248, in which we opened up our dependencies, but kept bootstrap-select locked.

I have to defer to @rhamilto regarding bootstrap-select.

rhamilto, line 26: IIRC, bootstrap-select was shipping CSS changes in patch versions, which resulted in PatternFly build CSS changes depending on who was doing the building and when he/she last ran `bower install`.  For example, I could set up a new checkout and build and get version 1.7.4 of bootstrap-select which would then result in a PatternFly CSS build change.  My advice is to upgrade bootstrap-select to the latest version and test.

priley86, line 26: @dlabrecq @rhamilto so I guess the idea is we should lock the dependency at latest to avoid unexpected css modifications from patch level changes (assuming the css checks out OK)? 

I'm ok with this, and will revise the dependency in bower.json to the following in there are no objections:

```
"bootstrap-datepicker": "1.6.1",
"bootstrap-select": "1.10.0"
```

Are there any other dependencies we should do this with?

dlabrecq, line 26: I don't believe Robb is suggesting that we lock bootstrap-select, just test the latest version.

Why lock bootstrap-datepicker?

priley86, line 26: @dlabrecq @rhamilto thanks for the feedback on this. I have rebased and marked these dependencies with `~` to allow patch level changes.

dlabrecq:
LGTM

dtaylor113:
This PR needs to be rebased

priley86:
@dtaylor113 done!

PR #372, opened 2016-07-26
PTNFLY-938 Add 'Find' button to the toolbar

junezhang:
https://patternfly.atlassian.net/browse/PTNFLY-938
Do the implementation for adding ‘Find’ button to the toolbar.

Changed Files:
tests/pages/toolbar.html
less/toolbar.less


LHinson:
Thanks for making those updates @junezhang! It looks great. I know you've already worked with @kybaker to update the visuals but wanted to loop him in here as well. 

See: https://rawgit.com/junezhang/patternfly/PTNFLY-938/dist/tests/toolbar.html

less/toolbar.less
```diff
@@ -53,25 +53,126 @@
...
+  .toolbar-pf-action-right{
+    float: right;
+    @media (max-width: @grid-float-breakpoint) {
+      float: left;
+      width: 100%;
     }
-    a {
-      color: @gray-darker;
-      &:hover{
+    .toolbar-pf-view-selector {
```
jeff-phillips-18, line 85: Moving this selector under toolbar-pf-action-right will break any applications currently using the view selector as they won't have the toolbar-pf-action-right selector added.

less/toolbar.less
```diff
@@ -53,25 +53,126 @@
...
-    .list-inline {
-      margin-bottom: 0;
+
+    .toolbar-pf-find {
+      font-size: (@font-size-base + 2);
+      position: relative;
+      .btn{
+        outline: none;
+      }
+      .find-dropdown-container{
```
srambach: Like .toolbar-pf-view-selector, I don't think this should need to be nested under .toolbar-pf-actions

less/toolbar.less
```diff
@@ -53,25 +53,126 @@
...
+            top: 9px;
+          }
+        }
+        input{
+          padding: 5px 95px 5px 5px;
+          width: 100%;
+          height: 30px;
+          line-height: 30px;
+        }
+        .find-buttons{
```
srambach: Should all new classes we define include -pf-?

less/toolbar.less
```diff
@@ -53,25 +53,126 @@
...
+      float: left;
+      width: 100%;
     }
-    a {
-      color: @gray-darker;
-      &:hover{
+    .toolbar-pf-view-selector {
+      font-size: (@font-size-base + 4);
+      @media (min-width: @grid-float-breakpoint) {
+        border-right: 0;
```
srambach: Would it be worth looking into putting the border-right:0 onto the :last-child of .toolbar-pf-actions or on the last .form-group instead of specifically on this element?

less/toolbar.less
junezhang: Okay, I will add -pf- to all new classes. :-)
btw, if all adding -pf- to new classes, I found it is a little long, for example: .toolbar-pf-find-dropdown-container
Is there any guideline for the class name length?

Thanks.

less/toolbar.less
junezhang: The original code is putting .toolbar-pf-view-selector under .toolbar-pf-actions.
I also agreed with you. I will get it to be outside, parallel with .toolbar-pf-actions.

less/toolbar.less
junezhang: Re-consider it, I found we can't put the border-right:0 to .form-group :last-child, because if you put something float right, then the last-child element is not the exactly last child.

For example: 
CSS part

```
div{ border-right: solid 1px black;}
div:last-child{ border-right: 0;}
A, B, C{ float: right;}
```

HTML part

```
<div class='A'></div>
<div class='B'></div>
<div class='C'></div>
```

Then C will be no border.
But actually in page, the order should be C, B, A. C is the first element.

less/toolbar.less
junezhang, line 85: In beginning, I wrote the coding like you suggested:
.toolbar-pf-view-selector{}
.toolbar-pf-find{}
As float:right, in the html coding, the order should be that toolbar-pf-view-selctor is before toolbar-pf-find.
then in the view page, what we can see is toolbar-pf-find is before toolbar-pf-view-selector
they are parallel, but when in small device, every element list in a row.
then what we can see is toolbar-pf-find is after toolbar-pf-view-selector.
The order is not the same with large screen.

So I put .toolbar-pf-action-right outside of .toolbar-pf-view-selector and .toolbar-pf-find.
set .toolbar-pf-action-right float to right,  .toolbar-pf-view-selector and .toolbar-pf-find are float to left.
In HTML code, toolbar-pf-find is before toolbar-pf-view-selector.
So based the document flow, small screen is same with large screen.

I don't know if I make sense?  I draw a picture to describe it.
https://devops-ucd.rhcloud.com/June/patternfly/toolbar-code.png

Any thought?

Thanks.

jeff-phillips-18, line 85: You could have both. Leave it in its original location, then add it to toolbar-action-right and add/override the necessary settings.

less/toolbar.less
srambach: Hmm, yes, I see. And changing it to border-left on the last element doesn't work for the items on the left side. I like doing something like .form-group^.form-group but that doesn't give you the extra line to the right of the two Action buttons. I don't see another way around putting a special case on something then, as you have it.

less/toolbar.less
junezhang: Oh, I got a piece of message about user are using .toolbar-pf-view-selector, If I got it outside, it maybe affect the production users. So how about don't do it this time?

junezhang:
@jeff-phillips-18, @srambach 
Hi, I have updated the code. Please have a review.
Thanks.

less/toolbar.less
```diff
@@ -74,6 +74,129 @@
...
+        color: @gray-darker;
+        &:hover{
+          color: @link-color;
+        }
+      }
+      .list-inline {
+        margin-bottom: 0;
+      }
+    }
+
```
jeff-phillips-18: The only change I see being necessary is to over-ride the float property:
  .toolbar-pf-view-selector {
    @media (min-width: @grid-float-breakpoint) {
      float: none;
    }

The rest of these properties are already set by the .toolbar-pf-view-selector when not also in .toolbar-pf-action-right

jeff-phillips-18: The only change I see being necessary is to over-ride the float property:
  .toolbar-pf-view-selector {
    @media (min-width: @grid-float-breakpoint) {
      float: none;
    }

The rest of these properties are already set by the .toolbar-pf-view-selector when not also in .toolbar-pf-action-right

junezhang:
I'd love to move to a new pull-request to check code easier next time, so close this pull-request

PR #379, opened 2016-07-28
add example for panel-like list group

vkrizan:
Added panel-like list group example. Line height of .list-group-item-heading was increased, because of
overflowing font character 'g' (characters descender height).

### DIFF
--- a/tests/pages/list-group.html
+++ b/tests/pages/list-group.html
@@ -54,6 +54,38 @@ resource: true
         <a href="#" class="list-group-item list-group-item-danger">Vestibulum at eros</a>
       </div>
       <hr>
+      <h2>Panel List</h2>
+      <div class="list-group list-view-pf panel panel-default">
+        <a href="#" class="list-group-item list-view-pf-stacked list-view-pf-top-align">
+            <div class="list-view-pf-main-info">
+                <div class="list-view-pf-body">
+                    <div class="list-view-pf-description">
+                        <div class="list-group-item-heading">
+                            List group item heading
+                        </div>
+                        <div class="list-group-item-text">
+                            Additional text
+                        </div>
+                    </div>
+                </div>
+            </div>
+        </a>
+        <a href="#" class="list-group-item list-view-pf-stacked list-view-pf-top-align">
+            <div class="list-view-pf-main-info">
+                <div class="list-view-pf-body">
+                    <div class="list-view-pf-description">
+                        <div class="list-group-item-heading">
+                            List group item heading
+                        </div>
+                        <div class="list-group-item-text">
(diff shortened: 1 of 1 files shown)


jeff-phillips-18:
I don't see any issues when the line-height is left unchanged. Looks fine to me in Chrome, FF, and Safari. Since list-group-item-heading is pervasive, I'm hesitant to change it.

vkrizan:
It looks like a glitch in font rendering in FF. It does not disappear until the line-height is 1.23em.
screenshot from 2016-07-29 16-20-22
I am using FF 47.0.

Interestingly, that Midori has a different glitch on combination of list-group and panel:
screenshot from 2016-07-29 16-18-27

I will revert the change of line-height, since it is a rendering glitch of FF.

jeff-phillips-18:
LGTM

vkrizan:
I've cleaned out classes from the example (removed `col-md-10` and `ng-binding`).

Glitch in Midori can be because of unsupported flexbox:
https://bugs.launchpad.net/midori/+bug/1591693

andresgalante:
Hi, sorry it took me so long to review this, I was on PTO 😄 

I know that list view is based on list group, but why add an example here since there is a whole page with all the list view examples?

vkrizan:
> Hi, sorry it took me so long to review this, I was on PTO :smile:

No problem. Thank you for replying.

Are you suggesting to move it to list view page instead? The examples does not have a panel-like list. If you would add a slight gray background, it would be more visible and the content won't look like its flowing away into "space".

Here is an example from our application:
screenshot from 2016-08-19 17-56-22

andresgalante:
Hi @vkrizan, sorry this has been open for ever.
I don't really get the use case. Maybe I am wrong but to me this is a simplified version of list view, isn't it?

andresgalante:
@kybaker Can you please help me a bit with this one? It adds a panel list like this one:
http://prnt.sc/ds8eus

I am ok to merge it since it doesn't add new CSS

vkrizan:
> @kybaker Can you please help me a bit with this one? It adds a panel list like this one:
> http://prnt.sc/ds8eus

Actually the result is the one noted in https://github.com/patternfly/patternfly/pull/379#issuecomment-241058308. The other one is another example of panel-like list group (needs more modifications than adding `panel panel-default` groups).

andresgalante:
@vkrizan Thanks a lot, we are creating a new structure for list for pf4, this is the base:
https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/list-pf.html

I am going to close this PR and we can reopen it if we need it after the new list is implemented in full.

PR #380, opened 2016-07-28
Set up travis to run against all branches (if they are configured).

dgutride:
This allows developers to have a separate branch with dist files checked in - created PR so we can discuss.


scripts/publish.sh
```diff
@@ -14,13 +18,13 @@ git config user.email "patternfly@redhat.com"
...
-git remote add upstream https://$AUTH_TOKEN@github.com/patternfly/patternfly.git
+git remote add upstream https://$AUTH_TOKEN@github.com/$TRAVIS_REPO_SLUG.git
 
 # Commit generated files
 git add dist --force
 git commit -m "Added files generated by Travis build"
 
 # Push to releases branch
-git push upstream master:releases --force -v
+git push upstream $TRAVIS_BRANCH:$TRAVIS_BRANCH-releases --force -v
```
priley86, line 29: @dgutride one issue with this - wouldn't we want to consolidate "actual" releases/merges to master/tags to `releases`? I like the concept of this (for sharing test pages), but would prefer to see one branch for the latest compiled `dist` files. This would also create several branches in our main repository. Maybe we could just do this if in a fork?

dlabrecq, line 29: The reason we test TRAVIS_REPO_SLUG is to ensure only one branch (i.e., master-releases) is created in the main repo. If you turn on Travis for your fork, you would build specific branches when pushed.

Keep in mind, each developer has the option of not turning on Travis at all.

bleathem:
I've raised this issue for broader discussion on the PatternFly ML here:
https://www.redhat.com/archives/patternfly/2016-July/msg00111.html

scripts/publish.sh
priley86, line 29: @dlabrecq ok - misread this the first time. That is the desired effect i.m.o. 👍  

Can we keep the name `releases` though? I guess this simplifies the script the way it is now, but maybe something to consider in further discussion.

Good work!

dlabrecq, line 29: FYI, the name "$TRAVIS_BRANCH-releases" was added because there is potential to build many branches. During today's meeting, I believe we decided to build/push each PR, so we can review contributors code. In this case, we need something more than "releases".

We might consider naming branches something more like $TRAVIS_BRANCH-dist, instead? That is, now that this will no longer be just a release branch.

We should probably change angular-patternfly to sync with whatever naming convention we decide.

priley86, line 29: @dlabrecq @dgutride after discussion, i'm OK with this! May be worth mentioning in the ML our test url is now:
https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/

Additionally, please remove `tests/index.html` as this redirect should no longer be necessary.

jeff-phillips-18:
Please update the README.md file to inform folks doing development how to:
1) Enable travis on their forks
2) Give the URL pattern for using rawgit to view test pages.

jeff-phillips-18:
LGTM (but please squash the commits)

scripts/publish.sh
```diff
@@ -2,10 +2,14 @@
...
-if [ -z "$TRAVIS_TAG" -a "$TRAVIS_BRANCH" != "master" ]
+if [ "$TRAVIS_REPO_SLUG" = "patternfly/patternfly" ]
 then
-  echo "This commit was made against $TRAVIS_BRANCH and not the master or tag! Do not deploy!"
-  exit 0
+  echo "This build is running against patternfly/patternfly."
+  if [ -z "$TRAVIS_TAG" -a "$TRAVIS_BRANCH" != "master" ]
+  then
+    echo "This commit was made against $TRAVIS_BRANCH and not the master or tag! Do not deploy!"
+    exit 0
```
dlabrecq: FYI, this should be "exit 1". In Bourne Shell, "exit 0" indicates success.

priley86:
👍 LGTM

dlabrecq:
LGTM

PR #386, opened 2016-08-01
Vertical Navigation: add tertiary support,  make usability improvements

jeff-phillips-18:
Improvements:
- Change the delay for the initial hover over
- Primary and secondary menus should always be the same width
- Non-persistent menu behavior should be an option
- Change the size of the main navigation icons (from 14px to 20px)
- Support tooltips for the icons in the primary navigation bar

*\* Note: This commit also removes the Alternate Layouts from the tests pages as it is recommended that
applications use the Vertical Navigation components.

Migrate persistent secondary navigation:

Navigation Component (with nav-pf-vertical selector):
- change “nav-pf-vertical-with-secondary-nav” selector to “nav-pf-persistent-secondary nav-pf-vertical-with-sub-menus”
- to use collapsible secondary menus add selector “nav-pf-vertical-collapsible-menus”
- to hide primary menu bar icons add “hidden-icons-pf”

For primary menu items w/ secondary menus:
- change “persistent-secondary” selector to “secondary-nav-item-pf”

For secondary menus:
- change “nav-pf-persistent-secondary” selector to “nav-pf-secondary-nav”
- change “persistent-secondary-header" selector to “nav-item-pf-header"

Container component:
- change “container-pf-nav-pf-vertical-with-secondary” selector to container-pf-nav-pf-vertical-with-sub-menus”


jeff-phillips-18:
You can access the tests pages at: http://rawgit.com/jeff-phillips-18/patternfly/navigation-dist/dist/tests/index.html

dtaylor113:
Hi, cloned this dev branch and ran through all the Navigation test pages.  Verified secondary and primary menus same width, 'alt' test pages removed, opening & closing of secondary and tertiary menus, etc..
LGTM
Only one issue, which I'm not sure if it was designed this way.  When I select a nav menu, I see the pin icon:
image
However, when I hover over another menu, I do not see the pin icon -until I select the menu:
image

Just want to confirm that this is the desired functionality.
Thanks,
- Dave

jeff-phillips-18:
@dtaylor113 Yes, that is the desired behavior. You can only pin the active menus.

less/vertical-nav.less
```diff
@@ -34,358 +18,494 @@
...
+  }
+  h5 {
+    color: @nav-pf-vertical-secondary-color;
+    cursor: default;
+    font-size: (@font-size-base + 1);
+    font-weight: 600;
+    margin: @nav-pf-vertical-secondary-list-header-margin;
+  }
+}
+.nav-pf-vertical.nav-pf-vertical-with-sub-menus.collapsed {
```
srambach: Can we get rid of some specificity here by removing .nav-pf-vertical? Or even go down to just .collapsed? Does a .nav-pf-vertical-with-sub-menus exist any other way that within .nav-pf-vertical? In general, we shouldn't much need to nest selectors for PF-specific classes.

less/vertical-nav.less
```diff
@@ -34,358 +18,494 @@
...
-        }
-      }
-    }
-    .secondary-visible-pf.collapsed.collapsed-secondary-nav-pf & {
-      @media (min-width: @screen-lg-min) {
-        left: 0;
-      }
-    }
+  .secondary-nav-item-pf.active &,
+  .secondary-nav-item-pf.mobile-nav-item-pf & {
```
srambach: it's so easy to miss the & at the end of the selector; can you add in some comments when using that?

srambach:
The Less/CSS looks way clearer than the old vertical nav. It would still be nice to have some comments saying what's going on for a section though, just to make it easier to find your way. :-)

srambach:
Also, would it be worth putting a comment at the top outlining the very basic expected HTML structure of a vertical nav? E.g. 
nav-pf-vertical [nav-pf-vertical-with-sub-menus optional] [nav-pf-vertical-collapsible-menus]
|_ ul.list-group
  |_ li.list-group-item [secondary-nav-item-pf] [active]
    |_ ul.list-group
      |_ li.list-group-item [tertiary-nav-item-pf] [active]
Use nav-item-pf-header for menu heading
(or whatever...etc.) Then it's easier when you come back to understand what all is going on here.

jeff-phillips-18:
Updated less file to reduce complexity and added comments.

jeff-phillips-18:
I haven't seen any Patternfly less files that include that kind of commentary (html structure). We use our tests pages to demonstrate the HTML use for the supported components.

srambach:
I haven't actually done that kind of comment either, but was thinking it would have been useful when trying to figure out how the menu was working. I know you get that in the test pages...I was thinking more of a really bare bones/essential elements/at a glance sort of thing.

jeff-phillips-18:
@srambach Added requested commentary.

dgutride:
Code looks good to me.

less/vertical-nav.less
```diff
@@ -34,358 +48,585 @@
...
+  left: (@nav-pf-vertical-width + @nav-pf-vertical-width);
+  opacity: 0;
+  overflow-x: hidden;
+  overflow-y: auto;
+  position: fixed;
+  top: @navbar-pf-height;
+  visibility: hidden;
+  width: @nav-pf-vertical-width;
+  z-index: @zindex-navbar-fixed;
+  .tertiary-nav-item-pf.active & {  // Show tertiary menu if active and collapsed
```
priley86, line 615: Could some of these styles (such as overflow-x, overflow-y, visibility) be shared with the secondary-nav class? I guess some of this seems redundant to me, but I'm OK with it if we think the separation makes sense.

jeff-phillips-18, line 615: I thought the same but I would have to add another class to both the items with nav-pf-secondary-nav and nav-pf-tertiary-nav in order to reuse these settings. We could make that change if it felt cleaner, I was just trying to reduce the number of required classes on the items.

priley86:
@jeff-phillips-18 downloaded your branch, built it, and reviewed each of the vertical nav test pages on mobile, desktop, and tablet emulations. They all worked great!

I noticed the differing functionality of the collapse toggle button (top left) in the `vertical-navigation-with-tertiary-pins` test on mobile and desktop. The mobile collapse toggle returns you to the previous nav (i.e. tertiary to secondary), whereas the desktop collapse toggle will collapse the current nav to the left hand side (i.e. tertiary collapses to left). I like this functionality, so kudos for the UX on this!

src/js/patternfly.js
```diff
@@ -1069,12 +1069,14 @@
         'tablet': 768,
         'desktop': 1200
       },
```
priley86: noticed these breakpoints were defined elsewhere in `$.fn.navigation`. Can we possibly consolidate these? Or is it suggested they be separate.

jeff-phillips-18:
@priley86 Thanks. Updated to remove the duplication.

priley86:
@jeff-phillips-18 Awesome! LGTM 👍

srambach:
LGTM! Nice enhancement.

srambach:
Actually, one thing I just noticed while investigating a bug in what I did - at a narrow breakpoint, when the menu goes under the hamburger icon, the second/third menus don't fly out. Should they? (@jeff-phillips-18 answered that is correct behavior... LGTM now)

PR #394, opened 2016-08-04
Modified -s switch of release.sh script

dlabrecq:
Modified -s switch to skip new clone, clean, and install to rebuild existing repo.


priley86:
LGTM 👍  This will speed things up quite a bit.

Issue #395, opened 2016-08-04
Need to remove this right-border in toolbar view selector


junezhang:
https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/toolbar.html
There should be a bug. Need to remove this right-border of the view selector.

junezhang:
Fixed this issue, close it.
https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/toolbar.html

PR #396, opened 2016-08-04
Fixed the issue #395

junezhang:
Need to remove this right-border in toolbar view selector
https://github.com/patternfly/patternfly/issues/395

When I check the new toolbar page, I found there is a bug, there is a right border in the element of view selector. So create this PR to fix it.

### DIFF
--- a/less/toolbar.less
+++ b/less/toolbar.less
@@ -54,28 +54,29 @@
       width: 25%;
     }
   }
-}
-.toolbar-pf-view-selector {
-  font-size: (@font-size-base + 4);
-  @media (min-width: @grid-float-breakpoint) {
-    border-right: 0;
-    float: right;
-    padding-right: 0;
-  }
-  .active a {
-    color: @link-color;
-    cursor: default;
-  }
-  a {
-    color: @gray-darker;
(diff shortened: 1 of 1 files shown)


jeff-phillips-18:
LGTM

junezhang:
@srambach Could you help to review it? Thanks.

srambach:
I hate to see this nested again because it is having specificity problems already. What about using .toolbar-pf-view-selector:last-child {border-right: 0px;} instead?

jeff-phillips-18:
Seems to me setting it on the last-child makes more sense anyhow. If there is no view selector, then there should be no right border on the find group either. So, .toolbar-pf-actions-right:last-child{border-right: 0px;}

I would say the same would hold true for the toolbar in general so also .toolbar-pf-actions:last-child{border-right: 0px;}

@LHinson Do you agree that there should be no vertical bar to the right when there is no more following content?

LHinson:
Yes agreed @jeff-phillips-18

junezhang:
Okay, Thanks all for your kind suggestions. I will close it and re-submit a new one.

PR #397, opened 2016-08-05
PTNFLY-1647 - Horizontal Navigation - Make it sticky

junezhang:
https://patternfly.atlassian.net/browse/PTNFLY-1647

Add class '.nav-fixed-top' to the navigation horizontal html
Add padding top to the container-fuild

Changed files
less/layouts.less
tests/pages/_includes/widgets/layouts/nav-horizontal-notification-drawer.html

I don't know how to get the demo link, let you see the demo page.
So I add the html page to my sever. 
Please check the demo page here: https://devops-ucd.rhcloud.com/June/patternfly/dist/tests/notification-drawer-horizontal-nav.html

### DIFF
--- a/less/layouts.less
+++ b/less/layouts.less
@@ -106,3 +106,25 @@
     }
   }
 }
+// when nav is fixed, set padding-top to the main container
+.navbar-fixed-top + .container-fluid{
+  padding-top: 72px;
+  @media (max-width: @grid-float-breakpoint){
+    padding-top: 38px;
+  }
+}
+.navbar-fixed-top + .nav-pf-vertical + .container-fluid{
+  padding-top: @navbar-pf-height;
+  padding-left: ( @nav-pf-vertical-width + 20px );
+  @media (max-width: @grid-float-breakpoint){
+    padding-left: 20px;
+  }
+}
+
+// move the transitions out of layout-pf-fixed
+.transitions .container-pf-nav-pf-vertical {
+  transition: @flyout-transition-pf;
+  -webkit-transition: margin-left @nav-pf-container-transition-period;
+  -moz-transition: margin-left @nav-pf-container-transition-period;
+  -o-transition: margin-left @nav-pf-container-transition-period;
+}
--- a/tests/pages/_includes/widgets/layouts/nav-horizontal-notification-drawer.html
+++ b/tests/pages/_includes/widgets/layouts/nav-horizontal-notification-drawer.html
@@ -1,4 +1,4 @@
-<nav class="navbar navbar-default navbar-pf" role="navigation">
+<nav class="navbar navbar-default navbar-pf navbar-fixed-top" role="navigation">
   <div class="navbar-header">
     <button type="button" class="navbar-toggle" data-toggle="collapse" data-target=".navbar-collapse-1">
       <span class="sr-only">Toggle navigation</span>
--- a/tests/pages/_includes/widgets/layouts/nav-vertical-notification-drawer.html
+++ b/tests/pages/_includes/widgets/layouts/nav-vertical-notification-drawer.html
@@ -1,4 +1,4 @@
-<nav class="navbar navbar-pf-vertical">
+<nav class="navbar navbar-pf-vertical navbar-fixed-top">
   <div class="navbar-header">
     <button type="button" class="navbar-toggle">
       <span class="sr-only">Toggle navigation</span>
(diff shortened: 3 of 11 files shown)


junezhang:
@kybaker @andresgalante @LHinson @serenamarie125 
Please have a review about it, if is it okay?
-June

dlabrecq:
I'm looking at the "Notification Drawer for Vertical Navigation" example and it also has a sticky header. However, we didn't add navbar-fixed-top there. I believe both examples should share similar solutions.

FYI, it appears that the vertical navigation example is using layout-pf-fixed Vs layout-static for the horizontal example. I can see layout-pf-fixed is being overridden in navbar-pf-vertical, which probably makes it sticky.

@cardosogabriel implemented these examples initially, so perhaps he can clarify what was done there? Either solution is fine by me, just thinking it would be best to have consistency.

jeff-phillips-18:
@dlabrecq Correct, the class layout-pf-fixed in navbar-vertical.less does exactly what navbar-fixed-top is doing. We could simply add navbar-fixed-top to the nav for the vertical nav component and remove .navbar-pf-vertical.layout-pf-fixed

jeff-phillips-18:
We should also update the horizontal navigation test page as well (without notification drawer) and any other pure full page view with horizontal nav (I believe only the include navbar-primary.html would need to be updated)

jeff-phillips-18:
While checking this out I noticed something I should have noticed while developing the vertical navigation component 😳 

The toast notifications (in particular the "These examples...") should also stick to the top of the page when using fixed layouts.

I believe adding:

.layout-pf-fixed .toast-pf-top-right {
  position: fixed;
}

to toast.less will accomplish this.

junezhang:
I agreed with @jeff-phillips-18. Keep it simply, we can use ".nav-fixed-top" to make the nav sticky.

junezhang:
@jeff-phillips-18  If I update the other files (navbar-primary.html), do i need to close this PR and re-summit a new one or just git push -f this commit?

jeff-phillips-18:
You can just add to this PR.

junezhang:
HI, @jeff-phillips-18 I updated the navbar-primary.html, please have a check.
@LHinson, @jeff-phillips-18  two things need to confirm with you:
1, About fixing the toast notification, could we move it to another PR?
2, Are we all sure that need to fix all the horizontal navigation test page as well, like these links:
- https://devops-ucd.rhcloud.com/June/patternfly/dist/tests/basic.html
- https://devops-ucd.rhcloud.com/June/patternfly/dist/tests/cards.html
- ......
- https://devops-ucd.rhcloud.com/June/patternfly/dist/tests/notification-drawer-horizontal-nav.html

jeff-phillips-18:
I actually see three places :(

layouts/navbar-primary.html (you covered this)
layouts/nav-horizontal-notification-drawer.html (you covered this)
navigation/horizontal-primary-nav-bar.html (this is not yet covered, see the Navigation -> Horizontal Navigation page)

Eventually, it would be nice to DRY this up and have the later 2 simply include the first.

I'm OK either with fixing the notification drawer in this PR or adding another PR (or filing an issue so that it is tracked). Simple enough change.

junezhang:
@jeff-phillips-18 fixed all the navigation bar. have a look. :-)

jeff-phillips-18:
LGTM. @dlabrecq are you OK with this not changing the vertical nav to use the common style? Can we just add an issue to investigate that and get this PR merged?

dlabrecq:
I'm on the fence. If we don't use a common style, then the sticky solutions for vertical and horizontal nav are not in sync. If we later change vertical nav to use nav-fixed-top, then existing code may break.

If we must get this in for the sprint, it's your call.

junezhang:
@dlabrecq I also updated the vertical navigation pages.  All using '.navbar-fixed-top' solution, the two navigations have been in sync.  Please have a review.

jeff-phillips-18:
Did you test vertical navigation on IE9 after this change?

The selector
  .ie9.layout-pf-fixed & {
    box-sizing: content-box; // IE9 incorrectly sizes the width if using padding-box
  }
will no longer come into play.

dlabrecq:
@jeff-phillips-18 Should we remove this from navbar-vertical.less or leave it for legacy apps? We're using layout-pf-static now, so it's just cleanup.

  .layout-pf-fixed & {
    left: 0;
    position: fixed;
    right: 0;
    top: 0;
    z-index: @zindex-navbar-fixed;
  }

jeff-phillips-18:
I believe we need to leave it for legacy apps. We could remove it in the next major release.

jeff-phillips-18:
Actually, we might want to think about using this consistently rather than adding navbar-fixed-top

junezhang:
Adopt another solution, close it.

PR #409, opened 2016-08-12
Fix for Fixed Accordions when no parent selector is given.

jeff-phillips-18:
The inner height variable was not initialized when no parent selector is given so the body's max height value was not being set.


jeff-phillips-18:
@dtaylor113  @dlabrecq  Please review.

dgutride:
looks good

dlabrecq:
LGTM

PR #418, opened 2016-08-17
PTNFLY-1647 - Horizontal Navigation - Make it sticky

junezhang:
https://patternfly.atlassian.net/browse/PTNFLY-1647
Horizontal Navigation - Make it sticky

Change the layout to layout-fixed
Add .navbar-pf to .layout-pf-fixed

### DIFF
--- a/less/layouts.less
+++ b/less/layouts.less
@@ -16,6 +16,13 @@
     body {
       padding-top: @navbar-pf-height; // make space for the navbar
     }
+    .navbar-pf {
+      left: 0;
+      position: fixed;
+      top:0;
+      right: 0;
+      z-index: 1030;
+    }
     .container-pf-nav-pf-vertical {
       margin-left: @nav-pf-vertical-width;
       &.collapsed-nav {
--- a/tests/pages/horizontal-navigation.html
+++ b/tests/pages/horizontal-navigation.html
@@ -1,7 +1,7 @@
 ---
 categories: [Navigation]
 css-extra: false
-layout: layout-static
+layout: layout-fixed
 resource: true
 full-page: true
 title: Horizontal Navigation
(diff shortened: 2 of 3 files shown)


less/layouts.less
```diff
@@ -16,6 +16,13 @@
     body {
       padding-top: @navbar-pf-height; // make space for the navbar
     }
+    .navbar-pf{
+      left: 0;
```
jeff-phillips-18, line 20: nit... please add a space before the {

jeff-phillips-18:
LGTM

less/layouts.less
junezhang, line 20: yes, indeed. Update it.

dlabrecq:
LGTM

cardosogabriel:
@junezhang , @andresgalante asked me to review your PR. The way you did it works, but the CSS you wrote is already supported by bootstrap. You just need to add the class navbar-fixed-top to the component. https://getbootstrap.com/components/#navbar-fixed-top
Can you please update the code and send a new PR? Also, please test in IE 9+.
Many thanks!

junezhang:
@cardosogabriel Thanks for review it.
Yeah, I think it is a long story about the stick nav. Please refrence this link: https://github.com/patternfly/patternfly/pull/397
At last, we decided to keep consistent with the vertical navigation. So adopt this solution.

btw, how do you test IE on Mac, using Browserstack(https://www.browserstack.com/)?

cardosogabriel:
Hi @junezhang. Perhaps I missed the point, but why u guys decided to go for the CSS instead of applying the class to every bar that needs to be fixed?

To test on IE you can use Browserstack, it's awesome. Thanks

andresgalante:
@junezhang @jeff-phillips-18 @dlabrecq  Can you please follow up @cardosogabriel question? Why not use a default bootstrap class and repeat the solution?
By using the bs class we are allowing people to opt for making it sticky or not if they are not using the notification drawer. Thanks!

dlabrecq:
If I recall correctly, the vertical navigation already had its own solution. Both achieved the same result, but we want to use the same approach for the sake of consistency. And changing vertical nav to use the Bootstrap selector now would likely break existing apps?

PR #420, opened 2016-08-17
Update to toast list to set appropriate width

jeff-phillips-18:
Also:
Fix toasts inside toast list so an extra div is not needed.
Remove the toast notifications list test page and update the toast
notifications test page to use the list.
Use toast notification list on full page layout test pages for example
warning text.


dgutride:
This looks good to me.

jeff-phillips-18:
Pushed minor updates to use grid-gutter-width / 2 rather than hard coded left margin for the toast list and to remove the left margin when not in mobile mode.

less/toast.less
```diff
@@ -3,7 +3,7 @@
 // --------------------------------------------------
 
 .toast-pf {
-  background-color: hsla(0, 0%, 100%, .94);
+  background-color: hsla(0, 0%, 100%, 1);
```
andresgalante: @jeff-phillips-18 please use the variable white.

less/toast.less
```diff
@@ -51,11 +51,21 @@
 }
 
 .toast-notifications-list-pf {
+  margin-left: (@grid-gutter-width / 2);
   position: fixed;
   right: 15px;
```
andresgalante: same here, gutter/2 for right and top

less/toast.less
```diff
@@ -3,7 +3,7 @@
 // --------------------------------------------------
 
 .toast-pf {
-  background-color: hsla(0, 0%, 100%, .94);
+  background-color: hsla(0, 0%, 100%, 1);
   border-color: @color-pf-black-400;
   box-shadow: 0 2px 6px hsla(0, 0%, 0%, .2);
```
andresgalante: please use a less fade http://lesscss.org/functions/#color-operations-fade of the white variable

less/toast.less
```diff
@@ -51,11 +51,21 @@
...
   position: fixed;
   right: 15px;
   top: 15px;
   z-index: 1050;
   .toast-pf {
+    clear: both;
     float: right;
   }
+  @media (min-width: @screen-sm-min) {
+    margin-left: 0;
```
andresgalante: now that I am thinking, if the width is 50% then it doesn't matter if we leave the margin left, right?

andresgalante:
Thanks @jeff-phillips-18 !

less/toast.less
jeff-phillips-18: correct. I will remove this.

less/toast.less
jeff-phillips-18: Changing this to straight @color-pf-white without any fading per comment above. This was really unnecessary once I changed the opacity to 1. Sorry about that.

jeff-phillips-18:
Addressed @andresgalante  issues.

PR #423, opened 2016-08-18
Creates a PR template

andresgalante:
@priley86 @dlabrecq I've created this PR template to help get better PR descriptions and improve our for release notes.

What do you think?

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,18 @@
+## Description
+A few sentences describing the overall goals of the pull request's commits.
+
+Include related PRs or Issues
+
+Include a list of changes:
+
+* Change 01
+* Change 02
+* Change 03
+
+## Link to rawgit and/or image
+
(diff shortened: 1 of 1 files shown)


priley86:
bitmoji

I like this a lot - it will remind us to write better notes in PRs. 

Great idea on this!

dlabrecq:
It would be incredibly helpful if folks provided better descriptions. When generating the release, we often have nothing except a couple words.

Unfortunately, I don't believe this would have captured @himdel's issues related to "will break any right floating toolbar" nor "will break toolbar spacing and separators". It's not clear that we knew that ourselves?

It would be good to discuss during the uxd/ui dev meeting and get consensus. Would like to know how all this will translate to release notes, how do we enforce it, etc.? 

I worry that a template may be overkill for some cases? For example, we made some simple script changes during the last release. I don't see the template being applicable for that (e.g., there is no rawgit). Perhaps we need an optional section to the template to cover more types of changes?

andresgalante:
Great point Dan. Which parts would you ass to the template? And which ones would be optional?

priley86:
👍  I'm ok with merging this template for now, and making adjustments as needed.

PR #428, opened 2016-08-23
Added two column layout for about modal

dlabrecq:
## Description

Added two column layout for "versions that contain both a release name and version number or in the event more space is needed." This is needed for the angular implementation.

PTNFLY-1101
## Changes
- Added .product-versions-cols-pf

### DIFF
--- a/less/about-modal.less
+++ b/less/about-modal.less
@@ -39,6 +39,12 @@
   }
 }
 
+.product-versions-cols-pf {
+  -webkit-columns: 2; /* Chrome, Safari, Opera */
+  -moz-columns: 2; /* Firefox */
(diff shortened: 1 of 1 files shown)


dlabrecq:
@jeff-phillips-18 This is needed for the angular implementation.

andresgalante:
@dlabrecq Can you please add an HTML example of it on the about modal test page so we can see what this looks like?

dlabrecq:
It would look like the image below. However, after speaking with Leslie, the layout should be one column -- the design doc appears to be incorrect. I'm going to close this and stick with one column.

screen shot 2016-08-23 at 3 28 17 pm

Issue #401, opened 2016-08-09
Variable referenced before declaration creates issues for SASS conversion


skateman:
The `font-size-base` is referenced before its declaration. This is acceptable in LESS, but not in SASS! Could you please do a small reorder in the `variables.less` file to ease the conversion process?

skateman:
ping @jeff-phillips-18

PR #429, opened 2016-08-23
Fix for #401 - Move use of font-size-base after declaration

jeff-phillips-18:
This fixes issue #401 

Move use of font-size-base after its declaration


andresgalante:
Thanjs @jeff-phillips-18

PR #431, opened 2016-08-23
Removing class select from cards of Card View.

cardosogabriel:
## Description

To support cards that are selectable or NOT selectable, I removed the class card-pf-view-select from the card widget. Now, this class should be declared in the view that consumes the widget.

This change is to support the Card View Page in patternfly-org: https://patternfly.atlassian.net/browse/PTNFLY-1586

### DIFF
--- a/less/card-view.less
+++ b/less/card-view.less
@@ -69,7 +69,7 @@
   }
   .card-pf-view-checkbox {
     position: absolute;
-    top: 15px;
+    top: 11px;
     left: 15px;
     input[type=checkbox] { display: none; }
   }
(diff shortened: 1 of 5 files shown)


jeff-phillips-18:
LGTM

jeff-phillips-18:
@dlabrecq @andresgalante Please take a look when/if you have a moment.

andresgalante:
👍  thanks a lot @cardosogabriel and @jeff-phillips-18 for the review.

dlabrecq:
LGTM

PR #432, opened 2016-08-23
references our dependencies from node_modules instead of bower

priley86:
## Description

This PR attempts to satisfy the requirements mentioned in [PTNFLY-1661](https://patternfly.atlassian.net/browse/PTNFLY-1661). 

Changes:
- Updates README and QUICKSTART to change language regarding Bower
- Updates karma.conf.js to reference `node_modules` instead of `components`
- Updates Gruntfile to reference `node_modules` instead of `components`
- noted formatting differences in `bootstrap-select` less files (npm vs bower)
- noted that `datatables.net-colreorder-bs` was needed to incorporate datatables-colreorder bootstrap css from NPM (`datatables.net-colreorder` is missing the css)


bleathem:
In a local checkout of this branch, I've successfully run an `npm install`, and `grunt server` both executes the build without error, and renders the test pages as expected.

:+1:

dgutride:
The code looks good - RCUE documentation will need to be updated in parallel as it depends completely on cloning patternfly and then applying styles only.

README.md
```diff
@@ -27,30 +35,22 @@ Are you using [Wiredep](https://github.com/taptapship/wiredep)?  PatternFly's CS
...
+  "node_modules/bootstrap-switch/dist/css/bootstrap3/bootstrap-switch.css",
+  "node_modules/patternfly-bootstrap-treeview/dist/bootstrap-treeview.min.css",
+  "node_modules/c3/c3.css",
+  "node_modules/datatables/media/css/jquery.dataTables.css",
+  "node_modules/datatables.net-colreorder-bs/css/colReorder.bootstrap.css",
+  "node_modules/drmonty-datatables-colvis/css/dataTables.colVis.css",
+  "node_modules/eonasdan-bootstrap-datetimepicker/build/css/bootstrap-datetimepicker.min.css",
+  "node_modules/font-awesome/css/font-awesome.css",
+  "node_modules/google-code-prettify/bin/prettify.min.css"
 ],
```
jeff-phillips-18, line 51: Will this require applications already using patternfly to update their wiredep config?

jeff-phillips-18:
The quick start guide has info on including scripts into the project HTML file(s). There is a bower section and an NPM section, but both appear to be the same now. Wouldn't the location of the files for bower depend on where their bower directory is (specified in the project's .bowerrc)? Why is it necessary to change the Quickstart guide in the bower install section from bower_components to npm_modules?

priley86:
I was also thinking about updating this [page](http://www.patternfly.org/get-started/setup/#_) in patternfly.org to reference node modules instead of bower. What do you guys think?

priley86:
@jeff-phillips-18 after rereading the QUICKSTART, i agree, it doesn't make sense. I missed this while updating references. I will change it to first suggest referencing NPM/node modules for Patternfly dependencies, and then second, suggest the use of Bower. Will push this update shortly.

priley86:
@jeff-phillips-18 i've updated the QUICKSTART guide.

priley86:
@skateman any idea if this change will impact patternfly-sass?

skateman:
@priley86 no

README.md
priley86: @jeff-phillips-18 I will send a note to the ML informing folks that we have made this change, and to just be aware that the documentation now references node_modules for wiredep.

priley86:
@skateman thanks :)

priley86:
@dgutride i will submit a PR for RCUE later today to update the documentation in README and QUICKSTART.

jeff-phillips-18:
LGTM 👍

priley86:
Linking [PTNFLY-1661](https://patternfly.atlassian.net/browse/PTNFLY-1661)

Issue #435, opened 2016-08-24
The badges on the primary menu for vertical navigation are not vertically aligned.


jeff-phillips-18:
When adding badges to the primary nav items in vertical navigation the badge does not align vertically with the text/icon:

![image](https://cloud.githubusercontent.com/assets/11633780/17939772/1c7b9f5e-69fa-11e6-8859-bc14257efc96.png)

PR #439, opened 2016-08-26
Add support for badges in Vertical Navigation

jeff-phillips-18:
## Description

This adds support for badges which will increase the width of the
navigation menus by 50px to allow space for the badges. To use this
badge support add the class ‘nav-pf-vertical-with-badges’ to the
nav-pf-vertical div as well as the body container.

This also fixes issue #435

http://rawgit.com/jeff-phillips-18/patternfly/navigation-dist/dist/tests/vertical-navigation-with-badges.html

image

### DIFF
--- a/less/variables.less
+++ b/less/variables.less
@@ -385,6 +385,7 @@
 @input-color:                                                       @gray-dark;
 @nav-tabs-active-link-hover-color:                                  @link-color;
 @nav-tabs-justified-link-border-color:                              @nav-tabs-border-color;
+@nav-pf-vertical-badges-width:                                      (@nav-pf-vertical-width + 50px);
 @padding-small-horizontal:                                          @padding-base-horizontal;
 @padding-small-vertical:                                            @padding-base-vertical;
 @panel-danger-border:                                               @brand-danger;
--- a/less/layouts.less
+++ b/less/layouts.less
@@ -22,6 +22,9 @@
     }
     .container-pf-nav-pf-vertical {
       margin-left: @nav-pf-vertical-width;
+      &.nav-pf-vertical-with-badges {
+        margin-left: @nav-pf-vertical-badges-width;
+      }
       &.collapsed-nav {
         margin-left: @nav-pf-vertical-collapsed-width;
         &.hidden-icons-pf {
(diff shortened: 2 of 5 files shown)


jeff-phillips-18:
@andresgalante @jgiardino @srambach Please review

jeff-phillips-18:
@kybaker Could you take a look please?

srambach:
If the navigation is collapsed to just icons, the badge shows instead of the icon.

srambach:
Also, there is some weirdness when the window is pulled small enough to collapse the navigation and then back out again. The page is a bit mixed up until a reload. Happens in at least Safari and Chrome.

jeff-phillips-18:
Fixed @srambach issues (badges over icons, sizing issues in persistent secondary mode), left in a separate commit to ease seeing what had to be changed. I will squash before merge.

less/variables.less
```diff
@@ -103,6 +103,7 @@
 @nav-pf-vertical-badge-bg-color:                                    @color-pf-black-900;
 @nav-pf-vertical-bg-color:                                          @color-pf-black-900;
 @nav-pf-vertical-border-color:                                      @color-pf-black-900;
```
srambach: This could be expressed as @nav-pf-vertical-width + 50 so that if the normal width is changed, this follows, but not really necessary.

less/variables.less
```diff
@@ -103,6 +103,7 @@
...
+@nav-pf-vertical-badges-width:                                      250px;
```
srambach: This could be expressed as @nav-pf-vertical-width + 50 so that if the normal width is changed, this follows, but not really necessary.

andresgalante:
@jeff-phillips-18 it doesn't support long names:
screen shot 2016-08-26 at 3 36 54 pm

jgiardino:
There are some declarations that are repeated for the primary, secondary,
and tertiary nav badges. Is there a possibility that these would need to be
different? I'm wondering if these can just be removed. See specific
examples below (copied from the browser inspector, not the .less files)...

The styles defined here…

.nav-pf-tertiary-nav .list-group-item .badge-container-pf {

```
background-color: #292e34;

top: 5px;
```

}

…Are also applied here, and these styles are applied to all instances of
.badge-container-pf targeted by the selectors in the previous ruleset:

.nav-pf-secondary-nav .list-group-item .badge-container-pf {

```
background-color: #292e34;

top: 5px;
```

}

Additionally, the background-color defined in the rulesets above seems
unnecessary since it’s also defined for the following selector, which
targets all instances of .badge-container-pf in the navigation.

.nav-pf-vertical .badge-container-pf

The styles defined in the following two rulesets…

.nav-pf-tertiary-nav .list-group-item .badge-container-pf .badge {

```
background: #292e34;

color: #fff;
```

}

.nav-pf-secondary-nav .list-group-item .badge-container-pf .badge {

```
background: #292e34;

color: #fff;
```

}

…Are also applied here, and these styles are applied to all instances of
.badge targeted by the selectors in the previous rulesets:

.nav-pf-vertical .badge-container-pf .badge {

```
background: #292e34;

color: #fff;

float: left;

font-size: 12px;

font-weight: 700;

line-height: 1.66666667;

margin: 0;

padding: 0 7px;

text-align: center;
```

}

Jenn Giardino
Senior Interaction Designer
User Experience Design Team
Red Hat, Inc.
1-919-716-5045
jgiardin@redhat.com

jeff-phillips-18:
Fixed issues:
  @srambach Use @nav-pf-vertical-width + 50px for @nav-pf-vertical-badges-width
  @andresgalante  Set max width and ellipsis for navigation item titles

jeff-phillips-18:
@jgiardino Thanks. I have removed the background color on badge-container-pf. For the badge background colors, each level is set to a color variable. Currently they are all the same value, but I did that purposefully so that we could use different colors for the badge backgrounds if desired (easier this way to change). If we can count on them always being the same background, then I can change this.

LHinson:
@kybaker see comment above. Would you ever recommend that the badges be a color other than gray? OTOH, might be nice to just go ahead and leave this built in flexibility.

andresgalante:
@jeff-phillips-18 thanks! the ellipsis on the primary nav works great, on secondary nav it "cuts" before the end of the line. What's the logic for the width of the word?

screen shot 2016-08-29 at 12 57 29 pm

jeff-phillips-18:
I was leaving room for 2 badges with icons. I could change the max-width to be wider when badges are not being used. Any ideas on setting it wider if there are actually badges?

jeff-phillips-18:
@srambach @jgiardino @andresgalante  Any further issues/concerns?

andresgalante:
@jeff-phillips-18 I am ok to merge this one. thanks a lot.

kybaker:
@jeff-phillips-18 @LHinson  @jgiardino We should absolutely leave the flexibility to modify the background colors for each level. There are many reasons not limited to shifting the branding which would require a change to accommodate the new level of badge contrast. Looks good to me Jeff.

jgiardino:
Sounds good to me :-)

Jenn Giardino
Senior Interaction Designer
User Experience Design Team
Red Hat, Inc.
1-919-716-5045
jgiardin@redhat.com

PR #441, opened 2016-08-29
PTNFLY-452 List View:Expand Row - PatternFly Core

junezhang:
## Description

Implemented two kinds of expand page for list view.
https://patternfly.atlassian.net/browse/PTNFLY-452

Demo Page
- Simple Expansion - [http://devops-ucd.rhcloud.com/June/patternfly/PTNFLY-452/dist/tests/list-view-simple-expansion.html](http://devops-ucd.rhcloud.com/June/patternfly/PTNFLY-452/dist/tests/list-view-simple-expansion.html)
- Compound Expansion - [http://devops-ucd.rhcloud.com/June/patternfly/PTNFLY-452/dist/tests/list-view-compound-expansion.html](http://devops-ucd.rhcloud.com/June/patternfly/PTNFLY-452/dist/tests/list-view-compound-expansion.html)

New flies for simply expansion
tests/pages/_includes/widgets/list-view/list-view-page-simple-expansion.html
tests/pages/_includes/widgets/list-view/list-view-standard-rows-simple-expansion.html
tests/pages/list-view-simple-expansion.html

New files for compound expansion
tests/pages/_includes/widgets/list-view/list-view-page-compound-expansion.html
tests/pages/_includes/widgets/list-view/list-view-standard-rows-compound-expansion.html
tests/pages/list-view-compound-expansion.html

### DIFF
--- a/less/list-view.less
+++ b/less/list-view.less
@@ -8,12 +8,20 @@
     align-items: flex-start;
     .clearfix(); //IE9 fallback
     display: flex;
-    padding-bottom: 0;
-    padding-top: 0;
-    &.active {
+    flex-wrap: wrap;
+    padding: 0;
+    margin-bottom: 0;
+    border-top: 0;
+    &.active,
+    &.expand-active {
       color: @list-group-link-color;
       background-color: @list-view-active-bg;
-      z-index:auto;
+      z-index: auto;
+      border-top: solid 1px @list-group-active-border;
+      margin-top: -1px;
+    }
+    &.expand-active {
+      border: solid 1px @list-group-active-border;
     }
     &:hover {
       background-color: @list-view-hover-bg;
@@ -53,9 +61,7 @@
 }
 .list-view-pf-actions {
   float: right; // IE9 fallback
-  margin-bottom: (@grid-gutter-width/2);
-  margin-left: (@grid-gutter-width/2);
-  margin-top: (@grid-gutter-width/2);
+  margin: (@grid-gutter-width/2) 15px;
   order: 2;
   button,
   > a {
@@ -122,12 +128,13 @@
 .list-view-pf-checkbox {
   border-right: 1px solid @list-view-divider;
   float: left; //IE9 fallback
-  margin-bottom: (@grid-gutter-width/2);
-  margin-right: 15px;
-  margin-top: (@grid-gutter-width/2);
+  margin: (@grid-gutter-width/2) 15px;
   padding: 3px (@grid-gutter-width/4) 3px 0;
   .list-view-pf-top-align & {
     align-self: flex-start;
+  }
+  .list-view-pf-expand + & {
+    margin-left: 5px;
   }
 }
 .list-view-pf-description {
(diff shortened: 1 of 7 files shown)


less/list-view.less
```diff
@@ -237,3 +237,87 @@
 .list-view-pf-view {
   margin-top: 30px;
 }
+.list-group-item-simple-expand, .list-group-item-compound-expand {
+  &.active {
+    border: solid 1px @gray-light !important;
+    box-shadow: 0 0 0 #e5e5e5, 0 0 2px rgba(0, 0, 0, .12), 0 2px 4px rgba(0, 0, 0, .24);
+    color: @list-group-link-color;
```
jeff-phillips-18: This looks like a very marginal color change. Worth doing?

less/list-view.less
```diff
@@ -237,3 +237,87 @@
...
+    margin-top: -1px;
+  }
+  &.active .list-group-item {
+    background-color: @list-view-hover-bg;
+    border-top: 1px solid @gray-light !important;
```
jeff-phillips-18: non-expanded active items have a blue border, why not keep that color when expanded? It is harder to tell if an item is active when expanded.

andresgalante:
Hi @junezhang sorry it took me so long to review this PR.

I think we should approach the expansion box in a different way. I don't like that it's a separate div outside the row itself.

I'd make it a div inside `list-group-item`. You'll probably need to set the `order` of that div to 3 and make it `flex` expand the 100% of the row.

Also, if you are going to use bootstrap grid, please for the correct structure with rows and cols and never use a clear fix as an empty div.

Please let me know if you have any question.

Thanks!

junezhang:
@andresgalante 

Thanks for your great suggestions. 
At beginning, I also wanted to put the panel into the .list-group-item, just tried some methods, they all didn't work.  So put it outside of .list-group-item. Currently, I have put the panel into .list-group-item with flex value. Also added some value to .list-group-item as following:
- flex-wrap: wrap, it will let the panel into the second row. 
- remove the list-group-item padding, because if we add padding to .list-group-item, then the panel can't be display in width 100%.
- As list-group-item padding removed, so need to add padding to checkbox and the right action buttons.

One question here: We put the panel into .list-group-item, then we will have two rows, but the first row not in an independent 'div', so it's difficult to bind event to the first row. For example, in simple expansion page, we need to click the first row to expand the panel, but like the below picture, we can't bind event to one specific ‘div’, because there are mutilple different blocks: A, B, C, D. 
So still need your help.

list-group-item

Thanks
June

andresgalante:
Hi June,

I want to keep the structure of the list group. So each line should be a list group item. It's ok to have an extra div to generate a clicable area, but please keep it inside each item.

Also, please review bootstrap grid system, you don't need to use an empty clearfix, each row has a clearfix.

Thanks

junezhang:
@andresgalante As this PR has conflicts, so I recreated a PR for it.
https://github.com/patternfly/patternfly/pull/487

PR #444, opened 2016-08-31
Fixed the border problem in list-view page

junezhang:
## Description

Fixed this bug for when selected a row in list view page, only the top blue border will display.

Reference this link: https://github.com/patternfly/patternfly/pull/434#issuecomment-243735388

### DIFF
--- a/less/list-view.less
+++ b/less/list-view.less
@@ -10,10 +10,14 @@
     display: flex;
     padding-bottom: 0;
     padding-top: 0;
+    margin-bottom: 0;
+    border-top: 0;
     &.active {
       color: @list-group-link-color;
       background-color: @list-view-active-bg;
       z-index:auto;
+      border-top: solid 1px @list-group-active-border;
+      margin-top: -1px;
     }
     &:hover {
       background-color: @list-view-hover-bg;


less/list-view.less
```diff
@@ -10,17 +10,25 @@
...
     &:hover {
       background-color: @list-view-hover-bg;
     }
     @media (min-width: @screen-md-min) {
       align-items: center;
     }
+    &.active:first-child {
+      border-top: 1px solid @list-group-active-border;
+      margin-top: -1px;
+    }
```
jeff-phillips-18: I don't believe this block is necessary. It is using the same values as the non first-child active items.

junezhang: Because in list-group.less, we have this code:
.list-group {
  border-top: 1px solid @list-group-top-border;
  .list-group-item:first-child {
    border-top: 0;
  }
}
set the first child border top is 0,  I need the top border, so add a 1px to top border.

jeff-phillips-18: But that does not effect .list-view-pf and does not come into play here

junezhang:
@jeff-phillips-18  Thanks, Jeff. Fixed it.

dlabrecq:
@jgiardino Wasn't this an intentional change? I'm looking at #434 and it states, "with this CSS removed, only a top blue border will display for selected items (for any row except first)."

jgiardino:
The update I checked in for #434 was to address the issue with additional borders displaying in the vertical nav that was introduced with #400. #400 did not provide details about why border styles for list groups were modified, and I didn't see details anywhere else that stated what the expected visual design was for list view borders, so for #434 I pulled out the css that caused the issue since it was unrelated to the actual issue stated in #400.

I don't have a strong opinion on the visual design for the borders, and am interested in what the rest of the design team thinks this should be. 

_However, there are two points I'd like to make related to this change:_
1. When changes are made, then the PR needs to document what changes were made and the history behind those changes (or mention the JIRA story) so that other team members are aware of what is changing and why. 
2. For optimal performance, we should avoid applying styles during user interaction (e.g. hover, click) that affect layout (e.g. border-width, margin) when we can achieve the same effect by only modifying styles that do not affect layout (e.g. border color). I included more comments related to this in #434.

jgiardino:
@dlabrecq  Please review latest comments posted to #434

dlabrecq:
It wasn't clear if the border should be there or not? And if this change will have any affect the navigation menu? It looked to me that the top blue border in the navigation menu was removed intentionally.

I just wanted to ensure that adding this border back in won't override what you were trying to achieve with #434. If the change looks good to you, I'm ok with it.

junezhang:
@jgiardino Thanks for your great suggestions. 
Based on the comments from #434, visual design team suggested to use the single border. 
So close this PR.

PR #449, opened 2016-08-31
Fix to hide badges on the primary nav menu in collapse mode.

jeff-phillips-18:
Fix to vertical navigation to hide badges on the primary navigation menu when in collapsed mode.

http://rawgit.com/jeff-phillips-18/patternfly/fixes-dist/dist/tests/vertical-navigation-with-badges.html

### DIFF
--- a/less/vertical-nav.less
+++ b/less/vertical-nav.less
@@ -138,38 +138,6 @@
       width: @nav-pf-vertical-badges-width;
     }
   }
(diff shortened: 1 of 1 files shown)


jeff-phillips-18:
@srambach @andresgalante Please review.

Any chance we could get this into a patch?

andresgalante:
oh! this is a complicated pattern! 👍  I am ok with the change.

PR #451, opened 2016-09-01
Move /less files to /src/less

srambach:
## Description

This PR moves the `/less` files into `/src/less` because they are source files that are compiled into css in are more appropriately contained under the source directory. 

Relates to story [PTNFLY-1599](https://patternfly.atlassian.net/browse/PTNFLY-1599)

`*.less` path changes:
- PatternFly `*.less` files within the `/less` folder have been moved to the `/src/less` folder.
  - The `*.less` files are available for downstream consumption via the `/dist/less` folder
- 3rd party `*.less` dependencies that were present in the `/less` folder are now consumed directly from the node_modules folder

The above `*.less` path changes were updated in the gruntfile.js build file.

Note: any downstream  applications consuming patternfly `*.less` source files in their build scripts will have to update the paths accordingly.


dlabrecq:
LGTM

jeff-phillips-18:
@skateman Impacts on patternfly-sass?

skateman:
@jeff-phillips-18 if is really just move, then no worries :) i'll adjust my scripts

Gruntfile.js
```diff
@@ -163,7 +146,16 @@ module.exports = function (grunt) {
         tasks: ['jekyll']
       },
       less: {
-        files: 'less/*.less',
+        files: [
+          'src/less/*.less',
+          'node_modules/bootstrap/less/**/*',
```
bleathem: We probably don't need to watch the less files in the node_modules.  They will only change when an `npm install` is run, and one is likely to re-run `grunt build` after doing so.

bleathem:
This looks really good @srambach.  I always like PRs that result in fewer LoC in our repo :)
I have the one comment in-line re: the watch folders.

@skateman it's just a move, but also we are no longer tracking our dependency less files in our repo - you'll have to grab them from the node_modules folder.

Gruntfile.js
srambach: OK, I've removed the watch on node_modules.

bleathem:
Thanks @srambach, looks good.

:+1:

@andresgalante do you want to take a quick look at the generated CSS before we merge?

jeff-phillips-18:
@bleathem Will this cause a major release?

bleathem:
@jeff-phillips-18 not a major no.  No api's changing, no new functionality.  Just moving source files around.  Downstream consumers of our dist files will not notice a change, just those extending/building off our source directly.

jeff-phillips-18:
Please copy the less files from src/less to dist/less so they may be consumed.

bleathem:
In an off-thread discussion we've agreed that we can include this change with a minor version bump because:
1. no APIs are changing; the less files themselves are the same
2. the path move will result in a build-time failure, rather than a run-time one
3. We can mitigate the impact of this change with clear communication in the release notes, and announcement e-mail to the patternfly ML.

With this agreement on the semver consequences, and all technical issues in this PR having been addressed, it looks to me like we are good to merge.

:+1:

andresgalante:
👍 great work @srambach. thanks a lot!

priley86:
+1. great to have the less in `src` now and still consumable in `dist`. great work @srambach!

dlabrecq:
LGTM.

PR #453, opened 2016-09-02
Add missing style definitions for treeview node images

skateman:
Node icons are working properly, but 5daf134 introduced optional
node image support. For this it is necessary to have some extra
style definitions.

screenshot from 2016-09-02 09-07-06

Maybe this PR is wrong and the bootstrap-treeview.css file needs
to be included in our codebase under the lib folder and loaded
from the top level less file...

### DIFF
--- a/less/bootstrap-treeview.less
+++ b/less/bootstrap-treeview.less
@@ -44,6 +44,16 @@
       margin-right: 10px;
     }
   }
+  span.image {
+    background-repeat: no-repeat;
+    background-size: contain;
+    display: inline-block;
+    height: 1.19em;
+    line-height: 1em;
+    margin-right: 5px;
+    vertical-align: middle;
+    width: 12px;
+  }
   span.indent {
     margin-right: 5px;
   }


priley86:
@skateman i am OK with this change as I recall testing it with you before. I think it was left out because I discussed at one point with @LHinson this use case not being needed (for icons), and PF preferring fonts. 

@LHinson do you want me to introduce a new test example in Patternfly for the case we use an icon? if so, what icons? Or are we OK with adding this change for @skateman as-is (assuming it meets his needs in mIQ/ bootstrap-treeview icons)? The other option is having his downstream application extend Patternfly's css (and leaving it out of core). Thoughts?

andresgalante:
We don't want to add images on patternfly. 
@skateman what do you think about @priley86 approach?

skateman:
well we need this in MIQ, ping @epwinchell

andresgalante:
@skateman just talk to @LHinson about it, we can add it as long on our test pages we use something like https://placehold.it as placeholders for images.

I'll merge it. Thanks @skateman !

PR #461, opened 2016-09-09
PTNFLY-877 Table View Implementation.

dlabrecq:
## Description

This introduces new functionality to the data table including row selection, filtering, and an empty table state. This example was used for a gap analysis between the current design requirements and datatables.net implementation. There will be follow up stories to tweak/add functionality.

Note that there was an initial data table implementation which has been in place for some time, so many selectors were already available for table striping, borders, sorting, and hover. However, styling was added to override selectors for the toolbar, row selection, etc.

The Patternfly toolbar was used for the filtering layout, so margins may be tweaked in a later story? (Currently, there is no spec for margins, colors, etc.) JavaScript was added to implement a custom filter in the toolbar, mimicking the angular-patternfly toolbar functionality.

Note: Filtering is case sensitive and you must press the enter key.

I didn't use the select plugin from datatables.net because it does not provide support for our select all feature. It also uses CSS images for custom checkboxes, which may not be accessible? Instead, I used standard HTML checkboxes, which are closer to the design and are inherently accessible. JavaScript was also added here to implement the select all feature.

Note that the JavaScript is modular, but is located in HTML files. The intention was to inject this code directly in the page, via Jekyll, for displaying pretty print examples on pf.org.

The inline actions (i.e., button and kebab) are separated by table cells. I implemented styling to remove button borders to give the appearance of filling the entire table cell. Developers may simply omit the selector (via the datatables.net config) if they want to place something else in that table cell. Or, omit the table cell all together.

Although datatables.net provides support for the features below, they have been intentionally omitted. These features are not covered in the current design, but will be covered in future sprints.
- Pagination
- Column customization
- Expand and collapse rows
- Multi column sorting
- Sticky column headers
- CSS image checkboxes
- Tooltips

JIRA: https://patternfly.atlassian.net/browse/PTNFLY-877
Design: http://www.patternfly.org/pattern-library/content-views/table-view/#_design
## Changes

- Added checkboxes to select individual rows
- Added select all rows checkbox to select multiple rows
- Added row hovering styling
- Added inline actions and styling
- Added toolbar for filtering
- Added bulk item actions
- Added actions to delete all rows and restore rows
- Added blank slate for an empty table

https://rawgit.com/dlabrecq/patternfly/table-view-dist/dist/tests/datatables.html

This is an image:

screen shot 2016-09-09 at 2 40 10 pm
## PR checklist (if relevant)


less/datatables.less
```diff
@@ -169,50 +174,131 @@
...
-        left: 7px;
-        line-height: @font-size-base;
+.datatable-pf {
+  .blank-slate-pf {
+    background: @table-bg-accent;
+    border: 1px solid @table-border-color;
+    border-radius: 0;
+    margin-top: -1px;
+  }
+  .datatable-pf-actions {
```
srambach: Can this be pulled out to the same level as .datatable-pf to reduce the specificity? Same for the other .datatable\* classes.

less/datatables.less
```diff
@@ -169,50 +174,131 @@
...
+  .datatable-pf-select {
+    width: 33px;
+  }
+  .datatable-pf-results-right {
+    text-align: right;
+  }
+  table {
+    margin-bottom: 0;
+    max-width: none !important;
+    > tbody {
```
srambach, line 187: Unless necessary (like a td within a tr.selected) can these selectors be un-nested?

less/datatables.less
dlabrecq: Not this one as we need to override the existing blank-slate-pf selector here.

less/datatables.less
dlabrecq: I moved the datatable-pf-select and datatable-pf-results-right, but not the table and toolbar selectors (i.e., the td within a tr.selected) due to the existing specificity. This is necessary to  override those particular styles.

srambach: Ah, existing specificity...nods sadly. ;-)

jeff-phillips-18:
In the toolbar pattern, the actions are not pulled to the right (the right actions are for view selection). Using the toolbar-pf-action-right selector not only pulls the buttons to the right but does not put the correct spacing between the buttons and the kebab.

jeff-phillips-18:
Showing he kebab menus for either the toolbar or a row cause horizontal scrolling on the test page.

dlabrecq:
@jeff-phillips-18 The table view design doc shows the actions are pulled to the right, so that's what I followed. Although, I'll work on tweaking the spacing.

That said, I don't see a horizontal scroll bar on any of the browsers I've tested. Which browser are you using?

jeff-phillips-18:
@dlabrecq I'm using chrome. When you select the kebab to show the menu, the scrollbar appears.

dlabrecq:
Thanks @jeff-phillips-18. I was able to change the kebab to pull right. This prevents the kebab from opening beyond the width of the browser window; thus, eliminating the horizontal scroll bar.

I also adjusted the spacing of the bulk actions.

dlabrecq:
Although the story is complete and comments have been addressed, this PR will be committed to a feature branch. Therefore, I've changed the merge base to table-view.

andresgalante:
@dlabrecq I am sorry I haven't been able to do a proper CSS review today. There are some design details I want to address. If its ok with you, lets merge this one on the feature branch and continue in the context of the new story.

dlabrecq:
Yes, that's fine @andresgalante. I've changed the merge base for this PR so it will be merged into a feature branch named table-view.

less/datatables.less
```diff
@@ -169,50 +174,154 @@
...
-        font-family: @icon-font-name-fa;
-        font-size: (@font-size-base - 2);
-        font-weight: normal;
-        height: (@font-size-base - 3);
-        left: 7px;
-        line-height: @font-size-base;
+.datatable-pf {
+  .blank-slate-pf {
+    background: @table-bg-accent;
+    border: 1px solid @table-border-color;
```
cardosogabriel, line 180: Can we avoid shorthand declarations? Specify background-color and break the border declaration?
Same for lines 266/267
According to our code guide: http://codeguide.patternfly.org/#css-shorthand

less/datatables.less
```diff
@@ -149,8 +149,13 @@
...
 }
 
+.dataTables_empty {
+  background: @table-bg-accent;
+}
+
 .dataTables_wrapper {
-  margin: @line-height-computed 0;
+  //margin: @line-height-computed 0;
+  margin: 0;
```
cardosogabriel, line 158: What happens if there is no margin specified? 
Removed declaration and could not see a change.

less/datatables.less
```diff
@@ -169,50 +174,154 @@
...
+          }
+        }
+      }
+      > tr.selected {
+        color: @color-pf-white;
+        background-color: @color-pf-blue-400;
+        > td {
+          border-bottom-color: @color-pf-blue-500;
+        }
+        > td.datatable-pf-actions {
```
cardosogabriel, line 205: Possible to nest this class under the above using the & ?
Same thing for line 214.

less/datatables.less
```diff
@@ -169,50 +174,154 @@
...
+      }
+      .sorting_asc,
+      .sorting_desc {
+        color: @link-color !important;
         position: relative;
-        top: 2px;
-        vertical-align: baseline;
-        width: @font-size-base;
+        &:after {
+          content: @fa-var-angle-down;
```
cardosogabriel, line 234: I don't see a difference of layout by removing the attributes: font-weight, height, line-height, vertical-align and width.
Can u check if they are really necessary?

less/datatables.less
```diff
@@ -169,50 +174,154 @@
...
+        .toolbar-pf-filter {
+          width: 40%;
+        }
+      }
+    }
+    .toolbar-pf-action-right {
+      > .btn {
+        margin-left: 5px;
+      }
+      > .dropdown-kebab-pf {
```
cardosogabriel, line 289: Do we need nesting here?

less/datatables.less
```diff
@@ -169,50 +174,154 @@
...
+        margin-left: 10px;
+      }
+      float: right;
+      @media (max-width: @grid-float-breakpoint) {
+        float: right;
```
cardosogabriel, line 294: Isn't this doing the same as line 292?

less/datatables.less
```diff
@@ -169,50 +174,154 @@
...
+      }
+    }
+    .toolbar-pf-results {
+      > div {
+        padding-left: 10px;
+        padding-right: 10px;
+      }
+      border-top: 1px solid @table-border-color;
```
cardosogabriel, line 302: Is there a reason to put first a child class declaration and later put the attributes of the class?
If the go direct under the class is easier to read the code.

less/datatables.less
```diff
@@ -169,50 +174,154 @@
...
+      height: 40px;
+      line-height: 40px;
+    }
+  }
+}
+
+.datatable-pf-actions {
+  background: @table-bg-accent;
```
cardosogabriel, line 310: Please use background-color

less/datatables.less
```diff
@@ -169,50 +174,154 @@
...
+  > .btn {
+    border: none;
+    box-shadow: none;
+    width: 100%;
+  }
+  .dropdown-kebab-pf {
+    padding-right: @table-cell-padding;
```
cardosogabriel, line 317: Also needs a padding-left here to avoid breaking in smaller viewport.

screenshot

cardosogabriel:
Added some comments in the CSS code.
The "Actions" buttons are not filling 100% of the height of the td, this can be seen better in smaller viewports:
smaller viewport
Also suggest to decrease the th/td line-height (which is too big for more than one line of text), and adjust the distances to the borders with padding.

@dlabrecq , are we suppose to support mobile viewport? In this case it is broken.

dlabrecq:
Thanks, Gabreil. There is a responsive plugin for datatable.net that I have not experimented with, which may better handle the line height? I'll try to address the action button height and padding in the feature branch.

andresgalante:
@cardosogabriel I'll merge this, we will continue in the new branch so @dlabrecq can close his story.

thanks!

less/datatables.less
dlabrecq: The existing toolbar-pf-action-right selector sets this to "float: none" for max-wdith. I need to ensure this always floats right.

less/datatables.less
dlabrecq: Yes, due to the specificity defined by the toolbar.

less/datatables.less
dlabrecq: Maybe? I'll try to use background-color, but the toolbar-pf selector uses background. The blank-slate-pf selector also uses border -- need to override the color here.

less/datatables.less
dlabrecq: I don't believe so. In order to override row highlighting, this was the only way I could get it to work, due to the specificity of existing styles.

less/datatables.less
dlabrecq: This was pre-existing code, so I'll have to experiment. Thanks for pointing this out.

less/datatables.less
dlabrecq: No reason. Will move it.

cardosogabriel:
Sounds good. Thanks for the responses @dlabrecq!

less/datatables.less
dlabrecq: FYI, I used background-color and split the border declaration for lines 266/257. I suspect line 180 is ok, considering all borders are being set. I see we're using the border declaration in many places; toolbar, vertical-nav, tables, card-view, etc.

After I address all comments, I'll send out a separate PR with these changes.

less/datatables.less
dlabrecq: FYI, I don't know much about this previously existing declaration, but it no longer seems to have any effect. Thus, I've commented it out for now. At some point we may need this when adding pagination controls?

less/datatables.less
dlabrecq: Yes, I was able to nest this class using &. Thanks!

less/datatables.less
dlabrecq: FYI, I don't observe any difference when these attributes are removed. However, I will leave them commented out in case cross browser testing shows an issue. So far, FF, Chrome, Safari, and IE11 look good.

less/datatables.less
dlabrecq: FYI, I confirmed that we must override the max-width declaration on lone 294, but removed the "float: right" on line 292. The toolbar-pf-action-right selector already defines "float: right", so that's redundant.

less/datatables.less
dlabrecq: Done.

Issue #296, opened 2016-05-26
table.table-bordered.table-hover tables lose bottom-border on hover


rhamilto:
Before:

![screen shot 2016-05-26 at 1 56 03 pm](https://cloud.githubusercontent.com/assets/895728/15584711/be6f9be6-2349-11e6-8276-28e93f72fb53.PNG)

See http://rawgit.com/patternfly/patternfly/v3.3.5/tests/tables.html

After:

![screen shot 2016-05-26 at 1 56 14 pm](https://cloud.githubusercontent.com/assets/895728/15584725/cefc9d56-2349-11e6-91e5-df927dbf7b94.PNG)

See http://rawgit.com/patternfly/patternfly/v3.4.0/tests/tables.html

https://github.com/patternfly/patternfly/commit/8f10bfa8870670aeed2f31b95a1aef0e53301f02#diff-a7693b847cd07d7aa4c819d11e5becc1L192 is the offending change.  

Attention:  @cardosogabriel, @andresgalante.

I found this bug when updating OpenShift to v3.4.0.  @jwforres, @spadgett:  is this a blocker for updating?  We use table.table-bordered.table-hover on browse/\* pages in OpenShift.

andresgalante:
@rhamilto thanks, I'll send a PR today

jwforres:
@rhamilto i'd wait for the fix, we've got plenty of time.  i don't think we urgently need anything we'd be picking up from the new version, @spadgett?

spadgett:
> i'd wait for the fix, we've got plenty of time

+1

PR #472, opened 2016-09-15
ListView example shows donut charts incorrectly

dlabrecq:
## Description

Fix for a bug Liz B. encountered on pf.org, where the ListView code example (variant # 1) show donut charts located off the page.

The problem is that the chart widths are auto-generated incorrectly. We've been setting height, but not width.

There should be no change for the test page itself. The difference will be best seen on pf.org, after this update.
## Changes
- Set the width of donut charts in list view example

https://rawgit.com/dlabrecq/patternfly/list-view-fix-dist/dist/tests/list-view-rows.html

This is an image from pf.org:

screen shot 2016-09-15 at 12 46 32 pm

### DIFF
--- a/tests/pages/_includes/widgets/list-view/list-view-variation-1.html
+++ b/tests/pages/_includes/widgets/list-view/list-view-variation-1.html
@@ -41,6 +41,7 @@
   var c3ChartDefaults = $().c3ChartDefaults();
   var customDonutChartConfig = c3ChartDefaults.getDefaultDonutConfig('34%');
   customDonutChartConfig.size.height = 71;
+  customDonutChartConfig.size.width = 60;
   customDonutChartConfig.donut.width = 5;
   customDonutChartConfig.bindto = '#{{include.id1}}';
   customDonutChartConfig.data = {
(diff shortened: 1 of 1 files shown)


dlabrecq:
@cardosogabriel Would you please review this one? This is related to the #296 issue for pf.org you recently reviewed.

cardosogabriel:
Your changes look good. Only thing is that we are specifying 67px of height, and the charts are about 58px height. This makes them not be in the middle (see attachment). Perhaps this needs to be addressed in a new issue.

donuts

dlabrecq:
@cardosogabriel There may be some extra height there, but lets save that for another issue.

For this PR, I'm only setting the width property to 60 just to get it to display properly. It's currently hanging off the page edge on pf.org.

Please merge?

cardosogabriel:
> There may be some extra height there, but lets save that for another issue.

Ok. I don't have merging privileges for this project. It needs to be @andresgalante or @bleathem .

jeff-phillips-18 (approved):
LGTM

PR #473, opened 2016-09-16
PNTLY-1829 Required fields for forms.

cardosogabriel:
## Description

Initial implementation for Forms: required fields.
https://patternfly.atlassian.net/browse/PTNFLY-1829
## Changes
- This PR proposes a reorganisation of the forms test page. With the addition of 6 more different forms, there was a need to better place all the elements of the page. Please give feedback.
- Introduced 6 new forms: 3 form variations for right-aligned labels and 3 more for top-aligned labels.

[link to Forms on rawgit](http://rawgit.com/cardosogabriel/patternfly/PTNFLY-1829-dist/dist/tests/forms.html)
## Considerations about the solution for right-aligned labels
- Since the optional needs to be in a new line, it has a display: block. Given that we don't know what is the breakpoint set by the developer using patternfly, I cannot specify when this element should go inline, which is the desirable behavior for when the label go above the input. So the result is that even when the label is above the input, the optional appears in a new line, and without the ().
- Placing the optional below the label generates a bigger vertical space between the inputs, breaking the rhythm of the design. I could set a negative margin in case of having optional, that would help specially when the label fills only one line, keeping the rhythm between the inputs but having less space between optional and next label.

negative margin
## PR checklist (if relevant)

Will still test.

### DIFF
--- a/src/less/forms.less
+++ b/src/less/forms.less
@@ -10,6 +10,11 @@
   color: @brand-danger;
 }
 
+.fields-status-pf {
+  color: @color-pf-black-500;
+  margin-bottom: 15px;
+}
+
 .form-control {
   &[disabled],
   &[readonly],
@@ -58,4 +63,13 @@
 
 label {
   font-weight: 600;
+  &.required-pf:after {
+    color: @brand-danger;
+    content: "*";
+    margin-left: 3px;
+  }
+}
+
+span.required-pf {
+  color: @brand-danger;
 }
--- a/tests/pages/forms.html
+++ b/tests/pages/forms.html
@@ -4,67 +4,136 @@ layout: page
 title: Forms
 resource: true
 ---
-      <h2>PatternFly Example</h2>
-      <form class="form-horizontal">
+      <h2>Right-aligned labels</h2>
+
+      <h3>Basic Form</h3>
+      <form class="form-horizontal" role="form">
         <div class="form-group">
-          <label class="col-sm-2 control-label" for="textInput">Default</label>
+          <label for="inputEmail1" class="col-sm-2 control-label">Email</label>
           <div class="col-sm-10">
-            <input type="text" id="textInput" class="form-control"></div>
+            <input type="email" class="form-control" id="inputEmail1">
+          </div>
         </div>
         <div class="form-group">
-          <label class="col-sm-2 control-label" for="textInputDisabled">Disabled</label>
+          <label for="inputPassword1" class="col-sm-2 control-label">Password</label>
           <div class="col-sm-10">
-            <input type="text" id="textInputDisabled" class="form-control" disabled></div>
+            <input type="password" class="form-control" id="inputPassword1">
+          </div>
         </div>
-        <div class="form-group has-error">
-          <label class="col-sm-2 control-label" for="inputError">With error</label>
-          <div class="col-sm-10">
-            <input type="text" id="inputError1" class="form-control">
-            <span class="help-block">Please correct the error</span>
+        <div class="form-group">
+          <div class="col-sm-offset-2 col-sm-10">
+            <div class="checkbox">
+              <label>
+                <input type="checkbox"> Remember me
+              </label>
+            </div>
+          </div>
+        </div>
+        <div class="form-group">
+          <div class="col-sm-offset-2 col-sm-10">
+            <button type="submit" class="btn btn-default">Sign in</button>
           </div>
         </div>
       </form>
-      <h2>Count Remaining Characters Control</h2>
-      <form>
+
+      <h3>All Fields Required</h3>
+      <form class="form-horizontal" role="form">
+        <p class="fields-status-pf">All fields are required.</p>
         <div class="form-group">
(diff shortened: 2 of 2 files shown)


andresgalante:
@cardosogabriel thanks for doing this.

@LHinson, @kybaker and @jsniezek: I think this proves the point I was trying to make. An implementation of the optional label under the input label means that there will be an extra vertical space between lines.

This creates a jump in the flow of the form, that can lead to a lower conversion rate.

I hope its clearer now why I didn't want to follow this option. What do you think?

cardosogabriel:
Updated the PR to use "optional" inside the input as placeholder, as requested by Leslie in the issue.

LHinson:
I forgot about the use case that @jsniezek mentions in https://patternfly.atlassian.net/browse/PTNFLY-1829

He says... 

> "As mentioned before, the edge case to consider here with Option B is when a form has a field that's optional and also has placeholder text. If people really need to do that, we could consider recommending that they combine the text into a single string, like: "enter your bank account number here (optional)". WDYT?" 

I think this is a common use case and a cause for concern. Therefore, what do you think about the following approach... 
- If only some fields are required, you mark all required fields with "*" 
- If all fields are required, you make the note at the top and no other indication is needed (as seen in the test page now) 
- If all fields are optional, you make the note at the top and no other indication is needed (as seen in the test page now) 

WDYT @andresgalante @cardosogabriel @jsniezek

cardosogabriel:
> If only some fields are required, you mark all required fields with "*"

Once I learned that a good approach was to use \* if a few required, and (optional) if a few optional.
Since we didn't find a robust solution for optional, I'm ok with the *

jgiardino:
I'm also in favor of using the "*" to indicate required fields. I know I have seen implementations of this method where additional text "Required" is associated with the "*" so that screenreaders will read "required" along with reading the field label.

cardosogabriel:
@LHinson, shall we go with the "*"?

andresgalante:
👍  to go "*"

cardosogabriel:
Updated the rawgit link using the \* as required fields. Please check.

LHinson:
@jgiardino @andresgalante will you take a look at this?

src/less/forms.less
```diff
@@ -58,4 +63,13 @@
 
 label {
   font-weight: 600;
+  &.required-pf:after {
```
andresgalante, line 66: @cardosogabriel Do we need to nest this class?

tests/pages/forms.html
```diff
@@ -4,67 +4,136 @@
...
+
+      <h3>Required and Optional Fields</h3>
+      <form class="form-horizontal" role="form">
+        <p class="fields-status-pf">The fields marked with <span class="required-pf">*</span> are required.</p>
         <div class="form-group">
-          <label for="messageArea"></label>
-          <textarea class="form-control" id="messageArea" name="text" placeholder="Type in your message" rows="5"></textarea>
+          <label for="input4" class="col-sm-2 control-label required-pf">Field One</label>
+          <div class="col-sm-10">
+            <input type="text" class="form-control" id="input4">
```
andresgalante: @cardosogabriel we are missing a `required` attribute on this input.

tests/pages/forms.html
```diff
@@ -4,67 +4,136 @@
...
-      <h2>Count Remaining Characters Control</h2>
-      <form>
+
+      <h3>Required and Optional Fields</h3>
+      <form class="form-horizontal" role="form">
+        <p class="fields-status-pf">The fields marked with <span class="required-pf">*</span> are required.</p>
         <div class="form-group">
-          <label for="messageArea"></label>
-          <textarea class="form-control" id="messageArea" name="text" placeholder="Type in your message" rows="5"></textarea>
+          <label for="input4" class="col-sm-2 control-label required-pf">Field One</label>
```
andresgalante, line 71: I wish there was a way to select the `label` that is before the element that has the attribute `required` with CSS. anyway.... 😄

andresgalante:
@LHinson @cardosogabriel I've just made 2 small comments, the rest looks great.
Thanks a lot Gabriel

src/less/forms.less
cardosogabriel: Yes, because there is a standalone class "required-pf" used for the top information "The fields marked with..." that has a different behavior. So, only when this class is in a label that we want these styles.
We could go for a different class for the label, but I wanted to make it easier for users and use the same class in both. Let me know if u think it is better to have a specific class here.

andresgalante: perfect, sorry about it, I missed the other class.

tests/pages/forms.html
cardosogabriel: I was thinking about this, but we only show the \* when the form has some elements required and others no. So this selection would not work for full required forms ;)

tests/pages/forms.html
cardosogabriel: Added to this and all other required fields.

andresgalante:
@cardosogabriel thanks a lot!

PR #500, opened 2016-10-06
DataTables extension for row selection

dlabrecq:
## Description

Created a DataTables extension, named pfSelect, to provide row selection functionality using traditional HTML checkboxes. Note that the DataTables "select" extension is used with pfSelect to maintain row selections and provide complex selection behaviors using ctrl/cmd, shift, and an unmodified click.

The following selection styles are supported for user interaction with the DataTable:
- api - Selection can only be performed via the API
- multi - Multiple items can be selected
- multi+shift - a hybrid between the os style and multi
- os - Operating system style selection with complex behaviours such as ctrl/cmd, shift and an unmodified click
- single - Only a single item can be selected

https://patternfly.atlassian.net/browse/PTNFLY-1843
## Changes

- Created a DataTables extension, named pfSelect
- Removed previous select.js utility from example page

https://rawgit.com/dlabrecq/patternfly/table-view-dist/dist/tests/datatables.html

This is an image:

screen shot 2016-10-06 at 12 34 30 am
## PR checklist (if relevant)


cardosogabriel:
Looks good to me, but I'm not a JS expert, so I looked more into the HTML.
If we are considering accessibility, every checkbox should have a label, so I'd suggest to update the HTML. We can hide the label for visual purposes though. https://www.w3.org/WAI/tutorials/forms/labels/

bleathem:
I did a walk through of the approach and code with @dlabrecq.  I'm really happy with the new patternfly extension, and how it leverages the functionality from the existing datatables.net extension.

dlabrecq:
@cardosogabriel Thanks for the suggestion. When elements are hidden, they are also inaccessible to screen readers. However, I believe the label can be placed in the second column of the table -- the second column should be relevant, tho. For example:

&lt;tr>
  &lt;td>
    &lt;input type="checkbox" name="select" id="r1" />
  &lt;/td>
  &lt;td>
  &lt;label for="r1">Geko&lt;/label>
  &lt;/td>
&lt;/tr>

I will introduce this with the next PR.

cardosogabriel:
Thank you @dlabrecq . The sr-only approach works well.

dlabrecq:
Yes, Andres mentioned that. Thanks, again.

PR #501, opened 2016-10-06
Datatables.net

rhwood:
## Description

Use the `datatables.net` package instead of the `datatables` package for [Datatables](http://datatables.net), since all installations instructions for Datatables use the `datatables.net` package.
## Changes
- Use `datatables.net` in Bower and NPM
- Use Datatables Buttons extension instead of the (deprecated) Datatables Colvis extension
- Provide default CSS for Datatables Buttons including the Datatables Colvis Button menu

This is an image:

Table with one column hidden
## PR checklist (if relevant)

---

This change is [Reviewable](https://reviewable.io/reviews/patternfly/patternfly/501)


dtaylor113:
Hi, so this doesn't work in IE?

rhwood:
It's just not tested yet, and is not known to not work.

bleathem:
This is great, thanks for the PR @rhwood!  Your changes look reasonable to me.

@dlabrecq is leading our datatables.net implementation atm, we'll wait for his review before merging.

rhwood:
@dtaylor113 I've tested this on IE 11 running in Edge (default), 10, and 9 emulation modes without issue. I've only marked it as working in IE 11, though since I don't know the protocols for declaring a browser tested.

src/less/datatables.less
```diff
@@ -54,7 +40,7 @@
   .btn {
     .box-shadow(none);
   }
-  .ColVis {
+  .dt-buttons {
```
dlabrecq, line 43: Do we want all .dt-buttons positioned absolutely or only .dt-button.buttons-columnVisibility?

karma.conf.js
```diff
@@ -20,7 +20,7 @@ module.exports = function(config) {
       'node_modules/patternfly-bootstrap-combobox/js/bootstrap-combobox.js',
       'node_modules/bootstrap-datepicker/dist/js/bootstrap-datepicker.js',
       'node_modules/bootstrap-select/dist/js/bootstrap-select.js',
-      'node_modules/datatables/media/js/jquery.dataTables.js',
+      'node_modules/datatables.net/js/jquery.dataTables.js',
```
dlabrecq, line 23: Yes! I recently changed the same line in my own code to use datatables.net.

QUICKSTART.md
```diff
@@ -124,9 +126,11 @@ We provide compiled CSS and JS (`patternfly.*`), as well as compiled and minifie
...
         <!-- Datatables, jQuery Grid Component -->
         <!-- Note: jquery.dataTables.js must occur in the html source before patternfly*.js.-->
-        <script src="bower_components/datatables/media/js/jquery.dataTables.min.js"></script>
-        <script src="bower_components/datatables-colvis/js/dataTables.colVis.js"></script>
-        <script src="bower_components/datatables-colreorder/js/dataTables.colReorder.js"></script>
+        <script src="bower_components/datatables.net/js/jquery.dataTables.min.js"></script>
+        <script src="bower_components/datatables.net-buttons/js/dataTables.buttons.js"></script>
+        <script src="bower_components/datatables.net-buttons-bs/js/buttons.bootstrap.js"></script>
+        <script src="bower_components/datatables.net-buttons/js/buttons.colVis.js"></script>
+        <script src="bower_components/datatables.net-colreorder/js/dataTables.colReorder.js"></script>
```
dlabrecq, line 133: @bleathem @LHinson I don't know that we want to add this to our quick start? The functionality to hide/show columns is not an officially supported design pattern.

I was thinking this example could be removed after the feature branch is finally merged?

bower.json
```diff
@@ -28,9 +28,11 @@
...
     "c3": "~0.4.11",
     "d3": "~3.5.17",
-    "datatables": "~1.10.12",
-    "datatables-colreorder": "~1.3.2",
-    "datatables-colvis": "~1.1.2",
+    "datatables.net": "~1.10.12",
+    "datatables.net-colreorder": "~1.3.2",
+    "datatables.net-colreorder-bs": "~1.3.2",
+    "datatables.net-buttons": "~1.2.2",
+    "datatables.net-buttons-bs": "~1.2.2",
```
dlabrecq, line 35: @bleathem Do we want to introduce new dependencies for features that have no design pattern?

src/less/datatables.less
```diff
@@ -2,46 +2,32 @@
 // Datatables
 // --------------------------------------------------
 
-.ColVis_Button {
```
dlabrecq, line 5: @bleathem @andresgalante Does the removal of these classes imply a major version number bump? Should we leave the old classes as is and somehow mark them as deprecated?

I don't know that the DataTable was officially supported, but thought to ask in case someone may be using these classes?

dlabrecq (changes requested):

src/less/datatables.less
rhwood: Datatables allows a buttons div with the class .dt-buttons to be created using the dom option when creating a table object where all buttons tied to the table are located. Absolutely positioning the button group does appear to match the [table view](https://www.patternfly.org/pattern-library/content-views/table-view/#_) design.

src/less/datatables.less
rhwood: These classes support an object (the ColVis extension) that requires the datatables package (but not the datatables.net package). The ColVis extension is apparently no longer being maintained with the recommendation that it be replaced with the DataTables column visibility button.

I've had problems mixing packages that require datatables and packages that require datatables.net in the same app, since they fight each other.

src/less/datatables.less
dlabrecq: Ok, the absolute position seems fine. Although, the toolbar design is a little different than what datatables.net can provide via this mechanism (e.g., I don't know that it can be used to create a kebab)? The design pattern has (optional) action buttons and kebabs on the top while filter results and selected rows text appear below.

rhwood: Are those showing a button within a button? What's the code that was used (they do need to be spaced apart, but also should not be doubly wrapped)?

bower.json
bleathem: We don't want to introduce these dependencies, no.  We could consume the dependencies in our example via CDN, and users could add dependencies to their own apps if they want the functionality.

QUICKSTART.md
bleathem: Agreed we should consider how this functionality can be incorporated into our pattern from a design perspective.

@LHinson do we have any design stories in the pipeline to address column re-ordering?

src/less/datatables.less
bleathem: The divergence of the patternfly patternfly toolbar design and the datatbles.net toolbar implementation is unfortunate.  @LHinson can I suggest a design story to consider if the two approaches can be better aligned?

dlabrecq: @bleathem @LHinson FYI... The DataTables.net has its own, simple toolbar for placing its pre-canned buttons and filter input elements. It's located similarly, but the Patternfly toolbar layout is far more complex with a custom filter, multiple rows etc. Therefore, we just droped-in our own HTML from the existing toolbar examples (i.e., for the feature branch).

Note that we can still create buttons to control DataTables.net features, so using the pre-canned buttons isn't absolutely necessary. We can still show/hide columns within our own toolbar via the DataTables API.

I don't know that we want to make the pre-canned DataTables.net buttons work with our custom toolbar? Although, a design pattern would help. We may not want to use a button to hide/show columns at all?

src/less/datatables.less
bleathem: We shouldn't build further on these classes, but we shouldn't remove them either in case a downstream user is consuming them.  We can mark them for removal in PF 4.

src/less/datatables.less
bleathem: @LHinson reminded me that we have a design constraint to co-ordinate the toolbars amongst our various patterns (eg. list view, toolbar, table view).  This rules out using the datatables.net toolbar.  Neither do I see any need to support the datatables.net buttons in our toolbar.

QUICKSTART.md
bleathem: We are tracking the column visibility with this story:
https://patternfly.atlassian.net/browse/PTNFLY-1489

bleathem:
Thanks again for taking the time to prepare and submit this PR @rhwood.  

We have made a design decision to use our own [custom toolbar](http://www.patternfly.org/pattern-library/forms-and-controls/toolbar) to achieve consistency across our patterns, therefore it doesn't make sense for us to consume the datatables.net buttons at this point.

We are currently re-working our toolbar ([PTNFLY-1893](https://patternfly.atlassian.net/browse/PTNFLY-1893)) to achieve this desired consistency.  We will also update this example when we resolve [PTNFLY-1489](https://patternfly.atlassian.net/browse/PTNFLY-1489).

bleathem:
@rhwood the table view branch that @dlabrecq has been working on is being proposed for a merge back to the master branch via the PR https://github.com/patternfly/patternfly/pull/530.  This PR introduces (among other things) the patternfly toolbar to the table.  Feel free to review the PR and comment, additional insight is always appreciated.

rhwood:
I see that pagination is removed. I'm dealing with data sets on the order of 15,000-65,000 items. Is there a new story on how large data sets like should be dealt with, or is pagination only temporarily removed?

LHinson:
@rhwood it's just temporarily removed... we are in the process of designing pagination now. You can check out a blog post about that here: https://blog.patternfly.org/data-view-pagination-options-conceptual-design/

We are looking to wrap up that design this sprint in hopes of being able to implement it next sprint.

dlabrecq:
Not removed, just not added to this new example. The older example is still available with pagination in the test pages.

LHinson:
Thanks for the clarification @dlabrecq. @rhwood just note we have a new design in the works.

PR #487, opened 2016-09-19
PTNFLY-452 List View:Expand Row - PatternFly Core

junezhang:
## Description

Create simple expansion and compound expansion page for list view. 
https://patternfly.atlassian.net/browse/PTNFLY-452
## Changes
- Add flex-wrap: wrap to .list-group-item, let the expand panel be the second row. 
- In simple expansion, user need to click the whole row to expand panel, so need to add an extra div to bind event.
- When expand panel, the first row need to have an active class, it is different with selected active class, selected active class is blue background-color, but expand active is gray color. Detail Information please refrence this story, [PTNFLY-452](https://patternfly.atlassian.net/browse/PTNFLY-452), so add a class named "expand-active". Put an expand panel in the .list-group-item, it has a problem about the padding. .list-group-item has a padding value, so the expand panel looks like having a margin value, so when expand a row, I set the .list-group-item padding to be 0. and add padding to the left checkbox and the right action button.
- Create these pages:
  - tests/pages/_includes/widgets/list-view/list-view-page-compound-expansion.html
  - tests/pages/_includes/widgets/list-view/list-view-page-simple-expansion.html
  - tests/pages/_includes/widgets/list-view/list-view-standard-rows-compound-expansion.html
  - tests/pages/_includes/widgets/list-view/list-view-standard-rows-simple-expansion.html
  - tests/pages/list-view-compound-expansion.html
  - tests/pages/list-view-simple-expansion.html
## Link to image

screen shot 2016-09-19 at 3 55 50 pm
screen shot 2016-09-19 at 3 56 05 pm
## PR checklist (if relevant)
- [x] **Cross browser**: works in IE9
- [x] **Cross browser**: works in IE10
- [x] **Cross browser**: works in IE11
- [ ] **Cross browser**: works in Edge
- [x] **Cross browser**: works in Chrome - 53.0.2785.116 (64-bit)
- [x] **Cross browser**: works in Firefox - 48.0.1
- [x] **Cross browser**: works in Safari - 9.1.1 (11601.6.17)
- [x] **Cross browser**: works in Opera - 39.0.2256.71
- [x] **Responsive**: works in extra small, small, medium and large view ports.
- [x] **Preview the PR**: An image or a URL for designer to preview this PR is provided.

PR #502, opened 2016-10-10
Expand row

jgiardino:
## Description

Made requested changes to list view expansion, and applied transparent borders to collapsed state so that contents do not shift when borders are applied on expanded state

NOTE: I have no idea why so many files were included in this PR. Let me know if I need to submit a different pull request.
## Changes

## PR checklist (if relevant)


dgutride:
It appears you merged master-dist into your branch (which added all of this dist files, only upstream master should be rebased in typically).  We don't check those in, only that first commit you added should have been pushed to this branch .

PR #503, opened 2016-10-11
Add style for selected items in dropdown to match bootstrap-select

jeff-phillips-18:
This PR adds the selected styling for dropdowns to match the styling set for bootstrap-select objects. Updated test pages to set the selected state on dropdowns where selection is applicable.

This fixes issue https://github.com/patternfly/angular-patternfly/issues/318

See the filter and sort dropdowns at http://rawgit.com/jeff-phillips-18/patternfly/fixes-dist/dist/tests/toolbar.html

@andresgalante @srambach

### DIFF
--- a/src/less/dropdowns.less
+++ b/src/less/dropdowns.less
@@ -78,6 +78,17 @@
     border-color: transparent;
   }
 }
+
+// Selected state
+.dropdown-menu > .selected > a {
+  background-color: @dropdown-link-active-bg;
+  border-color: @dropdown-link-active-border-color;
+  color: @color-pf-white;
+  small {
+    color: fade(@color-pf-white, 50%);
+  }
+}
+
 // Nuke hover/focus effects
 .dropdown-menu > .disabled > a {
   &:hover,
--- a/tests/pages/_includes/widgets/cards/dashboard-timeframe-footer.html
+++ b/tests/pages/_includes/widgets/cards/dashboard-timeframe-footer.html
@@ -11,6 +11,7 @@
         Last 30 Days <span class="caret"></span>
       </button>
       <ul class="dropdown-menu dropdown-menu-right" role="menu">
+        <li class="selected"><a href="#">Last 30 Days</a></li>
         <li><a href="#">Last 60 Days</a></li>
         <li><a href="#">Last 90 Days</a></li>
       </ul>
--- a/tests/pages/_includes/widgets/cards/dashboard-timeframe-header.html
+++ b/tests/pages/_includes/widgets/cards/dashboard-timeframe-header.html
@@ -5,6 +5,7 @@
         Last 30 Days <span class="caret"></span>
       </button>
       <ul class="dropdown-menu dropdown-menu-right" role="menu">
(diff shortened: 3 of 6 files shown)


src/less/dropdowns.less
```diff
@@ -78,6 +78,14 @@
     border-color: transparent;
   }
 }
+
+// Selected state
+.dropdown-menu > .selected > a {
```
andresgalante, line 83: Hi @jeff-phillips-18 Do we need the child selectors?

src/less/dropdowns.less
```diff
@@ -78,6 +78,14 @@
...
+  background-color: @color-pf-blue !important;
```
andresgalante: also, do we need the importants?

src/less/dropdowns.less
jeff-phillips-18: Yes, the structure is set such that selected is added to the li item and the style needs to be set on the a. This is similar to boostrap-select.less:
https://github.com/patternfly/patternfly/blob/master/src/less/bootstrap-select.less#L58

src/less/dropdowns.less
jeff-phillips-18: I believe so. Again, this is very similar to bootstrap-select.less:

https://github.com/patternfly/patternfly/blob/master/src/less/bootstrap-select.less#L58

src/less/dropdowns.less
andresgalante: right, ok then. We'll have to work on improving our selectors as we move forward with pf4.

Thanks Jeff 😄  you are awesome as always.

src/less/dropdowns.less
andresgalante: Maybe we don't need them here and the specificity of the class is enough, can you try it without the importants?

jeff-phillips-18:
I've updated to remove the !important and use the same color variables as bootstrap-select.

andresgalante:
@jeff-phillips-18 great, thanks a lot 👍 

@srambach if its ok with you please merge it.

src/less/dropdowns.less
```diff
@@ -78,6 +78,17 @@
     border-color: transparent;
   }
 }
+
+// Selected state
+.dropdown-menu > .selected > a {
+  background-color: @dropdown-link-active-bg;
+  border-color: @dropdown-link-active-border-color;
+  color: @color-pf-white;
+  small {
```
srambach, line 87: Pardon my ignorance - is this a media query? Why does the color change here? I'm probably missing it in the spec somewhere but wondered.

jeff-phillips-18: I can remove it. I only added it since it is there for bootstrap-select. Shall I?

andresgalante: @srambach thanks for the review. That's correct, the small text is faded. check it out here when you open the drop down:
https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/bootstrap-select.html
screen shot 2016-10-12 at 8 48 48 am

andresgalante:
Thanks a lot @jeff-phillips-18 !

PR #504, opened 2016-10-11
Expand row

jgiardino:
## Description

Compared updates I made with updates June made, and merged our conflicts.

## Changes
- For specific details about CSS changes, refer to the callouts in the images below.

Examples Before and After Updates
1. I modified the border styles so that left and right blue borders are applied to the expanded state.
2. I also modified the border styles to avoid the double border that occurs when one list row is expanded, and the next list row is expanded or selected.
3. I replaced the <ul> implementation with <dl class="dl-horizontal">

listviewcssupdates-before-callouts

listviewcssupdates-after-callouts
## PR checklist (if relevant)


andresgalante:
This PR is a continuation of #487

andresgalante:
Rawgit links are:
https://rawgit.com/jgiardino/patternfly/expandRow-dist/dist/tests/list-view-compound-expansion.html
and
https://rawgit.com/jgiardino/patternfly/expandRow-dist/dist/tests/list-view-simple-expansion.html#

src/less/list-view.less
```diff
@@ -230,5 +251,62 @@
...
+  background: #fff;
+  border-top: solid 1px @list-group-active-border;
+  box-sizing: content-box;
+  margin: -1px -15px 0;
+  order: 3;
+  padding: 10px 15px;
+  position: relative;
+  width: 100%;
+}
+.list-view-pf-expand-close {
```
junezhang, line 298: We can remove ".list-view-pf-expand-close" ".list-view-pf-expand-chart"
In the html page, using the below code snippet based on Andres's comment.

```
<div class="close">
    <span class="pficon pficon-close"></span>
</div>
```

junezhang:
@jgiardino Thanks for your update. Your code is helpful for me. and all is okay for me. just one minor point I left comment in the css code.

junezhang (approved):

src/less/list-view.less
andresgalante: 👍

LHinson:
@kybaker do you think we need to include additional stories to figure out how to get that blue highlight detail underneath or is the design ok as implemented? 

@jgiardino I've compared the implementation with the pattern page and it matches (minus what you've already mentioned). Therefore LGTM. 

Looping in @dgutride since he will be running with this. Also would be good to have Matt C. check it since he was the designer on this. I'll loop in Matt... can't find him on GitHub atm.

src/less/list-view.less
jgiardino: Good catch! Thanks for reviewing this!

jgiardino:
Closing this one, Created #507 instead

PR #507, opened 2016-10-12
List View Expand row

jgiardino:
## Description

Compared updates I made with updates June made, and merged our conflicts.

## Changes
- For specific details about CSS changes, refer to the callouts in the images below.
## Rawgit and Before/After Images

Rawgit links are:
https://rawgit.com/jgiardino/patternfly/expandRow4-dist/dist/tests/list-view-compound-expansion.html
and
https://rawgit.com/jgiardino/patternfly/expandRow4-dist/dist/tests/list-view-simple-expansion.html#

Examples Before and After Updates
1. I modified the border styles so that left and right blue borders are applied to the expanded state.
2. I also modified the border styles to avoid the double border that occurs when one list row is expanded, and the next list row is expanded or selected.
3. I replaced the <ul> implementation with <dl class="dl-horizontal">

listviewcssupdates-before-callouts

listviewcssupdates-after-callouts
## PR checklist (if relevant)

### DIFF
--- a/tests/pages/list-view-compound-expansion.html
+++ b/tests/pages/list-view-compound-expansion.html
@@ -1,0 +1,11 @@
+---
+categories: [Layouts]
+css-extra: false
+layout: layout
+resource: true
+full-page: true
+title: List View - Compound Expansion
+url-js-extra: ['//cdnjs.cloudflare.com/ajax/libs/c3/0.4.11/c3.min.js', '//cdnjs.cloudflare.com/ajax/libs/d3/3.5.17/d3.min.js']
+
+---
+{% include widgets/list-view/list-view-page-compound-expansion.html %}
--- a/src/less/list-view.less
+++ b/src/less/list-view.less
@@ -8,8 +8,14 @@
     align-items: flex-start;
     .clearfix(); //IE9 fallback
     display: flex;
+    flex-wrap: wrap;
     padding-bottom: 0;
     padding-top: 0;
+    &.expand-active {
+      background-color: @list-view-hover-bg;
+      border: solid 1px @list-group-active-border;
+      padding: 0;
+    }
     &.active {
       color: @list-group-link-color;
       background-color: @list-view-active-bg;
@@ -63,6 +69,9 @@
   }
   .list-view-pf-top-align & {
     align-self: flex-start;
+  }
+  .expand-active & {
+    margin-right: 15px;
   }
 }
 .list-view-pf-additional-info {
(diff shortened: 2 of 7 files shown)


dlabrecq:
The bottom border is missing on larger screens. Although, perhaps this can be fixed later?

screen shot 2016-10-12 at 11 05 06 am

tests/pages/list-view-compound-expansion.html
```diff
@@ -0,0 +1,11 @@
+---
+categories: [Layouts]
+css-extra: false
+layout: layout
+resource: true
+full-page: true
+title: List View - Compound Expansion
+url-js-extra: ['//cdnjs.cloudflare.com/ajax/libs/c3/0.4.11/c3.min.js', '//cdnjs.cloudflare.com/ajax/libs/d3/3.5.17/d3.min.js']
```
jeff-phillips-18, line 8: I don't believe the compound expansion example uses C3/D3 so these can be removed.

dlabrecq:
Also wanted to note that clicking on the kebab (or almost anywhere in that row) will collapse/expand the pane. I had expected to only be able to click the arrow toggle.

jeff-phillips-18 (approved):
One minor nit but LGTM.

jeff-phillips-18:
The issue is the kebab hit area is so small. Attempting to open the menu often misses and instead expands/collapses. Same is true for row selection. In general the kebabs should have more padding to increase the hit zone.

dlabrecq:
Yes, I may have clicked slightly off the kebab. However, I can click anywhere and expand/collapse the row. It was just a little confusing when I tried clicking the kebab and the row collapsed.

priley86:
@jgiardino nice work.

@dlabrecq @jeff-phillips-18 agreed on the kebabs hit zone increasing.

jgiardino:
I also agree with @dlabrecq and @jeff-phillips-18 about the hit zone for the kebab. I also think the checkbox is a small target area too, and should be fixed along with that.

@LHinson What do you think? And is that something that should require a separate story, or is that considered part of this story?

srambach:
Usability-wise, I find the borders confusing, and also with the hover color and selected color there's a lot going on when I move the mouse around. But the implementation LGTM with the comments above about clicking on the entire area/kebab.

LHinson:
Yeah, we have a bug already filed for this. https://patternfly.atlassian.net/browse/PTNFLY-1344

Let's keep it as a separate issue @jgiardino @dlabrecq @jeff-phillips-18. I'll move it up in priority to address though given we are starting to use these frequently.

andresgalante:
@dlabrecq what browser are you in to see the missing line?
also, +1 to kebab @LHinson lets fix it soon, it's a bug.

andresgalante:
@jgiardino thanks a lot, it looks great and @junezhang Thanks for setting the base for this to happen.
I am ok to merge it but we should consider fixing the kebab soon.

dlabrecq:
I found the border issue with Firefox.

Sorry if the kebab is causing confusion? I want to make it clear that clicking anywhere in the row will cause it to expand/collapse -- not just the toggle button.

PR #510, opened 2016-10-14
PTNFLY-1893 Update table view toolbar to match list view

dlabrecq:
## Description

The original intention of this PR was to update the table view toolbar so that it matches the list view and card view. However, there was a request to make the toolbar work both as an embedded toolbar and with a full-page navbar; thus, there are now two examples.

Note that I had a CSS solution for inline actions that worked across all browsers except IE9. I've removed that in favor of Jenn's solution. Unfortunately, that does not work on IE. For IE, JavaScript has been added to resize the height of inline actions -- this is disabled for browsers other than IE.

Also modified the pfSelect extension (for checkbox selection) to use the selector defined in the dataTable config. This will ensure data tables and the pfSelect extension are in snyc when selecting a row. It would also allow devs to use a radio button instead of a checkbox (e.g., for single row selection).

General cleanup (e.g., renamed files from datatables to tableview).

PTNFLY-1893 Table View: Update with Toolbar 
## Changes

- Added Sort, Search, and "Change View" controls to toolbar
- Modified pfSelect extension to use selectors defined in dataTable config. This allows for greater customization (e.g., a radio button could be used for single row selection, instead of a checkbox).
- General cleanup (e.g., renamed files from datatables to tableview).

https://rawgit.com/dlabrecq/patternfly/table-view-dist/dist/tests/table-view.html
https://rawgit.com/dlabrecq/patternfly/table-view-dist/dist/tests/table-view-navbar.html

This is an image:

screen shot 2016-10-18 at 1 16 13 pm

Table View with a Navbar:

screen shot 2016-10-18 at 1 17 01 pm
## PR checklist (if relevant)


dlabrecq:
@LHinson Please review?

I don't believe it makes sense to have a sort dropdown when sorting is already available via the column headers. Likewise, I don't agree that there should be "change view" controls for list view, card view, etc. The table does not support alternate views like that.

andresgalante:
We are still having issues with the actions btns:
screen shot 2016-10-20 at 1 57 00 pm
screen shot 2016-10-20 at 1 57 06 pm

I wonder if we can fix it by position them absolute and pinning the corners or maybe flexbox?

src/less/datatables.less
```diff
@@ -163,195 +170,66 @@
...
-      .sorting_asc,
-      .sorting_desc,
-      .sorting_asc_disabled,
-      .sorting_desc_disabled {
-        cursor: pointer;
-      }
-      .sorting_asc,
-      .sorting_desc {
-        color: @link-color !important;
+/*
```
andresgalante: Why comment out instead of remove?

dlabrecq:
This is expected. As mentioned in the description, I've intentionally removed the CSS that deals with inline actions. My solution worked for all browsers except ie9. 

Instead, I plan to add Jenn's solution. Although that does not work on IE, I've implemented a JavaScript solution to cover that.

I was going to create a separate PR, but will include Jenn's solution here now that the PR is available.

If you can provide a better solution (e.g., using flex box), please feel free to try.

src/less/datatables.less
dlabrecq: I meant to remove this.

dlabrecq:
@andresgalante I included Jenn's solution for inline actions from PR #520. Note that this CSS solution does not work on IE, so we have JavaScript to dynamically set the height there -- the JS is enabled for IE only.

andresgalante:
Thanks @dlabrecq this looks much better.
I've found that the position of the drop menu on the kebab is wrong:
screen shot 2016-10-20 at 9 24 26 pm

dlabrecq:
@andresgalante I added the dropdown-menu-right class to the kebab because of Jeff's review comment, where the drop down menu can appear off the page. If it's a large screen, no problem; otherwise, the menu can appear out of view.

andresgalante:
I understand but It's out of place, too much to the right. The arrow should land on the middle of the btn and the right border should be align to the table.

dlabrecq:
Ok, I've centered the arrow on the middle of the button. However, the arrow doesn't have enough room to align with the table edge, so it overlaps a bit.

screen shot 2016-10-21 at 9 56 44 am

bleathem:
I've pushed a recent build of this branch to the patternfly repo using the branch name `table-view-dist`:
https://rawgit.com/patternfly/patternfly/table-view-dist/dist/tests/index.html

Looking at the listing test page, I don't like how the `Data Tables with (deprecated) column visibility, reordering` example is listed first and is separate from the newer `Table View` examples.  Perhaps we should rename the `Data Tables` example to `Table View - [Deprecated example] Data Tables with column visibility, reordering`.  Or consider removing the example altogether (we are not removing any supporting code or CSS, but just the example).  I don't consider example pages to be API.

bleathem:
The arrow is still not centered on the button for me:
selection_039

bleathem:
The table icon should be highlighted in the _content view changer_:

selection_040

dlabrecq:
FYI, as I push changes, this is example always updated:
https://rawgit.com/dlabrecq/patternfly/table-view-dist/dist/tests/table-view.html

I've renamed the existing example as you suggested. I believe we should keep this for a while to ensure new changes don't break legacy apps.

srambach:
FWIW, I had the exact same problem with the kebab arrow yesterday, and for me, setting the right of the dropdown to the same as the padding on the kebab seems to reliably fix it. (I'm using 0.5em)

src/less/datatables.less
```diff
@@ -164,194 +172,15 @@
...
 }
 
+// Not in use with latest example: May be used with column reordering?
 .DTCR_pointer {
   background-color: @link-color;
   width: 1px;
   z-index: 201;
 }
-
-.datatable-pf {
```
bleathem, line 178: Is the removal of this class a breaking change?  Should we consider leaving it in place (marked as deprecated) even if we are no longer using it?

dlabrecq: I added that class name for the feature branch. The new class names are prefixed with table-view-pf.

bleathem:
I have a concern with the number of javascript files that exist as "sample code" in the test pages, where the user is expected to copy/paste this example code into their project in order to consume it.  Particularly for UX critical pieces like the button resize logic.

We should consider an extension (or several) that holds patternfly required logic, and ship those extensions for people to consume.  I'm curious to hear your thoughts on the feasibility/workload of such an effort @dlabrecq.

dlabrecq:
Thanks @srambach for the suggestion. Adding padding to the table cell isn't going to help me. In fact, we want the padding to be zero here. Otherwise, the column is too wide.

I suspect the problem is that the table cell width varies slightly based on the screen size. Since the kebab menu is positioned absolutely, I've set the kebab column to a fixed width instead. That should ensure arrow remains centered now.

dlabrecq:
@bleathem Per our discussion, I've moved the JavaScript for the inline action resize to a DataTables extension.

dlabrecq:
@bleathem It took quite a bit of effort, testing, and more testing, but I was able to move my example code into DataTables.net extensions. This includes resizing of inline actions, applying empty table state, and simple filtering. And of course, row selection.

For complex filtering; like ranges, a custom implementation may be necessary. In fact, the Angular toolbar example swaps the filter input field with a drop down when the last option is selected. The extension doesn't cover that scenario, but believe it should cover the most basic cases.

dlabrecq:
@bleathem It took quite a bit of effort, testing, and more testing (e.g., ensuring this worked with multiple tables on the same page), but rewrote my example code as DataTables.net extensions. This includes the resizing of inline actions, applying empty table state, and simple filtering. And of course, row selection.

For complex filtering (e.g., ranges), a custom implementation may be necessary. In fact, the Angular toolbar example swaps the filter input field with a drop down when the last option is selected. The extension doesn't cover that scenario, but believe it should cover the most basic cases.

bleathem:
+1, awesome work @dlabrecq !

dlabrecq:
Merging. Will create a new PR to merge with master.

PR #513, opened 2016-10-17
Allow navigation item labels to use all available space

jeff-phillips-18:
Rather than using a max-width for the item labels, use flex box display to allow the name to take up whatever space is available so as not to truncate the name unless necessary.

This only effects browsers supporting Flex box.

Examples can be see on the following pages, the "Amet" secondary menu has been updated to use really long menu item names to show these changes.

http://rawgit.com/jeff-phillips-18/patternfly/fixes-dist/dist/tests/vertical-navigation-with-badges.html

http://rawgit.com/jeff-phillips-18/patternfly/fixes-dist/dist/tests/vertical-navigation-with-tertiary-pins.html

### DIFF
--- a/src/less/vertical-nav.less
+++ b/src/less/vertical-nav.less
@@ -77,6 +77,7 @@
       position: relative;
       white-space: nowrap;
       width: @nav-pf-vertical-width;
+      // When flexbox is supported nav item names take up all available space
       @supports (display: flex) {
         display: flex;
         padding-right: 0;
@@ -128,6 +129,7 @@
       display: block;
       line-height: 25px;
       max-width: 120px;
+      // If flexbox is supported, do not set max-width, take all space with just some right padding
       @supports (display: flex) {
         flex: 1;
         max-width: none;
@@ -172,6 +174,8 @@
     position: absolute;
     right: 15px;
     top: 20px;
+    // If flexbox is supported, use relative positioning to place to the right of the label
+    // and adjust the top position so that the secondary and tertiary nav items don't need to change
     @supports (display: flex) {
       padding-left: 0;
       padding-right: 15px;
@@ -335,6 +339,7 @@
       top: 0;
     }
     .list-group-item-value {
+      // If flex box is supported add some padding to account for the submenu indicator
       @supports (display: flex) {
         padding-right: 35px;
       }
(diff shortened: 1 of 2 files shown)


jeff-phillips-18:
@andresgalante @srambach

srambach:
I'm not sure what this covers. If I make the heading long, the items are still truncated with room to spare.
screen shot 2016-10-18 at 1 54 38 pm

srambach:
Also, on the pinnable version, is the desired behavior to wrap under the icon or even with the text? (not sure if this is even part of the pr) 
screen shot 2016-10-18 at 1 58 26 pm

src/less/vertical-nav.less
```diff
@@ -163,6 +172,14 @@
...
     right: 15px;
     top: 20px;
+    @supports (display: flex) {
+      padding-left: 0;
+      padding-right: 15px;
+      position: relative;
+      right: 0;
+      margin-top: -3px;
+      top: 5px;
+    }
```
srambach, line 186: Is this conditional on flex support because of what it contains? Sorry, I'm having a hard time following what's going on. Maybe some comments? :-)

jeff-phillips-18:
Odd... where are you running this from? This change also changed the items on the Amet menu and  I am not seeing that change in your shots.

This PR does not cover the submenu heading as you see in the second shot you've included.

srambach:
I pulled down a local copy but was not seeing the Amet menu changes. I altered the content of the menus in the Chrome dev tool.
Edit: My bad - forgot to rebuild.

srambach (commented):
Ok, not covered here.

src/less/vertical-nav.less
jeff-phillips-18: Yes, when we can use flex and do not set a max width, we can use relative positioning rather than absolute. I changed  the top margin so that we don't need to also modify the top position for the secondary and tertiary nav items. Less checks for @supports this way :)

srambach: OK, got it. I'd still like to see comments :)

jeff-phillips-18: Added the comments.

srambach:
LGTM, thanks for the inline comments!

andresgalante:
👍 thanks a lot @jeff-phillips-18 and @srambach for the review

PR #514, opened 2016-10-17
Fix empty page on datatables.

ruhickey:
## Description

In the current version of Patternfly I believe there is a fault with the DataTables plugin.  
Number of rows : `n`
Number of rows per page : `m`
If `n % m == 0` then we can go to an extra blank page.

If we have 10 rows and `pageLength` is set to 5.
We are then able to go to page 3 which is now blank.
The top of the table would say `Showing 11 to 10 of 10 Items`.
## Changes

The fix I propose is to change:
`if (iNewStart > oSettings.fnRecordsDisplay())` to
`if (iNewStart >= Settings.fnRecordsDisplay())` on line `174` of patternfly.js .
## Link to Example

Here is a [JFiddle](https://jsfiddle.net/12651kgk/) which shows the problem.
If you go onto the page and change the page from 1 to 2 in the textbox, a blank table will show up.

### DIFF
--- a/src/js/patternfly.js
+++ b/src/js/patternfly.js
@@ -171,7 +171,7 @@
             }
 
             iNewStart = oSettings._iDisplayLength * (this.value - 1);
-            if (iNewStart > oSettings.fnRecordsDisplay()) {
+            if (iNewStart >= oSettings.fnRecordsDisplay()) {
               /* Display overrun */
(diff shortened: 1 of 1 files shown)


dlabrecq:
LGTM

andresgalante:
Thanks a lot @Dobflem !

PR #519, opened 2016-10-19
Improving wizard files organization.

cardosogabriel:
## Description

Improved how the code for wizard is distributed in the files, optimizing to be consumed by patternfly-org.

Jira issue: https://patternfly.atlassian.net/browse/PTNFLY-1881
## Changes
- Moved the code that is not used in more than one place to pages/wizard.html. They don't need to be in includes.
- Instead of having 3 includes, went back to 1 include, with the code ready to be consumed by patternfly-org. Deleted the unnecessary includes, since code was moved to the page itself.

Rawgit link: http://rawgit.com/cardosogabriel/patternfly/PTNFLY-1881-dist/dist/tests/wizard.html


andresgalante:
yeap, this is much better, thanks!

tests/pages/_includes/widgets/communication/wizard.html
```diff
@@ -1,444 +1,278 @@
...
+                <a>
+                  <span class="wizard-pf-substep-number">3A.</span>
+                  <span class="wizard-pf-substep-title">Summary</span>
                 </a>
               </li>
               <li class="list-group-item">
                 <a>
-                  <span class="wizard-pf-substep-number">1C.</span>
-                  <span class="wizard-pf-substep-title">CDN Channel</span>
+                  <span class="wizard-pf-substep-number">3B.</span>
```
dlabrecq, line 64: The changes look good, but having trouble finding where "1C", "General", and "CDN Channel" are in the existing example -- changes from line 30-64. Perhaps you can provide a before/after screen shot?

cardosogabriel: With the reorganization, they are no longer in this include, but ended up in the wizard page itself, since they are not in the Complete wizard: https://github.com/patternfly/patternfly/blob/master/tests/pages/wizard.html

dlabrecq: Thanks for the explanation. I couldn't find this because we never get to see these steps with the initialization example, only the spinner.

cardosogabriel: I didn't bother so much since it is only an example that doesn't even go to patternfly.org. But feel free to ask Jenn if u think it is unnecessary.

dlabrecq: No, don't bother. We're good.

PR #520, opened 2016-10-20
Updated CSS for inline actions in data tables.

jgiardino:
## Description

Updated the css for buttons that display in data tables. This css works for FireFox and Chrome, but in IE the height of the button is not equal to the height of the table row when text wrapping occurs in the table row. Height of buttons in IE will need to be set dynamically.

https://patternfly.atlassian.net/browse/PTNFLY-1897
## Changes
- modified where the btn-default class is applied (i.e. moved it back to the button).
- included a wrapper class around buttons (this was needed to avoid extra css changes to the styles defined for btn-default that affected width, like box-sizing and padding)

https://rawgit.com/jgiardino/patternfly/table-view-dist/dist/tests/datatables.html

Example of buttons in IE
screen shot 2016-10-14 at 4 08 15 pm
## PR checklist (if relevant)

### DIFF
--- a/src/less/datatables.less
+++ b/src/less/datatables.less
@@ -185,10 +185,12 @@
   }
   // Override table styles -- child selectors are due to specificity
   table {
+    height: 100%;
     margin-bottom: 0;
     max-width: none !important;
     > tbody {
       > tr {
+        height: 100%;
         > td {
           // Styling for inline actions
           &.datatable-pf-actions {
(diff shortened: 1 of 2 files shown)


src/less/datatables.less
```diff
@@ -185,10 +185,12 @@
   }
   // Override table styles -- child selectors are due to specificity
   table {
+    height: 100%;
     margin-bottom: 0;
     max-width: none !important;
     > tbody {
       > tr {
+        height: 100%;
```
dlabrecq, line 193: Thanks for the PR. I included your changes in my latest table and found that the height:100% wasn't necessary on the tr. Do you think this is ok to omit?

jgiardino: Yes, thanks for catching that. The 100% height is necessary for the table though.

jgiardino:
Closing this PR since Dan is including my changes in his PR.

PR #522, opened 2016-10-21
Allow users to select multiple list items or cards with shift key

dabeng:
## Description

With the help of the shift key, users can select multiple cards or list items. For details, please refer to [PTNFLY-1721](https://patternfly.atlassian.net/browse/PTNFLY-1721).
## Link to rawgit

[Card View - Multiple Select](https://rawgit.com/dabeng/patternfly/shift-key-dist/dist/tests/card-view-multi-select.html)

[List View](https://rawgit.com/dabeng/patternfly/shift-key-dist/dist/tests/list-view.html)

### DIFF
--- a/tests/pages/_includes/widgets/layouts/card-view-multi-select.html
+++ b/tests/pages/_includes/widgets/layouts/card-view-multi-select.html
@@ -1,6 +1,6 @@
 {% include widgets/layouts/navbar-primary.html %}
 {% include widgets/layouts/toolbar.html %}
-<div class="container-fluid container-cards-pf">
+<div class="container-fluid container-cards-pf" id="card-container">
   <div class="row row-cards-pf">
     <div class="col-xs-12 col-sm-4 col-md-3 col-lg-2">
 {% include widgets/cards/object-status.html class1="card-pf-view-select card-pf-view-multi-select" %}
@@ -66,7 +66,7 @@
       { $(this).parent().parent().addClass('active'); }
     });
     // allow users to select multiple cards with shift key
-    $('.container-cards-pf').on('click', '.card-pf-view-checkbox>input', function(event) {
+    $('#card-container').on('click', '.card-pf-view-checkbox>input', function(event) {
       var $cardsContainer = $('.container-cards-pf');
       var prevIndex = $cardsContainer.data('prevIndex');
       var $cards = $cardsContainer.find('.card-pf');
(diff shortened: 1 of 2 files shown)


dabeng:
Hi @jeff-phillips-18, my rawgit pages is ready, please try it out.

tests/pages/_includes/widgets/layouts/card-view-multi-select.html
```diff
@@ -65,5 +65,20 @@
       else
       { $(this).parent().parent().addClass('active'); }
     });
+    // allow users to select multiple cards with shift key
+    $('.container-cards-pf').on('click', '.card-pf-view-checkbox>input', function(event) {
```
dlabrecq: Not certain we want to add the shift key to our card examples on pf.org (that's a question for @LHinson ), but I can comment on the JS.

We need to be careful when referencing all '.container-cards-pf' selectors because this particular JS may appear more than once in the page. Note that the pages placed under the tests/pages/_includes directory are not only used in the test pages, but also included on pf.org.

For pf.org, this page will be included once under the overview tab and again for the code tab. See: http://www.patternfly.org/pattern-library/content-views/card-view

If the JS is included more than once, it will attempt to apply an onclick event multiple times. The JS may also interfer with other examples in the page. This is likely not the only test included in the code tab, for example.

It may be best to rewrite this using an ID or reference all '.container-cards-pf' selectors under a particular div tag using an ID? 

This JS may work (if the example is isolated in an iframe), but needs to be tested with pf.org.

bleathem: +1 for writing a more selective query selector.

dlabrecq (approved):
LGTM.

@LHinson Are you ok with allowing users to select multiple list items?

LHinson:
+1 makes sense to me. Looping in @mcarrano

mcarrano:
@LHinson +1 to allowing for multi-select via keyboard as long as the list supports it.  Multi-select should be an optional feature of the list view so keyboard selection should work as specified when checkboxes are present.  Do we need to consider other selection actions like drill-down?  Or is that beyond the scope of this?

dlabrecq:
@LHinson Do you have any further comments to add before merging?

dlabrecq:
@LHinson gave me the go ahead to merge.

PR #523, opened 2016-10-21
Implemented expand-collapse section in forms

jgiardino:
## Description

Included an example of an expand-collapse section in a form.
https://patternfly.atlassian.net/browse/PTNFLY-1683
## Changes
- Implemented the section using fieldset and legend and added new css for these elements
- Included javascript to enable the toggle feature
- Included two examples of the form on the forms.html test page:
  - One form is inline - under header "Expand-Collapse Section"
  - One form is in a modal - under header "Expand-Collapse Section inside a modal"
## Questions

I included an href on the anchor link inside the legend so that it would be keyboard accessible. The alternative would be to remove the href and include tabindex="0" if the behavior of the anchor link is awkward (it's a little weird on the forms.html page, but I'm not sure it would be so weird on a normal form page). Any thoughts?

Here is a [link to Forms on rawgit](https://rawgit.com/jgiardino/patternfly/form-expand-collapse-3-dist/dist/tests/forms.html). Scroll down to the section labeled “Expand-Collapse Section”

This is an image:
screen shot 2016-10-21 at 11 26 26 am
## PR checklist (if relevant)

### DIFF
--- a/tests/pages/_includes/widgets/forms/expand-collapse-section.html
+++ b/tests/pages/_includes/widgets/forms/expand-collapse-section.html
@@ -1,0 +1,80 @@
+<form class="form-horizontal">
+  <div class="form-group">
+    <label class="col-sm-{{include.col-size-label}} control-label" for="{{include.id-blueprintname}}">Blueprint Name</label>
+    <div class="col-sm-{{include.col-size-field}}">
+      <input type="text" id="{{include.id-blueprintname}}" class="form-control">
+    </div>
+  </div>
+  <div class="form-group">
+    <label class="col-sm-{{include.col-size-label}} control-label" for="{{include.id-catalog}}">Catalog</label>
+    <div class="col-sm-{{include.col-size-field}}">
+      <input type="text" id="{{include.id-catalog}}" class="form-control"></div>
+  </div>
+  <div class="form-group">
+    <label class="col-sm-{{include.col-size-label}} control-label" for="{{include.id-number}}">Number</label>
+    <div class="col-sm-{{include.col-size-field}}">
+      <input type="text" id="{{include.id-number}}" class="form-control"></div>
+  </div>
+
+  <fieldset class="fields-section-pf" id="{{include.id-fieldsection}}">
+    <legend class="fields-section-header-pf">
+      <span class="fa fa-angle-right fa-angle-down field-section-toggle-pf"></span>
+      <a href="#{{include.id-fieldsection}}" class="field-section-toggle-pf">Advanced Options</a>
+    </legend>
+    <div class="form-group">
+      <label class="col-sm-{{include.col-size-label}} control-label" for="{{include.id-entry1}}">Entry Point 1</label>
+      <div class="col-sm-{{include.col-size-field}}">
+        <input type="text" id="{{include.id-entry1}}" class="form-control"></div>
+    </div>
+    <div class="form-group">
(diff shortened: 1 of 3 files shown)


catrobson:
Reviewed example and it looks good from a design standpoint for implementation.  Is there design and usage documentation that will also go with this on the website?

LHinson:
@catrobson that can be seen here: http://www.patternfly.org/pattern-library/forms-and-controls/expand-collapse-section/#/api

LHinson:
impl looks great to me too from a design perspective. looping in @andresgalante @kybaker

tests/pages/_includes/widgets/forms/expand-collapse-section.html
```diff
@@ -0,0 +1,80 @@
...
+  <div class="form-group">
+    <label class="col-sm-{{include.col-size-label}} control-label" for="{{include.id-number}}">Number</label>
+    <div class="col-sm-{{include.col-size-field}}">
+      <input type="text" id="{{include.id-number}}" class="form-control"></div>
+  </div>
+
+  <fieldset class="fields-section-pf" id="{{include.id-fieldsection}}">
+    <legend class="fields-section-header-pf">
+      <span class="fa fa-angle-right fa-angle-down field-section-toggle-pf"></span>
+      <a href="#{{include.id-fieldsection}}" class="field-section-toggle-pf">Advanced Options</a>
```
andresgalante, line 22: The `#` on the `href` makes the page scroll to that point creating a jump. I'd leave that empty so it stays in the same place

jgiardino: Thanks for reviewing! I noticed that just having href="#" resulted in the page scrolling to the top. The alternative that I can try is replacing href with tabindex="0" so that it is seen as a tab stop.

andresgalante: Just leave the `href` empty: `href=""` and problem solved 😄  
Good point with the tabindex="0", please add it, it's needed.

jgiardino: I just tried changing the <a> line to this: 
    `<a href="" class="field-section-toggle-pf" tabindex="0">Advanced Options</a>`
This resulted in the page scrolling to the top when I clicked the link.

I also tried removing href:
    `<a class="field-section-toggle-pf" tabindex="0">Advanced Options</a>`
This resulted in the expand/collapse behavior defined for the <a> not being triggered by the Enter key. 

Does anyone have any suggestions? The only options I can think of are:
1. Use javascript to trigger the behavior on pressing the Enter key
2. Keep the code as-is. 
    \* The forms page isn't a realistic example of an actual page that would display a form since there is way more content displaying above and below the form in question. If it helps to evaluate this solution, I can create separate test pages with better form examples.

andresgalante: @jgiardino you are right, this isn't realistic. Thanks for trying.
The CSS looks great. If you want to try someone else you could change the link for a button with the class `btn btn-link`. But then we enter the discussion if this should be a link or a btn.
@bleathem @dlabrecq Can you guys check the JS? if its ok lets merge it.

andresgalante:
@jgiardino thanks a lot! 👍

dlabrecq:
If you're looking to stop the click event for the link, add onclick="return false;". I edited the page directly and it seems to stop the jumping.

bleathem:
Looks good to me.  Merging.

PR #527, opened 2016-11-02
Break up javascript for functions and settings

jeff-phillips-18:
## Description
This PR aims to separate the jQuery dependent functions from the basic patternfly settings in the javascript file. This PR splits the patternfly javascript into two files:
  patternfly-settings.js
  patternfly-functions.js

This allows non-jquery applications to include only the patternfly-settings to get the settings and defaults without the jquery dependency. At build time, the two files get concatenated into patternfly.js. This preserves backwards compatibility. 

I have left the patternfly.js file in the src/js directory to preserve its history but we should consider adding it to .gitignore so that it is not an extra file being changed whenever either the settings or functions file changes. I'm open to making this change as part of this PR if desireable.

## Changes
dist/js now contains:
  patternfly.js
  patternfly.min.js
  patternfly-functions.js
  patternfly-fuctions.min.js
  patternfly-settings.js
  patternfly-settings.min.js

This adds a dependency on the npm package grunt-concat.

@dtaylor113 @dgutride @dlabrecq @bleathem


dtaylor113:
Should 'patternfly-functions.js' be renamed to 'patternfly-jquery-functions.js' to be crystal clear on what the file contains?

dtaylor113 (approved):

jeff-phillips-18:
@dtaylor113 Not sure we need to be _that_ crystal clear. If it comes to a point where we want to have non-jquery dependent functions available to jquery independent apps then we could break it up further.

priley86:
@jeff-phillips-18 why not split out patternfly-functions.js further? I guess I am missing some context here. It seems that downstream there is the case you would not need all functions/plugins. Maybe that case doesn't exist right now though.

dgutride (approved):
Looks good

jeff-phillips-18:
I don't see the need to split up further at the moment. I believe all the functions are currently dependent on jquery

dgutride:
@dtaylor113  - are you good with merging based on Jeff's response re: file names?

dtaylor113:
Done :-)

PR #528, opened 2016-11-03
Remove generated patternfly.js file from /src

jeff-phillips-18:
## Description
Now that the patternfly.js file is generated from patternfly-settings.js and patternfly-functions.js, it is no longer a ‘src’ file and should only be found in the /dist directory.

## Note
Any application that might have used patternfly.js from /src should update to use it from /dist.

@bleathem @dtaylor113 @dgutride


bleathem:
+1 this makes sense to me.  We can consider `/dist` as our repo _API_ and `/src` as our internal implementation.  Hopefully folks are already consuming _patternfly.js_ from `/dist` and the change will go unnoticed.

dtaylor113 (approved):

dgutride (approved):
Looks good

Issue #517, opened 2016-10-19
Menu glitch after refreshing a page with pinned navbar


skateman:
Steps to reproduce:
1. Pin a menu item in the navbar
2. Refresh the page
3. Move your mouse quickly to an expandable 1st level navbar item other than the pinned
4. After the pinning is reset, the hovered 1st level navbar item is still visible

![screencast from 2016-10-19 11-48-42](https://cloud.githubusercontent.com/assets/649130/19515900/bae4c622-95f9-11e6-95e3-034d1dc15397.gif)

andresgalante:
Thanks guys 😄

skateman:
I'm still able to reproduce this issue with v3.13:
![screenshot from 2016-10-31 11-20-16](https://cloud.githubusercontent.com/assets/649130/19851025/19bed190-9f5c-11e6-95b5-70d3d659ebed.png)

jeff-phillips-18:
@skateman Can you try to reproduce on http://rawgit.com/jeff-phillips-18/patternfly/tertiary-nav-dist/dist/tests/vertical-navigation-with-tertiary-pins.html ?

skateman:
@jeff-phillips-18 :+1:

PR #533, opened 2016-11-10
Fix for vertical navigation showing primary item when secondary pinned

jeff-phillips-18:
Fixes issue #517

This only occurs on a reload if the user happens to be hovering on an
area where the primary item would be drawn if nothing was pinned.


jeff-phillips-18:
@srambach @andresgalante Please review

andresgalante:
it works! 👍  

Thanks a lot @jeff-phillips-18

Issue #535, opened 2016-11-18
Broken Webpack compatibility in 3.14.0


lorthirk:
Hi all. I'm using Patternfly (and Angular-Patternfly, for what it's worth) with Webpack. Up until 3.13.0 everything worked fine, but today upgrading to 3.14.0 I discovered that the `patternfly` global variable isn't exported anymore. Specifically, these lines (patternfly-settings.js:497)

```
if (typeof define === 'function' && define.amd) {
  define("patternfly", function () {
    return patternfly;
  });
} else if ('undefined' !== typeof exports && 'undefined' !== typeof module) {
  module.exports = patternfly;
} else {
  window.patternfly = patternfly;
}
```
Gets translated by Webpack to:
```
if (true) {
  !(__WEBPACK_AMD_DEFINE_RESULT__ = function () {
    return patternfly;
  }.call(exports, __webpack_require__, exports, module), __WEBPACK_AMD_DEFINE_RESULT__ !== undefined && (module.exports = __WEBPACK_AMD_DEFINE_RESULT__));
} else if ('undefined' !== typeof exports && 'undefined' !== typeof module) {
  module.exports = patternfly;
} else {
  window.patternfly = patternfly;
}
```
So anyone who needs the patternfly global variable will fail (e.g.  patternfly-functions.js:223):
```
(function ($) {
  'use strict';
  if (patternfly !== undefined) {
    $.pfBreakpoints = patternfly.pfBreakpoints;  // Fails here
  }
}(jQuery));
```
On a side note I tried to fix this on the Webpack side using [expose-loader](https://github.com/webpack/expose-loader) but being a newbie with Webpack I can't get it to find the patternfly module 😒

lorthirk:
This can be closed, right?

dgutride:
Yes - closing :)

PR #537, opened 2016-11-22
Move patternfly back to window to support webpack

dgutride:
## Description
This reverts a change earlier to move PatternFly to a module loading system.  It is now just on window (the way it was before 3.14.0.

https://github.com/patternfly/patternfly/issues/535
## Changes
Simplify Patternfly window object until a better module system is chosen.

## PR checklist (if relevant)
- Checked in webpack
- Checked in regular dist (test pages) to make sure script still loads correctly

cc: @lorthirk, @jeff-phillips-18

### DIFF
--- a/src/js/patternfly-settings.js
+++ b/src/js/patternfly-settings.js
@@ -492,15 +492,6 @@
     'desktop': 1200
   };
 
-  if (typeof define === 'function' && define.amd) {
-    define("patternfly", function () {
-      return patternfly;
-    });
-  } else if ('undefined' !== typeof exports && 'undefined' !== typeof module) {
-    module.exports = patternfly;
-  } else {
-    window.patternfly = patternfly;
-  }
-
+  window.patternfly = patternfly;
 })(window);
 


lorthirk:
Yes, that should do it.

As said I would also explore a better integration of the "module aware" version with Webpack, but that would need someone with a better understanding of its internals than me.

bleathem:
Pasting a conversation from slack:

---
bleathem	[9:52 AM]  
@dana @jphillips: can you explain https://github.com/patternfly/patternfly/pull/537/ for me?

[9:52]  
it looks like the amd stuff was added in https://github.com/patternfly/patternfly/pull/527

[9:52]  
is that correct?

[9:53]  
from 527 I'm not seeing how the patternfly global was set initially

dana	[9:58 AM]  
yeah, webpack falls apart with the change because it tries to use patternfly as a module and the other code that is separated out can’t find patternfly

[9:59]  
hang on - I’ll get you the actual line that fails with it

[9:59]  
@bleathem  - anything that looks like this in patternfly-functions.js breaks now:

[9:59]  
(function ($) {
 'use strict';
 if (patternfly !== undefined) {   // this line breaks because patternfly is not defined (which is different than undefined)
   $.pfBreakpoints = patternfly.pfBreakpoints;
 }
}(jQuery)); (edited)

[10:00]  
patternfly can’t be checked that way

bleathem	[10:04 AM]  
k, thanks

dana	[10:05 AM]  
personally - I think the actual fix is to move to es6 modules if we want to support them and then use babel instead of trying to do it this way

bleathem	[10:05 AM]  
mixing modular and non-modular approaches is going to get messy.  I appreciate the solution of simply using the global for now.

dana	[10:06 AM]  
ditto - I think we just said the same thing

bleathem	[10:06 AM]  
lol, yeah, I'd already typed it out so just hit enter :slightly_smiling_face:

[10:06]  
this doesn't break anything for a-pf?

new messages
dana	[10:07 AM]  
no - I tested in my apf solution, too

bleathem:
This approach sounds reasonable to me.  +1 to merge.

PR #541, opened 2016-11-30
PTNFLY-1344: make kebab button wider

michpetrov:
## Description
Widened the kebab button to make it easier to click
JIRA: [PTNFLY-1344](https://patternfly.atlassian.net/browse/PTNFLY-1344)

## Links to rawgit

* [Main page](https://rawgit.com/michpetrov/patternfly/ptnfly1344-dist/dist/tests/)
* [Dropdowns](https://rawgit.com/michpetrov/patternfly/ptnfly1344-dist/dist/tests/dropdowns.html)
* [Toolbar](https://rawgit.com/michpetrov/patternfly/ptnfly1344-dist/dist/tests/toolbar.html)

### DIFF
--- a/src/less/card-view.less
+++ b/src/less/card-view.less
@@ -6,7 +6,6 @@
   border: 2px solid transparent;
   .card-pf-heading-kebab {
     .dropdown-kebab-pf {
-      margin-left: 10px;
       margin-top: -3px;
     }
     + .progress-pf-legend {
--- a/src/less/dropdowns.less
+++ b/src/less/dropdowns.less
@@ -169,7 +169,9 @@
     color: @gray-darker;
     font-size: (@font-size-base + 4);
     line-height: 1;
-    padding: 4px 0;
+    padding: 4px (@grid-gutter-width/4);
+    margin-left: (@grid-gutter-width/(-4));
+    margin-right: (@grid-gutter-width/(-4));
     &:active,
     &:focus,
     &:hover {
--- a/src/less/toolbar.less
+++ b/src/less/toolbar.less
@@ -37,6 +37,11 @@
       &:focus,
       &:hover { color: @link-color; }
     }
+    .dropdown-kebab-pf .btn-link {
+        padding: 4px (@grid-gutter-width/4);
+        margin-left: (@grid-gutter-width/(-4));
+        margin-right: (@grid-gutter-width/(-4));
+    }
   }
 }
 .toolbar-pf-actions {
--- a/src/less/list-view.less
+++ b/src/less/list-view.less
@@ -79,7 +79,8 @@
   margin-top: (@grid-gutter-width/2);
   order: 2;
   button,
-  > a {
+  > a,
+  .dropdown-kebab-pf {
     margin-left: (@grid-gutter-width/4);
   }
   .list-view-pf-top-align & {


LHinson:
Nice, thanks @michpetrov! This has been needed for awhile. Unfortunately don't have the chops to review it but wanted to make sure to reach out and thank you for this contribution. 

@bleathem @jgiardino @srambach will you help review this?

jgiardino:
@michpetrov Thanks for posting this PR! The kebab definitely needs a larger click area. I reached out to our visual design lead @kybaker for feedback regarding spacing of elements. I'd like to wait to review the code after he provides some guidance on that.

jgiardino:
Hi @michpetrov,

 I chatted with @kybaker and @andresgalante about this and their main requirement is that the layout and spacing remains the same as it is today in the examples listed here: https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/

When I review your changes, I see that the spacing around the kebab is not identical to the current implementation where it appears in these examples: 

- toolbar
- toast notification
- notification drawer
- list view 

Another solution that can achieve the same goal is to use padding and negative margin instead of increasing the width. A solution like this might be better at preserving the same layout we have today while still increasing the clickable area. I think this would also reduce the need for additional css.

For example, in dropdowns.less, we can update the padding and margin values as shown here:

 		// Kebab dropmenu
		 .dropdown-kebab-pf {
		   .btn-link {
		     color: @gray-darker;
			     font-size: (@font-size-base + 4);
			     line-height: 1;
			     padding: 4px (@grid-gutter-width/4);
			     margin-left: (@grid-gutter-width/(-4));
			     margin-right: (@grid-gutter-width/(-4));


After updating this ruleset, I reviewed the components and pages where the kebab displays. The only case where I see that additional css is needed is the list view file, list-view.less.

In this file in the following ruleset, the margin-left is getting applied to the button descendent of the dropdown, instead of the parent .dropdown div.  Since the left margin for the kebab button is now getting overwritten by the ruleset above, then we can add .dropdown to the list of selectors where the left margin is getting applied, as shown in the following example:

	.list-view-pf-actions {
	  float: right; // IE9 fallback
	  margin-bottom: (@grid-gutter-width/2);
	  margin-left: (@grid-gutter-width/2);
	  margin-top: (@grid-gutter-width/2);
	  order: 2;
	  button,
	  > a,
	  .dropdown {
	    margin-left: (@grid-gutter-width/4);
	  }

@michpetrov and @srambach - Please let me know what you think about this solution and if I missed anything or if you have an alternate solution to suggest.

src/less/card-view.less
```diff
@@ -6,7 +6,7 @@
   border: 2px solid transparent;
   .card-pf-heading-kebab {
     .dropdown-kebab-pf {
-      margin-left: 10px;
+      margin-right: -10px;
```
jgiardino: If we resolve this issue using padding and negative margin instead, then these updates should not be necessary.

src/less/dropdowns.less
```diff
@@ -223,4 +224,17 @@
...
+      &:before {
+        left: auto;
+        right: 17px;
+      }
+    }
+    &:after,
+    &:before {
+      left: 17px;
+    }
+  }
```
jgiardino: If we resolve this issue using padding and negative margin instead, then these updates should not be necessary.

src/less/toolbar.less
```diff
@@ -25,8 +25,8 @@
     .btn-group {
       + .btn,
       + .btn-group { margin-left: 5px; }
-      + .btn-link,
-      + .dropdown { margin-left: 10px; }
+      + .btn-link { margin-left: 10px; }
+      + .dropdown:last-child { margin-right: -10px; }
```
jgiardino: If we resolve this issue using padding and negative margin instead, then these updates should not be necessary.

michpetrov:
Hi @jgiardino,

the global negative margin seems to work better, though the issue still remains with toolbars - the clickable area doesn't increase, so the `.toolbar-pf .form-group` needs to be updated as well

    .dropdown-kebab-pf .btn-link {
        padding: 4px (@grid-gutter-width/4);
        margin-left: (@grid-gutter-width/(-4));
        margin-right: (@grid-gutter-width/(-4));
    }

(To be honest I don't like this solution but a rule with more selectors overrides the one with less)

Regarding the layout, now that the button is wider it ends up misaligned (even though its borders are not visible), is that not a concern? 
screenshot from 2016-12-07 12-05-42

src/less/card-view.less
michpetrov: After the change it looks like this particular line can be safely removed.

jgiardino:
You're right, thanks for catching that in the toolbar! I completely missed that one. I don't have a better solution than what you suggested.

You raise an interesting point about the layout of the kebab. Based on discussions I've had with the team in the past, I think the general consensus is that the positioning of the layout should be based on how much visual space the kebab occupies, and not the actual space (e.g. clickable space) that it occupies. If that's true, then the example you show is correct because the kebab is visually left-aligned with the other contents.

But I haven't had a chance to discuss this with other team members (and unfortunately @andresgalante is out on vacation). @srambach do you have any thoughts on this?

bleathem:
Thanks for the contribution @michpetrov! And thanks for a very comprehensive review, thanks @jgiardino!  This PR review is in great hands with the patternfly css army :-)

srambach:
This is a great improvement, @michpetrov! I agree that we want good visual alignment along with the increased clickable area evenly distributed around the icon, and this solution does that well. If everything looks ok after those few lines are removed as @jgiardino notes, then I think it's good to go.

michpetrov:
I've implemented the changes and rebased the commit.

srambach:
LGTM

jgiardino:
Thanks for making these updates! They look good to me, too.

PR #547, opened 2016-12-02
Automate the Patternfly releases process via Travis

dlabrecq:
Automate the Patternfly releases process via Travis. This uses a custom tag, which automatically triggers Travis to build, publish, and release. If the Patternfly release is successful, RCUE and Angular Patternfly are built next. If Angular is successful, Patternfly Org is released s well.

Instead of spending between half and a full day releasing Patternfly, Angular PAtternfly, RCUE, and Patternfly Org; the automation manages to get this effort down to about 15-30 minutes.

Should the build fail at any point, it can be fixed and restarted. For example, If Angular Patternfly fails, it can be restarted and Patternfly Org will be released as well. We don't necessarily need to bump the npm version number again or rebuild Patternfy. The npm publish is one of the last steps in the build.

Of course, we still have the ability to run the release manually with the existing scripts I created earlier. The new automation uses these scripts itself.

Spoke to Brian about creating a separate repo for the build and GitHub publish scripts. That way, there is one version of the scripts for all repos. Although, combining scripts would be a next step as each repo has subtle differences in the build.


scripts/release/README.md
```diff
@@ -0,0 +1,39 @@
...
+4. PR reviewed & merged (no dist files should be checked in)
+5. NPM Publish
+ - Create an NPM account and become a collaborator for https://www.npmjs.com/package/patternfly
+ - Run sh ./scripts/release/release.sh -p
+6. Release Notes Published (via GitHub)
+ - Tag “master-dist” branch commit with updated version
+7. Community email sent (patternfly@redhat.com)
+
+### release-all.sh
+
```
dtaylor113, line 30: Great README!  Can you please add some comments on the difference between `release.sh` and `release.all.sh`, and when you would use one over the other -or one before another?

scripts/release/README.md
```diff
@@ -0,0 +1,39 @@
...
+If the release is successful, RCUE, Angular Patternfly, and Patternfly Org will be released as well.
+
+Note: Builds can only be stopped via the Travis CI.
+
+1. Choose version using [semantic versioning](https://docs.npmjs.com/getting-started/semantic-versioning) ([details](https://github.com/patternfly/patternfly/blob/master/README.md#release))
+2. Run sh ./scripts/release/release-all.sh -v <version>
+3. Release Notes Published (via GitHub)
+ - Tag “master-dist” branch commit with updated version
+4. Run sh ./scripts/release/notify.sh -v <version>
```
dtaylor113, line 43: Hi, I noticed there is a `publish.sh`, should that be documented in this README, or is it executed by one of the release scripts? -thanks

dtaylor113 (approved):

scripts/release/README.md
dlabrecq: The release.sh script is what we have today. It creates a branch that must be merged with a PR. Although it's used by the release-all.sh script, there is no reason why someone couldn't use either approach. Other than it's more of a manual process.

The new release-all.sh script is used to chain releases together. When Patternfly is complete, Angular Patternfly and RCUE are kicked off. When Angular Patternfly is complete, Patternfly Org will begin.

scripts/release/README.md
dlabrecq: It's executed only by the release scripts. There are only a couple scripts that we would normally interact with; release-all.sh, and notify.sh.

scripts/release/README.md
dlabrecq: Updated readme to clarify.

PR #548, opened 2016-12-04
Automate the Patternfly releases process via Travis and patternfly-re

dlabrecq:
Automate the Patternfly releases process via Travis and patternfly-re.

Note: This was formally PR #547. The scripts have simply been moved to patternfly-re.

### New patternfly-re repo

Instead of copying the same scripts (with subtle changes) to each repo, a new patternfly-re repo has been created. This will store all release engineering scripts which can be used across all repos. The scripts can be maintained in one place and included via npm.

See: https://github.com/patternfly/patternfly-re

### Normal Builds

For typical builds, the build/build.sh script of patternfly-re is used. This script determines if we're building a pull request, simple tag, or merge. If the build/commit is tagged for release, the build/release/build.sh script takes over.

### Automated Release Build

The automated release build begins with the build/release/release-all.sh script of patternfly-re. This script will push a custom release tag to the repo's master branch. 

This triggers Travis to run the build/release/build.sh script, which will bump version numbers, build, shrinkwrap, test, install, npm publish, etc. Version bump changes are pushed back to the master branch. The version bump and generated build changes are pushed to master-dist and tagged (e.g., v3.15.0).

Release builds are chained together by pushing a new custom release tag to the next repo. For example, if the Patternfly RE release is successful, Patternfly is built next. If the Patternfly release is successful, RCUE and Angular Patternfly are built simultaneously. If Angular is successful, Patternfly Org is released as well.

Should a release build fail at any point, it can be fixed and restarted. For example, If Angular Patternfly fails, it can be restarted and the Patternfly Org release will follow. We don't necessarily need to bump the npm version number again or rebuild Patternfy. The npm publish is one of the last steps in the build.

Of course, we still have the ability to run the release manually using the build/release/release.sh  script I created earlier. In fact, the release build uses this script itself.


dtaylor113 (approved):

package.json
```diff
@@ -36,6 +36,7 @@
     "karma-phantomjs-launcher": "^1.0.1",
     "matchdep": "~1.0.1",
     "nsp": "^2.6.1",
+    "patternfly-re": "git+https://github.com/patternfly/patternfly-re.git",
```
bleathem: I would rather see us do an npm release of patternfly-re, then projects can depend on a specific version.  This way when a breaking change is introduced to patternfly-re, we don't have to worry about breaking the build of consuming repos.  Instead they can be updated as required.

dlabrecq: Sure, we can do that. My worry was that if we need a quick fix to the release scripts, we would need to version bump patternfly-re and update each repo. Or, wait until the next release to get the change.

I actually have the scripts set up version bump and tag patternfly-re, along with the other repos. My plan was to keep the version numbers the same for all.

dlabrecq:
@bleathem I changed each repo to use the npm release of patternfly-re. The scripts are already set to version bump, tag, and npm publish patternfly-re.

bleathem:
I like both the automation around the release process, as well as centrally locating the scripts in a shared artifact.

:+1: this is good to merge.

dlabrecq:
@bleathem I renamed patternfly-re as patternfly-eng-release per our discussion. The publish-githubpages script has been removed -- this will be placed in its own repo. And the npm verision of patternfly-eng-release will continue to match the other repos as before.

The angular, patternfly-org, and rcue repos have alsobeen updated to reflect this change.

PR #552, opened 2016-12-09
Updated readme and quick start to reflect dependency install locations

dlabrecq:
## Description
Updated readme and quick start to reflect dependency install locations when npm 3 and shrink wrap are used.

The npm install of patternfly 3.16.0 includes a shrinkwrap file; thus, npm 3 installs dependencies are installed under the node_modules/patternfly/node_modules directory. Previous versions v3.0.0 - v3.9.0 also included this shrinkwrap file and those dependencies were installed under the same directory as well. For some reason, the npm install for versions v3.10.0 - v3.15.0 omitted the shrinkwrap file (even though it was properly committed), so dependencies for those versions inadvertently install at the root node_modules directory.

If users prefer a flat dependency structure, they can define their own dependencies explicitly. That will flatten out the node_modules structure and place dependencies in the root node_modules directory.

## Changes

Updated readme and quick start to reflect proper dependency install locations.

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -53,54 +53,54 @@ We provide compiled CSS and JS (`patternfly.*`), as well as compiled and minifie
 4. Add the following script includes to your HTML file(s), adjusting where necessary to pull in only what you need:
 
         <!-- jQuery -->
-        <script src="node_modules/jquery/dist/jquery.min.js"></script>
+        <script src="node_modules/patternfly/node_modules/jquery/dist/jquery.min.js"></script>
 
         <!-- Bootstrap JS -->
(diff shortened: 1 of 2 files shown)


dlabrecq:
@bleathem I'd like to merge this so the docs reflect the proper directory structure.

dlabrecq:
Last chance?

dgutride (approved):

andresgalante:
@priley86 @bleathem this is a simple change, it should be easy to review.

jeff-phillips-18 (approved):
👍

priley86:
looks great to me @dlabrecq. thanks for tackling this and flattening our dependency tree.

bleathem (commented):
Looks good, thanks Dan!

PR #562, opened 2017-01-04
Table view pagination

jgiardino:
## Description
Implemented CSS for the pagination row under the data table.
[PTNFLY-2074 Content Views: Pagination - CSS](https://patternfly.atlassian.net/browse/PTNFLY-2074)

## Changes

* Included new css in the file pagination.less
* Included a new include file pagination.html

## Link to rawgit

[Table View with Navbar](https://rawgit.com/jgiardino/patternfly/TableViewPagination-dist/dist/tests/table-view-navbar.html)
[Table View](https://rawgit.com/jgiardino/patternfly/TableViewPagination-dist/dist/tests/table-view.html)

## PR checklist (if relevant)

### DIFF
--- a/src/less/pagination.less
+++ b/src/less/pagination.less
@@ -65,20 +65,27 @@
 
 .content-view-pf-pagination {
   background-color: @table-bg-accent;
+  border: 1px solid @table-border-color;
   border-top: none;
-  border-left: 1px solid @table-border-color;
-  border-right: 1px solid @table-border-color;
-  border-bottom: 1px solid @table-border-color;
   display: flex;
   flex-wrap: wrap;
   justify-content: space-between;
-  padding: 5px;
   .form-group {
     align-items: baseline;
     display: flex;
     flex-grow: 1;
+    float: left; //IE9 fallback
+    @supports (display: flex) {
+      float: none;
+    }
     margin: 5px;
-    &:last-child { justify-content: flex-end; } /* if pagination controls wrap, pagination buttons stay on the right */
+    &:last-child {
+      justify-content: flex-end; // if pagination controls wrap, pagination buttons stay on the right
+      float: right; //IE9 fallback
+      @supports (display: flex) {
+        float: none;
+      }
+    }
     .pagination-pf-pagesize.bootstrap-select.btn-group {
       display: flex;
       margin-right: 5px;
@@ -96,8 +103,10 @@
     display: flex;
     margin: 0 0 0 10px;
     a {
-      display: block;
       float: none;
+      @supports (display: flex) {
+        display: block;
+      }
     }
   }
 }
--- a/src/less/pagination.less
+++ b/src/less/pagination.less
@@ -79,11 +79,11 @@
     flex-grow: 1;
     margin: 5px;
     &:last-child { justify-content: flex-end; } /* if pagination controls wrap, pagination buttons stay on the right */
-  }
-  .form-group .pagination-pf-pagesize.bootstrap-select.btn-group {
-    display: flex;
-    margin-right: 5px;
-    width: auto;
+    .pagination-pf-pagesize.bootstrap-select.btn-group {
+      display: flex;
+      margin-right: 5px;
+      width: auto;
+    }
   }
   .pagination-pf-page {
     margin-left: 10px;
--- a/src/less/pagination.less
+++ b/src/less/pagination.less
@@ -62,3 +62,42 @@
     }
   }
 }
+
+.content-view-pf-pagination {
+  background-color: @table-bg-accent;
+  border-top: none;
+  border-left: 1px solid @table-border-color;
+  border-right: 1px solid @table-border-color;
+  border-bottom: 1px solid @table-border-color;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+  padding: 5px;
+  .form-group {
+    align-items: baseline;
+    display: flex;
+    flex-grow: 1;
+    margin: 5px;
+    &:last-child { justify-content: flex-end; } /* if pagination controls wrap, pagination buttons stay on the right */
+  }
+  .form-group .pagination-pf-pagesize.bootstrap-select.btn-group {
+    display: flex;
+    margin-right: 5px;
+    width: auto;
+  }
+  .pagination-pf-page {
+    margin-left: 10px;
+    margin-right: 5px;
+    padding: 0 2px 2px;
+    text-align: right;
+    width: 2.5em;
+  }
+  .pagination {
+    display: flex;
+    margin: 0 0 0 10px;
+    a {
+      display: block;
+      float: none;
+    }
+  }
+}
--- a/tests/pages/_includes/widgets/table-view/tmpl/pagination.html
+++ b/tests/pages/_includes/widgets/table-view/tmpl/pagination.html
@@ -1,0 +1,31 @@
+<form class="content-view-pf-pagination">
+  <div class="form-group">
+    <select class="selectpicker pagination-pf-pagesize">
+      <option value="6">6</option>
+      <option value="10">10</option>
+      <option value="15" selected="selected">15</option>
+      <option value="25">25</option>
+      <option value="50">50</option>
+    </select>
+    <span>per page</span>
+  </div>
+  <div class="form-group">
+    <span>1-15 of 75</span>
+    <ul class="pagination">
+      <li class="disabled"><a href="#"><span class="i fa fa-angle-double-left"></span></a></li>
+      <li class="disabled"><a href="#"><span class="i fa fa-angle-left"></span></a></li>
+    </ul>
+    <input class="pagination-pf-page" type="text" value="1" />
+    <span>of 5</span>
+    <ul class="pagination">
+      <li><a href="#"><span class="i fa fa-angle-right"></span></a></li>
+      <li><a href="#"><span class="i fa fa-angle-double-right"></span></a></li>
+    </div>
+  </div>
+</form>
+
+<script src="/components/bootstrap-select/dist/js/bootstrap-select.min.js"></script>
+<script>
+  // Initialize Boostrap-select
+   $('.selectpicker').selectpicker();
+</script>
--- a/tests/pages/_includes/widgets/table-view/tmpl/pagination.html
+++ b/tests/pages/_includes/widgets/table-view/tmpl/pagination.html
@@ -20,7 +20,7 @@
     <ul class="pagination">
       <li><a href="#"><span class="i fa fa-angle-right"></span></a></li>
       <li><a href="#"><span class="i fa fa-angle-double-right"></span></a></li>
-    </div>
+    </ul>
   </div>
 </form>
 
--- a/tests/pages/_includes/widgets/table-view/tmpl/pagination.html
+++ b/tests/pages/_includes/widgets/table-view/tmpl/pagination.html
@@ -7,16 +7,16 @@
       <option value="25">25</option>
       <option value="50">50</option>
     </select>
-    per page
+    <span>per page</span>
   </div>
   <div class="form-group">
-    <span class="pagination-pf-items-current">1-15</span> of <span class="pagination-pf-items-total">75</span>
+    <span><span class="pagination-pf-items-current">1-15</span> of <span class="pagination-pf-items-total">75</span></span>
     <ul class="pagination pagination-pf-back">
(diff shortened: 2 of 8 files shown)


src/less/pagination.less
```diff
@@ -62,3 +62,42 @@
...
+      margin-right: 5px;
+      width: auto;
+    }
+  }
+  .pagination-pf-page {
+    margin-left: 10px;
+    margin-right: 5px;
+    padding: 0 2px 2px;
+    text-align: right;
+    width: 2.5em;
```
jgiardino, line 100: I wasn't sure what to do with this. The design has the text input width pretty small, but the width should be large enough to accommodate the width of the last page (e.g. "215" or "1008"). Should the width of this be dynamically set based on the width of the last page, or should it be arbitrarily hard-coded like it is here?

tests/pages/_includes/widgets/table-view/tmpl/pagination.html
```diff
@@ -0,0 +1,31 @@
...
+  </div>
+  <div class="form-group">
+    <span>1-15 of 75</span>
+    <ul class="pagination">
+      <li class="disabled"><a href="#"><span class="i fa fa-angle-double-left"></span></a></li>
+      <li class="disabled"><a href="#"><span class="i fa fa-angle-left"></span></a></li>
+    </ul>
+    <input class="pagination-pf-page" type="text" value="1" />
+    <span>of 5</span>
+    <ul class="pagination">
```
jgiardino: These `<ul>`s will probably need an extra class to indicate "previous" versus "next" so that we can target the correct set for disabling/enabling based on the current page.

src/less/pagination.less
```diff
@@ -62,3 +62,42 @@
...
+
+.content-view-pf-pagination {
+  background-color: @table-bg-accent;
+  border-top: none;
+  border-left: 1px solid @table-border-color;
+  border-right: 1px solid @table-border-color;
+  border-bottom: 1px solid @table-border-color;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
```
andresgalante, line 72: PatternFly 3 supports IE9, please provide a float fallback if you are using flexbox.

src/less/pagination.less
```diff
@@ -62,3 +62,42 @@
     }
   }
 }
+
+.content-view-pf-pagination {
+  background-color: @table-bg-accent;
+  border-top: none;
+  border-left: 1px solid @table-border-color;
+  border-right: 1px solid @table-border-color;
+  border-bottom: 1px solid @table-border-color;
```
andresgalante: maybe use a margin shorthand and then remove it from the top?

src/less/pagination.less
andresgalante: Good question, I'd hardcoded.

src/less/pagination.less
```diff
@@ -62,3 +62,42 @@
...
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+  padding: 5px;
```
andresgalante: I'd remove this 5px, the padding on the table cell is enough IMO.

andresgalante:
@jgiardino This looks fantastic, thanks a lot! I left a few comments.

Remember that pf3 supports IE9, so if you want to use flexbox please provide a float fallback. There is an example of it on listview.

Also, to test it on IE you can use browserstack, if you have an account let me know and I'll send you an invite.

tests/pages/_includes/widgets/table-view/tmpl/pagination.html
```diff
@@ -0,0 +1,31 @@
+<form class="content-view-pf-pagination">
+  <div class="form-group">
```
andresgalante: This doesn't show:
screen shot 2017-01-05 at 10 21 12 am

andresgalante:
Maybe @jeff-phillips-18 or @bleathem can help with the bootstrap-select.min.js issue and raw git

jeff-phillips-18:
@jgiardino  Add '//cdnjs.cloudflare.com/ajax/libs/bootstrap-select/1.10.0/js/bootstrap-select.min.js' to the url-js-extra list in both table-vew.html and table-view-navbar.html and you can remove it from the pagination.html file.

jeff-phillips-18:
@jgiardino  Actually, looking again those table view files are includes, you need to add the change to url-js-extra to all the files that include either of them.

tests/pages/_includes/widgets/table-view/tmpl/pagination.html
dlabrecq: Yes, having an extra class to indicate previous, next, etc. would help identify the controls for JS. When the JS plugin is created, we will need to find specific elements to attach pagination events to. This would be much easier if we had a specific class identifying each control. Otherwise, I'd have to search for 'fa-angle-double-left' or make an assumption that the first, second anchors are always previous, next, etc. which may not always be the case.

src/less/pagination.less
```diff
@@ -62,3 +62,42 @@
     }
   }
 }
+
+.content-view-pf-pagination {
+  background-color: @table-bg-accent;
```
dlabrecq, line 67: I assume pagination will be used in many places other than table view, but will the generic pagination controls always have a background-color of @table-bg-accent? I'm thinking we should also have a table-view-pf-pagination class to define styles specific to the table?

I'm asking mainly because the JS plugin will need a specific selector to identify its pagination controls. The idea is to simply add a JS (plugin) file to the page and have table functionality work automatically, without providing extra setup, defining IDs, etc.

Something other than (or in addition to) content-view-pf-pagination. For example, something like table-view-pagination would help the JS differentiate the table's pagination controls from other controls that may appear in the same page.

You can see an examples with the table toolbar and empty slate. I couldn't wrap the entire table with a class because of the full page nav bar example.

dlabrecq (changes requested):
Looking good. I’m glad you didn’t try top override the DataTable pagination controls. This will actually be easier to implement.

The only change/addition I'd like to see is specific classes to help the JS plugin identify table view controls.

src/less/pagination.less
jgiardino: ok, I left this as-is

tests/pages/_includes/widgets/table-view/tmpl/pagination.html
jgiardino: Except for your last comment, I think I included all the classes you would need.

src/less/pagination.less
jgiardino: I included css for IE9, which now looks like the example below. Please let me know if you have any suggestions about the updates I made for IE.
screen shot 2017-01-05 at 9 58 15 am

src/less/pagination.less
jgiardino: I updated the border styles, assuming you were referring to border instead of margin :-)

src/less/pagination.less
jgiardino: works for me

src/less/pagination.less
jgiardino: I didn't see this until after I pushed my latest updates, but here are a couple of comments/questions until I get around to it...

From the examples I was looking at, it did look like pagination rows for all content views would have the same background color.  

Is it possible that there could be two tables on the same page (like in a dashboard or some two panel view)? I'm wondering if class is the best way to target the pagination controls for a specific table, or if there should be some other more specific identifier instead? Although I get your point about not requiring extra setup.

jeff-phillips-18:
Looks like basic.html and tab.html also include table-view.html and likely need to add bootstrap-select.min.js to the url-js-extra setting.

jgiardino:
Good catch, @jeff-phillips-18! I updated those files and also included the class that @dlabrecq suggested.

jeff-phillips-18 (approved):
👍

jgiardino:
One last comment for @dlabrecq... The pagination controls are still visible when the empty state displays. I'm assuming this would be included as part of the js updates you will be working on.

src/less/pagination.less
dlabrecq: Yes, there can be multiple tables in the same page. Absolutely!

There may also be custom content in this area? Not only would this help the JS plugin, but it would allow the table to be more easily customized by app developers.

If you don't want to add this now, that's fine. However, I suspect I'll create something when implementing the JS plugin. I need to avoid custom IDs with the JS plugin.

dlabrecq:
I'll need to modify the JS plugin to hide the pagination controls for the empty state.

tests/pages/_includes/widgets/table-view/tmpl/pagination.html
dlabrecq: I can search for "fa-angle-double-left", "fa-angle-left", etc. Just need something unique to grab.

dlabrecq: Although, having specific classes might make it easier for developers to customize the buttons? For the JS plugin to work, app developers must (and always) provide "fa-angle-double-left", "fa-angle-left", etc.

tests/pages/_includes/widgets/table-view/tmpl/pagination.html
```diff
@@ -0,0 +1,25 @@
...
+    per page
+  </div>
+  <div class="form-group">
+    <span class="pagination-pf-items-current">1-15</span> of <span class="pagination-pf-items-total">75</span>
+    <ul class="pagination pagination-pf-back">
+      <li class="disabled"><a href="#"><span class="i fa fa-angle-double-left"></span></a></li>
+      <li class="disabled"><a href="#"><span class="i fa fa-angle-left"></span></a></li>
+    </ul>
+    <input class="pagination-pf-page" type="text" value="1" />
+    of <span class="pagination-pf-pages">5</span>
```
andresgalante: There is no space between the "5" and the "of" and also with the "75"
screen shot 2017-01-05 at 2 29 33 pm

tests/pages/_includes/widgets/table-view/tmpl/pagination.html
```diff
@@ -0,0 +1,25 @@
+<form class="content-view-pf-pagination table-view-pf-pagination clearfix">
+  <div class="form-group">
+    <select class="selectpicker pagination-pf-pagesize">
```
andresgalante, line 3: The shadow is too strong, isn't it?

screen shot 2017-01-05 at 2 29 40 pm

tests/pages/_includes/widgets/table-view/tmpl/pagination.html
jgiardino: Oh, thanks for seeing that! Pulling out the wrapping `<span>` tags was a last minute change and I didn't even notice.

tests/pages/_includes/widgets/table-view/tmpl/pagination.html
jgiardino: Do we apply a different shadow based on background colors? This is the shadow that's applied to .btn, but probably looks stronger due to the darker background color. Should I just remove it from this case, or lighten the opacity? Related to this, the pagination buttons have no shadow. Is that what you'd expect?

andresgalante: To be honest I don't know. Maybe the bootstrap select has more shadow than the dropdown. 
@kybaker Can you please take a look at this and decide if we should remove or do something with the shadow here?

tests/pages/_includes/widgets/table-view/tmpl/pagination.html
andresgalante: no problem, thanks for solving it 😄

tests/pages/_includes/widgets/table-view/tmpl/pagination.html
kybaker: @andresgalante It should be the same as dropdown which should be the same as button. It does not matter which color it is over. A dropdown is at the same level as a button so the shadow should reflect that.

jgiardino: The dropdowns are the same css-wise for select menus and buttons. The visual difference is due to the difference in backgrounds.

However, the [pagination buttons](http://www.patternfly.org/pattern-library/widgets/#pagination) currently have no shadow applied. Is that correct? Should that be fixed here or as part of another story?

dlabrecq (approved):
LGTM

Issue #555, opened 2016-12-20
DataTables bootstrap_input pagination control triggers a js error when the table contains precisely one row


ameron83:
When manually typing a new page number value in the bootstrap input text of a DataTable, if the table contains exactly one row and the new page number is higher than the current table pages (which is 1 since the table must contain only 1 row), I get a javascript error:
`Uncaught TypeError: Cannot read property 'nTr' of undefined`

This seems to be caused by the following line:
`oSettings._iDisplayStart = (Math.ceil((oSettings.fnRecordsDisplay() - 1) / oSettings._iDisplayLength) - 1) * oSettings._iDisplayLength;`

If `oSettings.fnRecordsDisplay()` equals 1, then the resulting value of `oSettings._iDisplayStart` will be negative and that will trigger the error.

PR #565, opened 2017-01-06
Fixed calculation of oSettings._iDisplayStart in patternfly-functions.js

ameron83:
## Description
Fix for DataTables bootstrap_input pagination controller when the table includes exactly one row.

Relates to https://github.com/patternfly/patternfly/issues/555

## Changes
Changed calculation of oSettings._iDisplayStart when iNewStart is greater than oSettings.fnRecordsDisplay()

### DIFF
--- a/src/js/patternfly-functions.js
+++ b/src/js/patternfly-functions.js
@@ -171,7 +171,7 @@
             }
 
             iNewStart = oSettings._iDisplayLength * (this.value - 1);
-            if (iNewStart >= oSettings.fnRecordsDisplay()) {
+            if (iNewStart >= oSettings.fnRecordsDisplay() || iNewStart < 0) {
               /* Display overrun */
               oSettings._iDisplayStart = (Math.ceil(oSettings.fnRecordsDisplay() /
                 oSettings._iDisplayLength) - 1) * oSettings._iDisplayLength;
--- a/src/js/patternfly-functions.js
+++ b/src/js/patternfly-functions.js
@@ -173,7 +173,7 @@
             iNewStart = oSettings._iDisplayLength * (this.value - 1);
             if (iNewStart >= oSettings.fnRecordsDisplay()) {
               /* Display overrun */
-              oSettings._iDisplayStart = (Math.ceil((oSettings.fnRecordsDisplay() - 1) /
+              oSettings._iDisplayStart = (Math.ceil(oSettings.fnRecordsDisplay() /
                 oSettings._iDisplayLength) - 1) * oSettings._iDisplayLength;
               fnDraw(oSettings);
               return;


bleathem:
Thanks for the PR @ameron83.  Have you confirmed that the change you are suggesting doesn't break any existing functionality?  If you could share a code sample we could use to reproduce and verify this fix, that would help us in evaluating this PR.

ameron83:
As far as I can tell, no existing functionality is broken with this fix.
Here is an example ReactJS component that displays datatables. The data now includes 2 items, therefore there is no issue. If you remove one item, the error will show up.

`class ItemsTable extends React.Component {
	constructor(props) {
		super(props);
		this.table = null;
	}

	componentDidMount() {
		this.initTable();
	}

	get tableDOM() {
		return $(ReactDOM.findDOMNode(this.refs.table));
	}

	initTable() {
		this.table = this.tableDOM.DataTable({
			autoWidth: false,
			info: true,
			lengthChange: false,
			data: [{c1: 5, c2: 6}, {c1: 1, c2: 2}],
			language: {
				zeroRecords: ReactDOMServer.renderToStaticMarkup(
					<div className="text-center">
						No matching records found.
					</div>
				),
			},
			dom:
				"<'dataTables_header' <'text-right' i > >" +
				"<'table-responsive' t >" +
				"<'dataTables_footer' p >",
			columns: [
				{
					title: 'Col 1',
					name: 'col1',
					data: 'c1',
					className: 'text-right',
					width: '50%',
				},
				{
					title: 'Col 2',
					name: 'col2',
					data: 'c2',
					className: 'text-right',
					width: '50%',
				},
			],
		});
	}

	render() {
		return (
			<div>
				<table
					ref="table"
					className="datatable table table-striped table-bordered table-hover"
				/>
			</div>
		);
	}
}`

jmikulka:
Hi @ameron83, I've tested your fix on your example and found out that the error is still thrown if you enter anything less than 1.

@bleathem, the original implementation (without @ameron83 fix) throws the error if you enter anything less than 1 or anything greater than the last page.

My working example:
```
import $ from 'jquery';
import React from 'react';

export class Table extends React.Component {
	componentDidMount() {
		$(this.table).DataTable({
			data: [
				{c1: 1, c2: 2},
				{c1: 3, c2: 4},
			],
			columns: [
				{ title: 'c1', name: 'c1', data: 'c1' },
				{ title: 'c2', name: 'c2', data: 'c2' },
			],
		});
	}

	render() {
		return (
			<table
				ref={(table) => { this.table = table; }}
				className="datatable table table-striped table-bordered table-hover"
			/>
		);
	}
}
```

ameron83:
@jmikulka Thanks for pointing out the cases in which an error would still be triggered. I committed a fix that takes care of them

src/js/patternfly-functions.js
```diff
@@ -171,9 +171,9 @@
             }
 
             iNewStart = oSettings._iDisplayLength * (this.value - 1);
-            if (iNewStart >= oSettings.fnRecordsDisplay()) {
+            if (iNewStart >= oSettings.fnRecordsDisplay() || iNewStart < 0) {
               /* Display overrun */
-              oSettings._iDisplayStart = (Math.ceil((oSettings.fnRecordsDisplay() - 1) /
+              oSettings._iDisplayStart = (Math.ceil(oSettings.fnRecordsDisplay() /
```
dlabrecq, line 176: Downloaded the source and tested this solution against our test page. That is, after modifying the table to have only one row.

I see the JS error on Chrome, but Firefox doesn't appear to generate any errors.

Oddly, pagination jumps to the last page when a negative value is entered. Although, this doesn't appear to be related to these changes -- seems like a DataTables quirk to me.

That said, I don't believe folks will often enter negative values, so I'm good with merging. Much better than a JS error.

dlabrecq (approved):
LGTM

PR #567, opened 2017-01-13
Remove direct child selectors from setupVerticalNavigation

### LINKED ISSUE #549: Vertical navigation with item fixed at the bottom
Is there currently any way to display an item fixed at the bottom of vertical navigation?

Tried to use `position: fixed` on the item ([jsfiddle with example](https://jsfiddle.net/knokit/ehum0hqu/)), however, it doesn't display properly on Chrome and Safari when used together with a css animation (e.g.: spinner).

Not querying elements hierarchically, in `bindMenuItemsBehavior()`, as in [this line](https://github.com/patternfly/patternfly/blob/50c3a251e12bee98e23230b5152cb227e41b19db/src/js/patternfly-functions.js#L864), would allow changing the structure of the nav (e.g.: wrap `.list-group` elements with another element).

knokit:
## Description
Attempt to resolve #549 

Removes direct child selectors from _setupVerticalNavigation_ and also simplifies the method to set active items.

Fiddle with example: [jsfiddle](https://jsfiddle.net/knokit/ehum0hqu/4/)

### DIFF
--- a/src/js/patternfly-functions.js
+++ b/src/js/patternfly-functions.js
@@ -843,43 +843,45 @@
       },
 
       bindMenuItemsBehavior = function (handleSelection) {
-        $(document).find('.nav-pf-vertical > .list-group > .list-group-item').each(function (index, primaryItem) {
-          var $primaryItem = $(primaryItem);
+        $(document).find('.nav-pf-vertical .list-group-item').each(function (index, item) {
+          var onClickFn,
+            $item = $(item),
+            $nav = $item.closest('[class*="nav-pf-"]');
 
-          // Set main nav active item on click or show secondary nav if it has a secondary nav bar and we are in the mobile state
-          $primaryItem.on('click.pf.secondarynav.data-api', function (event) {
-            var $this = $(this), $secondaryItem, tertiaryItem;
+          if ($nav.hasClass('nav-pf-vertical')) {
+            // Set main nav active item on click or show secondary nav if it has a secondary nav bar and we are in the mobile state
+            onClickFn = function (event) {
+              var $this = $(this), $secondaryItem, $tertiaryItem, $activeItem;
 
-            if (!$this.hasClass('secondary-nav-item-pf')) {
-              hideSecondaryMenu();
-              if (inMobileState()) {
-                updateMobileMenu();
-                navElement.removeClass('show-mobile-nav');
-              }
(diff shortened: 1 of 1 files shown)


dlabrecq (commented):
I'm not familiar enough with the behaviors of vertical nav. However, I noticed that a couple functions have been removed and replaced with new code. For example, setPrimaryActiveItem, setSecondaryActiveItem, and setTertiaryActiveItem have been replaced with setActiveItem. I believe there may be expected behaviors associated with secondary and tertiary items not necessarily found in our test pages? Perhaps someone more familiar should take a look to see if this is equivalent to what we had.

@jeff-phillips-18 Can you please take a look at this?

jeff-phillips-18:
When a tertiary or secondary item is currently active and a primary (with no secondary) is clicked to become active, only the primary item that was active gets updated to not be active. The secondary and tertiary items still have the active class set.

Same with selecting a secondary item, if a tertiary item was previously active it remains active (though its parent item are no longer active).

knokit:
You're right, I didn't realize that... my changes are not handling those cases.

Is there any specific situation when you'll want that behavior? Because it seems to me that it wouldn't make sense to leave active items in other levels that are not related to the currently active one.

jeff-phillips-18:
I can't think of any reason to leave any other active items.

knokit:
Ok. Will this be good to merge then? Or is there anything I still need to change?

Thanks

jeff-phillips-18:
Sorry, there is still the issue I mentioned above. I guess I misunderstood your question. When an item becomes active there should be no other active items except for it and it's parent items.

knokit:
Correct me if I'm wrong. :) That's how it behaves with my changes.
Here's a link to [rawgit](https://rawgit.com/knokit/patternfly/vertical-nav-dist/dist/tests/vertical-navigation-with-badges.html).

### Original version:
Tertiary item remains active (though its parent is no longer active).
patternfly-vertical-navigation_original

### PR version:
patternfly-vertical-navigation_pr-knokit

jeff-phillips-18:
very strange... when I pull your changes and run I see the issue, when I run your rawgit version I do not. When I run current PF rawgit I also see the issue which previously I did not.

Must be something in my environment 😖

jeff-phillips-18 (approved):
👍

PR #578, opened 2017-01-31
including examples of pagination in card and list views

jgiardino:
## Description
Included pagination in the card and list views.
https://patternfly.atlassian.net/browse/PTNFLY-2108

## Todos
- [x] cross browser test (previously tested for the table view in all the browsers, but still need to check card and list views)
- [x] Is it reponsive?

- [Pagination page](https://rawgit.com/jgiardino/patternfly/Pagination-CardAndListViews-dist/dist/tests/pagination.html) - includes example of just the pagination row (also with links to the view pages, but these are broken)
- [Card View with pagination](https://rawgit.com/jgiardino/patternfly/Pagination-CardAndListViews-dist/dist/tests/pagination-card-view.html)
- [List View with pagination](https://rawgit.com/jgiardino/patternfly/Pagination-CardAndListViews-dist/dist/tests/pagination-list-view.html)
- [Table View with pagination](https://rawgit.com/jgiardino/patternfly/Pagination-CardAndListViews-dist/dist/tests/pagination-table-view.html)

### DIFF
--- a/src/less/pagination.less
+++ b/src/less/pagination.less
@@ -66,10 +66,12 @@
 .content-view-pf-pagination {
   background-color: @table-bg-accent;
   border: 1px solid @table-border-color;
-  border-top: none;
   display: flex;
   flex-wrap: wrap;
   justify-content: space-between;
+  &.table-view-pf-pagination {
+    border-top: none;
+  }
   .form-group {
     align-items: baseline;
     display: flex;
--- a/tests/pages/_includes/widgets/pagination/pagination.html
+++ b/tests/pages/_includes/widgets/pagination/pagination.html
@@ -1,9 +1,9 @@
-<form class="content-view-pf-pagination table-view-pf-pagination clearfix">
+<form class="content-view-pf-pagination {{include.class1}} clearfix">
   <div class="form-group">
     <select class="selectpicker pagination-pf-pagesize">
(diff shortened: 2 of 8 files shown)


jgiardino:
@serenamarie125 @wrnkld @jeperry @catrobson @maryclarke @kybaker 

Please review these pages and let me know if you have any suggestions.

jeperry:
For pagination, will we specify that buttons should be disabled if they are not a valid option? For example you cannot 'go back' when you are on page 1, etc? Otherwise this looks great and covers any use case I can imagine. I'm glad to see the size options.

maryclarke:
Looks good to me @jgiardino

jgiardino:
Thanks, @maryclarke @jeperry!

Regarding the disabled buttons, Dan had been working on the [js implementation of pagination](https://rawgit.com/jgiardino/patternfly/Pagination-CardAndListViews-dist/dist/tests/table-view.html), so in the static example the buttons don't show as disabled but in the js version, they are. I'll see if it's possible to show them as disabled when it's included as a static example.

wrnkld:
looks good @jgiardino

kybaker:
@jgiardino Looks good to me.

andresgalante:
@jgiardino thanks a lot! 😄

PR #585, opened 2017-02-09
PatternFly list

andresgalante:
## Description
This PR adds a bare bones lists, which is a good nest for list-view
The full description is on the jira story [PTNFLY-2188](https://patternfly.atlassian.net/browse/PTNFLY-2188)

@bleathem will be able to give more details on the Jquery implementation.

## Todos
- [ ] cross browser test
- [ ] Is it reponsive?

##  Link to rawgit
https://rawgit.com/andresgalante/patternfly/pf-list-dist/dist/tests/list-pf.html


rhamilto (commented):
Nice work, @andresgalante!

My $0.02: 

* The hover color seems a bit too saturated.  It looks the same as the expanded header color, which is confusing as it suggests the header has already been expanded.
* The drop shadow seems like unnecessary visual noise.
* The doubled border in between sections could be simplified to a single border.
* Should an expand/collapse indicator (e.g., >) be an optional element?  I definitely think it improves the usability.

For reference, here is where we landed on the comparable element within OpenShift.  http://imgur.com/a/PZ8rH

cc: @jwforres

srambach:
I love having this generic listview!
First impression from just looking at the example, not code yet :-)
Is it intended that the hover only works on top-level items?  
I'm not sure whether the gray background on an expanded first line is just to show that it's the first line, or is it that it's selected? I agree also that the colors should be different to better illustrate the different roles of the highlighting, completely apart from whatever the final visual design is. That will also help to demonstrate clearly what is taking precedence - expansion color, selection, or hover.

andresgalante:
@srambach @rhamilto Yeap, that makes a lot of sense, @kybaker Can you please take a look at this?

@rhamilto I will add the arrow as an optional element. I though that since we need to place the arrow on the left in some way (floating, flexing or grid) then that would interfear with how the user wants to build their list.

src/less/variables.less
```diff
@@ -49,6 +49,11 @@
 @img-spinner-xs:                                                    "spinner-xs.gif";
 @input-border-hover:                                                @color-pf-blue-200;
 @input-border-focus:                                                @color-pf-blue-400;
+@list-pf-border-color:                                              @color-pf-black-200;
+@list-pf-active-border-color:                                       @color-pf-black-400;
+@list-pf-header-background-color:                                   @color-pf-black-200;
+@list-pf-hover-background-color:                                    @color-pf-black-200;
+@list-pf-padding:                                                   20px;
```
srambach: Use a spacer variable here? 
Oh, but this is Patternfly 3...nevermind. Next version :)

jgiardino:
This looks really nice, Andres!

Piggy-backing on Sarah and Rob's comments about hover color, I really like how hover works in the table examples here, where the color doesn't change but just becomes more saturated on hover: http://blog.andresgalante.com/patternfly-atomic/?p=viewall-basics-table

I don't know how much of this work is intended to address the visual styling we currently use, but 
it would be nice to avoid what we have here: https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/list-view-simple-expansion.html
where selected list items are highlighted blue, unless they're expanded. It seems like selection is more important and therefore should override the color highlight associated with expansion. 

Since I've been reviewing the [a11y guidelines](http://a11yproject.com/checklist.html), are you planning to add keyboard and screenreader support to this? Supporting keyboard and screenreader access might be simpler if we always include the arrow, by giving screenreaders and keyboard users something to place focus on.

src/less/list-pf.less
```diff
@@ -0,0 +1,47 @@
+//
+// Close icons
```
rhamilto: Comment needs updating

kybaker:
@andresgalante I agree with Robb about removing the shadow. I think if we replaced the hover blue with a lighter gray than the selected state background it would have the same smooth transition that we see in your PF5 data table example.

andresgalante:
@rhamilto @kybaker I've removed the shadow and changed the hover color

rhamilto:
@andresgalante, ETA on when this will merge?  We're assuming it'll be in v3.21.0 in 2 weeks?

andresgalante:
@srambach @jgiardino can we merge this one?

jgiardino:
LGTM!

rhamilto:
@andresgalante, regarding your most recent push:  I wouldn't expect the hover on a non-clickable/non-header area.

rhamilto:
And an already expanded header probably shouldn't hover, either?

srambach:
I was going to add that the light hover on a dark, expanded header seemed like it should darken it further, along the lines of the table example the @jgiardino mentions above.
@rhamilto, to you does the hover only indicate expandable then? Not that you can do something (expand or collapse)?

rhamilto:
> @rhamilto, to you does the hover only indicate expandable then? Not that you can do something (expand or collapse)?

I'm thinking so.  Because the non-header area will have all different kinds of content in it that may or may not be individually clickable.

andresgalante:
:) makes sense, I'll role back that change. 

Does it make sense to add a modifier class to enable or disable hovers in each row or do we just go back to the way it was before?

srambach:
I like the idea of hover by default but a modifier class to disable hovers. Or vice-versa but that seemed the more likely case.

rhamilto:
> Or vice-versa but that seemed the more likely case.

In Bootstrap, hover behavior on tables is opt-in.  I think it would be good to be consistent with what Bootstrap is doing.

jgiardino:
@rhamilto's suggestion to follow bootstrap's model makes sense. But to me, that means one class for `list-pf`. I'm not sure if you were thinking a class for each `list-pf-container` or not when you said this, @andresgalante:

> Does it make sense to add a modifier class to enable or disable hovers in each row

andresgalante:
@srambach @jgiardino @rhamilto I've added a `-hover` class, and it displays a bg when the row is clicable.

The hover class on tables adds the background to all rows.

I am a bit lost here, What do we want to do? 😄 thanks!

rhamilto:
I think the hover should be present on headers that aren't already expanded by default (with no option to disable).

And hover on non-header (content) areas should be opt-in if we deem necessary.  I personally don't think there is a need to have hover on the content area.

[This is what we're using in OpenShift today](http://recordit.co/2c0Y8yDW5r).

jgiardino:
Following up on Rob's comment regarding hover over expanded headers, I could go with either of these solutions:

- no hover color change 
- hover color change to a darker color (but not a lighter color like it is now) e.g.:
```
.list-pf-hover .active > .list-pf-container:not(:last-child):hover {
    background-color: #dfdfdf;
}
```
_Although in this example, I don't think the `:not(:last-child)` is needed? And `#dfdfdf` is just a random darker color for example purposes but I think it's too dark._

I also don't think there is a need to have hover over the content area, but I don't think I would make that optional either. That seems like such an edge case that it would need custom code anyway.  

I don't have a strong opinion on whether the hover over headers should be optional.

andresgalante:
ok, lets try this one more time.
I've removed the opt-in hover and added hovers only on non active "headers", like the open shift example.

Does it make more sense now?

rhamilto:
I'm still seeing the hover color on "open" headers.  :/

andresgalante:
@rhamilto wait a few mins, travis takes time to compile and re deploy

rhamilto:
@spadgett and I just reviewed the latest and noticed a couple things:

* the cursor should probably change to pointer (hand) when mousing over the headers.
* currently there is no keyboard navigation.  should there be?
* is the chevron obvious enough?  @spadgett and I are partial to the larger chevron with a "pipe" that we're using in OpenShift.  


 [1]

screen shot 2017-02-17 at 11 29 36 am

screen shot 2017-02-17 at 11 31 10 am

screen shot 2017-02-17 at 11 31 15 am

kybaker:
@andresgalante I agree with Robb about the pipe and chevron. I know the focus was not on the content in your example, but these elements are a part of the interaction.

andresgalante:
Thanks! I've added the pipe and increase the size of the chevron. Also keyboard navigation now works.

Thanks for the reviews!

rhamilto:
The chevrons should be pointed down since these two sections are expanded by default.

screen shot 2017-02-17 at 4 07 17 pm

andresgalante:
you are right, just sent an update, thanks

PR #591, opened 2017-02-13
Fix broken contextual panels in accordion

nireno:
## Description
This fixes issue [#511](https://github.com/patternfly/patternfly/issues/511) which caused contextual panels to break when used inside .panel-group. 

@andresgalante

PR #598, opened 2017-02-16
Fix broken contextual panels inside accordion

nireno:
Styles for default panel were leaking into panels in other states (such as
success or warning). This fix constrains the default style and adds additional
selectors for the other states.

Since there is a visual overlap of borders in the accordion, a hierarchy
(danger > warning > success > primary) is enforced. For example the top and
bottom borders for a .panel-danger will take precedence over all other panel
types.

The test is added under panels.
@andresgalante @kybaker


nireno:
Duplicate of [#591](https://github.com/patternfly/patternfly/pull/591)

Issue #479, opened 2016-09-18
It's better to let the user resize only the height of <textarea>


dabeng:
![screenshot from 2016-09-18 14-19-47](https://cloud.githubusercontent.com/assets/5689075/18613400/a214b2b0-7d3f-11e6-9063-dd896fa184f8.png)

andresgalante:
@dabeng you are absolutely right. `resize: vertical` should do it:
https://developer.mozilla.org/en/docs/Web/CSS/resize

Do you want to send a PR with it?

PR #599, opened 2017-02-16
just allowing users to resize the textarea vertically

dabeng:
## Description
allowing the users to resize the textarea on wizard, which may only be resized vertically. This PR aims to fix the [bug](https://github.com/patternfly/patternfly/issues/479).

You can review the [result](https://cdn.rawgit.com/dabeng/patternfly/wizard-v-dist/dist/tests/wizard.html).

@andresgalante , would you like to review this PR?

### DIFF
--- a/src/less/wizard.less
+++ b/src/less/wizard.less
@@ -215,6 +215,10 @@
   }
 }
 
+.wizard-pf-contents textarea {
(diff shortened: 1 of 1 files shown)


andresgalante:
👍  thanks!

Issue #491, opened 2016-09-20
really, really long words overflow the inline error notification


andreasn:
We have an issue in the Image Registry where really long strings overflow the inline notification error.
The error itself warns against names that are 63+ characters long, but if the word is 100+ letters long, that word will overflow the notification itself.

![screenshot from 2016-09-20 12-57-50](https://cloud.githubusercontent.com/assets/209895/18667852/86109e36-7f32-11e6-9eaf-d34ef0cbbf1e.png)

Filed in Cockpit here:
https://github.com/cockpit-project/cockpit/issues/5068

andreasn:
word-wrap: break-word; might be a good solution here.

andresgalante:
@andreasn Thanks a lot for reporting this, would you like to send a PR with the `word-wrap` fix?

andreasn:
Sure. I'll get one together.

andresgalante:
Awesome, you are the best 😄  !

andresgalante:
@andreasn did we ever merge a PR related to this issue?

andreasn:
no, I very much forgot it.
Let me take care of that this week.

andresgalante:
:) thanks you are awesome!

PR #601, opened 2017-02-16
Avoid long words overflowing alerts in narrow situations

andreasn:
Really long words broke outside the alert in narrow situations.

Fixes issue #491


andresgalante:
Thanks a lot!

Issue #337, opened 2016-07-07
Timepicker loses background and border on window resizing


dabeng:
![screenshot from 2016-07-07 10-29-52](https://cloud.githubusercontent.com/assets/5689075/16640861/440c075c-43c1-11e6-999c-0043d936c24e.png)

andresgalante:
@dabeng I can't reproduce it on the website or on the test page: https://rawgit.com/patternfly/patternfly/master/tests/time-picker.html

What browser are you using?

dlabrecq:
I also cannot reproduce on Firefox, Chrome and Safari.

dabeng:
![chrome](https://cloud.githubusercontent.com/assets/5689075/16675351/ddd564d6-4485-11e6-99ac-468cee3b3cfc.jpg)
![ie](https://cloud.githubusercontent.com/assets/5689075/16675350/dd566ce4-4485-11e6-8cc1-515430b46677.jpg)
![opra](https://cloud.githubusercontent.com/assets/5689075/16675349/dd55c8e8-4485-11e6-9017-f2d02c78a12c.jpg)

Hi @andresgalante @dlabrecq , the problem has nothing to do with specific browser. It's caused by window resizing shown as above.

srambach:
I replicated the behavior in Win 8 Chrome on the Patternfly page. However, on the same browser, it doesn't happen in the Rawgit page. 
I can't replicate it in OSx or Win 10 at all.
![bs_win10_chrome_51 0](https://cloud.githubusercontent.com/assets/19825616/16686297/11f06abe-44df-11e6-827e-b4012e01d6e7.jpg)

dabeng:
Hi guys, I really get the same issue on standalone local test page on Chrome 50, Opera 32 and Firefox 47 under Fedora 22.
![screenshot from 2016-07-11 11-56-26](https://cloud.githubusercontent.com/assets/5689075/16719604/1dd77e02-46f2-11e6-9953-1b46c5ee584a.png)
![screenshot from 2016-07-11 11-56-46](https://cloud.githubusercontent.com/assets/5689075/16719605/1dd99a20-46f2-11e6-99c2-c31126ced8d5.png)

LHinson:
Looks like this is an issue with the pattern page vs an issue with the component. I've created a bug here to address when we can: https://patternfly.atlassian.net/browse/PTNFLY-2178

michpetrov:
[...]

PR #604, opened 2017-02-22
PTNFLY-2178: fix time-picker repositioning

michpetrov:
## Description
Fix for #337. The timepicker is set to reposition itself if there's not enough room below the input (switching from `.dropdown-menu.bottom` to `.dropdown-menu.top`). The original rule was set globally so it wasn't working.

I opted for negative margin to avoid using `!important`, since that's what was causing the issue in the first place. 

## Todos
- [ ] cross browser test (tested in Chrome, Firefox, Opera)
- [ ] Are you sure it works on IE9? (I would assume so)
- [x] Is it reponsive?

[Rawgit](https://rawgit.com/michpetrov/patternfly/time-picker-fix-dist/dist/tests/time-picker.html)

### DIFF
--- a/src/less/time-picker.less
+++ b/src/less/time-picker.less
@@ -16,10 +16,15 @@
   &.dropdown-menu {
     left: 0!important;
     padding: 0;
-    top: 23px!important;
     width: calc(~"100% - 25px");
     &:before, &:after {
       content: none;
+    }
+    &.top {
+      margin-bottom: -1px;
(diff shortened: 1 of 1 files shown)


andresgalante:
Thanks a lot! 😄 

👍  @bleathem I am ok to merge this one

bleathem:
Great, thanks for providing the fix @michpetrov

PR #614, opened 2017-03-09
autoprefixer: support old firefox versions and placeholder fix

### LINKED ISSUE #597: Removal of vendor prefixes from placeholder mixin results in inconsistent styles
@priley86's [removal of vendor prefixes in the placeholder mixin](https://github.com/patternfly/patternfly/blame/v3.10.0/less/mixins.less#L78) resulted in the introduction of inconsistent placeholder styles since [all browsers do not yet support un-prefixed placeholder pseudo elements](http://caniuse.com/#feat=css-placeholder).  This is resulting in a [bug in OpenShift](https://github.com/openshift/origin-web-console/issues/1259) where we have inconsistent placeholder text across Chrome, IE/Edge and Firefox, Safari.

Chrome 56 / Firefox 51
screen shot 2017-02-15 at 1 39 48 pm

cc: @sg00dwin, @spadgett

priley86:
## Description
This fix adds other Firefox version prefixes to autoprefixer that were missed before, along with a fix for `form-control::placeholder` prefixed css.

Fixes #597 

cc @andresgalante @rhamilto

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,10 @@ Development setup requires Node.js and Ruby. If you do not already have Node.js,
 
 The PatternFly code includes a number of dependencies that are not committed to this repository.  To add them, follow the instructions below under "Install NPM Dependencies".  Please make sure you keep them updated (see [Keeping NPM Dependencies Updated](#keeping-npm-dependencies-updated)).
 
+## Autoprefixer
+
+Patternfly uses [Autoprefixer](https://github.com/postcss/autoprefixer) to auto add prefixes to its output CSS. Since Patternfly extends some of the core Bootstrap3 less which contains prefixes, we also explicitly add prefixes in these cases to ensure backwards compatibility with Bootstrap3. If consuming Patternfly LESS and compiling, you can define your own target prefixes using [browserlist](https://github.com/ai/browserslist).
+
 ### Install NPM Dependencies
 
 The development includes the use of a number of helpful tasks. In order to setup your development environment to allow running of these tasks, you need to install the local nodejs packages declared in `package.json`. To do this run:
--- a/src/less/mixins.less
+++ b/src/less/mixins.less
@@ -75,7 +75,10 @@
 
 // Placeholder text
 .placeholder(@color: @input-color-placeholder) {
-  &::placeholder            { color: @color; font-style: italic; }
+  &:-moz-placeholder            { color: @color; font-style: italic; } // Firefox 4-18
+  &::-moz-placeholder           { color: @color; font-style: italic; opacity: 1; } // Firefox 19+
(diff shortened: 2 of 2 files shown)


andresgalante:
👍  it works for me. Thanks a lot Patrick

rhamilto:
I think I see a downside to this solution.  If you're working from PatternFly's source less (which we are in OpenShift), you have to be using autoprefixer (which we are, but I don't see that listed as a requirement in PatternFly's readme) **and** your autoprefixer settings have to match PatternFly's.  Currently our autoprefixer settings are 'last 3 versions', so applying the autoprefixer settings change in OpenShift is going to result in additional prefixes that we don't necessarily need (disregarding the fact that I really probably shouldn't have to do that for a dependency to work), all for the sake of fixing placeholder styles (additionally, I'm having trouble getting PatternFly's autoprefixer settings to run with our outdated version of grunt-autoprefixer).  Since PatternFly is overriding [Bootstrap's placeholder mixin](https://github.com/twbs/bootstrap/blob/master/less/mixins/vendor-prefixes.less#L100-L109), I think the better solution is to include the vendor prefixes in PatternFly source less.

Thoughts?

cc:  @jwforres, @spadgett

priley86:
@rhamilto not really sure I follow this logic. If you are consuming the PF less directly, you can modify the less to your liking (and extend with prefixes or modify the autoprefixer build to your liking). If you are consuming PF CSS directly, you would get the autoprefixed styles (which we'd likely want to include more prefixes for better support in that case). I guess I'm looking at this from the other angle which is, why assume downstream will want those prefixes in the less? Happy to change this and add them if we decide that's what helps most teams though. Also, will update README with a note about Autoprefixer settings in this PR.

rhamilto:
> why assume downstream will want those prefixes in the less

Because, presumably, you would want the compiled less to be equivalent of the upstream CSS.  

Let me try to restate things:  PatternFly includes Bootstrap's source less.  Bootstrap's source less [includes vendor prefixes](https://github.com/twbs/bootstrap/blob/master/less/mixins/vendor-prefixes.less#L100-L109).  In order for PatternFly's source less for placeholders to correctly override Bootstrap's source less, it too must include the vendor prefixes.

priley86:
i've updated this PR to mimic upstream BS3 and have explicitly added all `::placeholder` prefixes to ensure backwards compatibility of the `.placeholder()` mixin. I've also appended a note about how to change `autoprefixer` settings in the README.

Anything else here @rhamilto ?

rhamilto:
> Anything else here @rhamilto ?

I don't think so.  Thanks, @priley86!

priley86:
Sure, again thanks for reporting @rhamilto and working thru this.

@dlabrecq @andresgalante ok to merge this?

PR #618, opened 2017-03-16
Fix NO_EFFECT warning in Coverity scanning

stefwalter:
This is a loop with a tiny bit of dead code. It's not a big deal
but changing this line helps projects that use Patternfly not
have to deal with another waiver.

```
Defect Type: NO_EFFECT
remediation: Perhaps this code is incomplete or mistyped in some way.
```


stefwalter:
CC @dperpeet

bleathem:
This fix looks fine to me.  Thanks for the PR @stefwalter

dgutride (approved):

PR #619, opened 2017-03-17
List view example 1

srambach:
## Description
This PR implements example list view rows in list-pf
[PTNFLY-2199](https://patternfly.atlassian.net/browse/PTNFLY-2199)
Note: Checkbox selection is not currently working because the click is directed to expanding the list item. @bleathem please take a look?

## Todos
- [x] cross browser test
- [x] Tested on IE 10, 11+
- [x] Is it responsive?

Look at the List test page.

```sh
git pull --prune
git checkout <feature_branch>
bundle; script/server
```

Please review @andresgalante @jgiardino


andresgalante:
Here is a preview:
https://rawgit.com/andresgalante/patternfly/list-view-example-1-dist/dist/tests/list-pf.html

andresgalante:
@srambach I ❤️ it. 
I'll take a closer look at it later today, but the overview I just did looks excellent.
Did you tested on IE10+?

andresgalante:
It'd be good if when they are not stacked additional info gets vertically centered:
screen shot 2017-03-20 at 3 14 23 pm

jgiardino:
Great work, @srambach!

I agree with Andres's comment about vertically centering all the elements when not stacked.

The only question I have is whether the class `list-pf-ellipse-text` should be generalized as `pf-ellipse-text` so that it can be used anywhere, and not just in lists?

andresgalante:
cc: @rhamilto and @jwforres I just wanted to keep you on the loop that this is happening. A much flexible and nicer solution to list view to replace the old one on pf4.

It supports long stings and solves all the issues you guys mention you were having with list view.

andresgalante:
@srambach Is the color of the separator on the right of the checkbox right? it look darker than what we have now.

bleathem:
This looks really good.  I love the responsive behaviour.  Good job @srambach!

srambach:
@andresgalante You are correct that it is darker than the current list-vew implementation, which is #d1d1d1 (@color-pf-black-300).  However, it's still using the variable put in when you did the generic list, which is @list-pf-active-border-color which is @color-pf-black-400, or #bbb. Shall I change @list-pf-active-border-color to be @color-pf-black-300 or was that change part of the new design? @kybaker can you say which color the separator line, and the borders when expanded, should be?

andresgalante:
I really don't recall why I used `@list-pf-active-border-color` probably my mistake. if `black-300` works lets go with it.

andresgalante:
👍 thanks a lot @srambach ! This looks great, I am ok to merge it.

@jgiardino ?

jgiardino:
In `.list-pf-stacked` everything is vertically centered to the top except for the contents of `.list-pf-content-wrapper`.  Can we use `:not(.list-pf-stacked) .list-pf-content-wrapper {align-items: center;}` so that only the non-stacked list item contents get centered?

andresgalante:
@srambach sorry I missed that one, @jgiardino is right, when stacked, additional content should be aliened top.

srambach:
Great catch @jgiardino! I simplified it to just select `.list-pf-stacked .list-pf-content-wrapper` and then align to flex-start so the default is centering. What do you think?

jgiardino:
Sure, that works.

andresgalante:
Cool I'll merge it then. Thanks a lot @srambach this is a superb job.

PR #628, opened 2017-04-13
NPM Script Update & Readme

cdcabrera:
## Description

Goal, simplify install process for users unfamiliar with dev installs. Implementation of NPM scripts as aliases/shortcuts towards Grunt tasks and automated Jekyll postinstall. 

May need some review input on the "jekyll" NPM script within package.json to verify syntax. Per @bleathem removed redundant jekyll install.

## Todos
- N/A, cross browser test
- N/A, Are you sure it works on IE9?
- N/A, Is it reponsive?

```sh
$ git pull https://github.com/cdcabrera/patternfly.git NPM-Scripts-4x:NPM-Scripts-4x
$ npm install
$ npm start
```

@bleathem

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ PatternFly can be installed and managed through [NPM](https://www.npmjs.com/). T
 npm install patternfly --save
 ```
 
-Patternfly stays up to date with the Node LTS [Release Schedule](https://github.com/nodejs/LTS#lts_schedule). If you're using Patternfly downstream, we suggest the use of an actively supported version of Node/NPM, although prior versions of Node may work.
+PatternFly stays up to date with the Node LTS [Release Schedule](https://github.com/nodejs/LTS#lts_schedule). If you're using PatternFly downstream, we suggest the use of an actively supported version of Node/NPM, although prior versions of Node may work.
 
 ### Install with Bower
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -89,8 +101,15 @@ Additionally you may need to install the grunt command line utility.  To do this
...
-    gem install bundle
-    bundle install
+```
+npm run jekyll
+```
+
+or
+
+```
+gem install jekyll -v 3.1.6
```
bleathem: The preferred way to install jekyll is using `bundle install` so we all have the same dependency versions.

bleathem: So +1 for the `npm run`wrapper, but the alternate command should be to use `bundle install`

bleathem (changes requested):

README.md
cdcabrera: Update applied. Looks like the rebase messed around with the review comment... should now appear as
```
gem install bundle
bundle install
```

package.json
```diff
@@ -65,7 +66,11 @@
     "patternfly-bootstrap-treeview": "~2.1.0"
   },
   "scripts": {
-    "test": "./node_modules/karma/bin/karma start"
+    "test": "grunt karma",
+    "build": "grunt build",
+    "jekyll": "ruby --version || echo \"Ruby required\"; gem install bundler || sudo gem install bundler; bundle install; set -e",
```
bleathem: Let's drop the sudo part here

bleathem (changes requested):

package.json
cdcabrera: updated

bleathem (approved):

PR #631, opened 2017-04-18
Fix Datatable Api pfEmpty.updateState()

iransamaneh:
in patternfly.dataTables.pfEmpty.js registered Datatable Api pfEmpty.updateState() has a wrong function call.

## Description
update function is not exists but updateState is correct function. it is defined in call scope and accessible but update function is not exist and using Datatable Api lead to error


bleathem (approved):
This change looks reasonable to me, as the `update` function does not exist in the codebase.

Can you please confirm @dlabrecq?

dlabrecq (approved):
LGTM

PR #632, opened 2017-04-19
Feature Application Launcher 3x

cdcabrera:
## Description

Styling for Angular-PF Application Launcher

## Todos
- [x] cross browser test
- [x] Is it reponsive?

https://rawgit.com/cdcabrera/patternfly/Feature-AppLauncher-3x-dist/dist/tests/application-launcher-nav.html

@LHinson @andresgalante @cshinn

### DIFF
--- a/src/less/application-launcher.less
+++ b/src/less/application-launcher.less
@@ -1,0 +1,193 @@
+//
+// Application Launcher
+// --------------------------------------------------
+
+.applauncher-pf {
+
+  display: inline-block;
+  overflow: visible;
+
+  .applauncher-pf-title {
+    .sr-only;
+  }
+
+  .dropdown-toggle {
+    &.disabled {
+      cursor: not-allowed;
+    }
+  }
+
+  .dropdown-menu {
+    padding: @applauncher-pf-dropdown-menu-padding;
+    min-width: @applauncher-pf-dropdown-menu-width;
+  }
+
+  .applauncher-pf-link {
+    border-style: solid;
+    border-width: @applauncher-pf-menu-link-border-width;
+    border-color: transparent;
+    padding: @applauncher-pf-menu-link-padding;
+
+    &:hover {
+      background-color: @applauncher-pf-menu-link-background-color-hover;
+      border-color: @applauncher-pf-menu-link-border-color-hover;
+      color: @applauncher-pf-menu-link-color-hover;
+      text-decoration: none;
+
+      .box-shadow(0 0 2px 0 @applauncher-pf-menu-link-shadow);
+    }
+  }
+
+  .applauncher-pf-link-icon {
+    font-size: @applauncher-pf-menu-link-icon-font-size;
+  }
+
+  .applauncher-pf-link-icon,
+  .applauncher-pf-link-title {
+    display: inline-block;
+    width: auto;
+    vertical-align: middle;
+  }
+
+  &.applauncher-pf-block-list {
+
+    .applauncher-pf-item {
+      display: inline-block;
+      min-width: 0;
+      text-align: center;
+      vertical-align: top;
+      width: calc(50% - (@applauncher-pf-menu-link-border-width * 2));
+    }
+
+    .applauncher-pf-link {
+      @media (min-width: @screen-sm-min) {
+        padding: @applauncher-pf-menu-link-padding-block-list;
+      }
+    }
+
+    .applauncher-pf-link-icon,
+    .applauncher-pf-link-title {
+      display: block;
+      padding: 0;
+    }
+
+    .applauncher-pf-link-title {
+      white-space: normal;
+      word-wrap: break-word;
+    }
+  }
+}
+
+.navbar-pf,.navbar-pf-alt {
+
+  .navbar-utility .applauncher-pf,
+  .applauncher-pf {
+    display: block;
+    overflow: hidden;
+
+    > a:hover {
+      background-color: @navbar-pf-navbar-utility-hover-bg-color;
+    }
+
+    &.open > a,
+    &.open > a:focus {
+      background-color: @navbar-pf-navbar-nav-active-bg-color;
+      color: @navbar-pf-active-color;
+
+      @media (min-width: @screen-sm-min) {
+        background-color: @navbar-pf-navbar-utility-open-bg-color;
+        border-color: @navbar-pf-navbar-utility-border-color;
+        color: @navbar-pf-navbar-utility-color;
+      }
+    }
+
+    .dropdown-menu {
+      border-width: @applauncher-pf-menu-link-border-width !important;
+
+      @media (min-width: @screen-sm-min) {
+        margin-top: 11px;
+      }
+    }
+
+    &.dropdown > .dropdown-toggle,
+    .dropdown-toggle {
+      background-color: inherit;
+      color: @applauncher-pf-menu-link-color;
+      text-align: left;
+      text-decoration: none;
+      border-width: 0;
+      display: block;
+      padding-left: 20px;
+
+      &.disabled {
+        color: @applauncher-pf-menu-link-color-disabled !important;
+      }
+
+      @media (min-width: @screen-sm-min) {
+        border-left: 1px solid @navbar-pf-navbar-utility-border-color;
+        padding: @applauncher-pf-dropdown-toggle-padding;
+        line-height: 1;
+      }
+    }
+
+    .applauncher-pf-icon {
+      padding-right:@applauncher-pf-icon-padding-right;
+
+      @media (min-width: @screen-sm-min) {
+        padding: 0;
+      }
+    }
+
+    .applauncher-pf-title {
+      display: inline;
+      position: relative;
+    }
+
+    .applauncher-pf-item {
+      display: block;
+      min-width: inherit;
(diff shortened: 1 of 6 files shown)


dtaylor113:
Is this a new test page for application launcher?  If so, it seems like overkill.  It's unfortunate that test page initially shows a toast notification over the app. launcher:
image
Is there a way to remove the toast notification and actually most of the page.  I think we'd only need the top nav bar with the app. launcher icon.  We don't need the vert. nav, dashboard, or charts.

cdcabrera:
Agree on the toast notification, taking a look. @dtaylor113 recommendations on trimming it down? And just realized you made a recommendation... taking a look

dtaylor113:
> @dtaylor113 recommendations on trimming it down?

If it's a test page for the application launcher, then I would just show the app. launcher in the top nav bar.  No need for vert. nav, dashboard, or charts -they are just a distraction :-)  -thanks!

dtaylor113:
Also, when I pull down your branch and do 'grunt server'.  I see at 'localhost:9000':

image
Shouldn't App. Launcher be under Navigation instead of Layout?

cdcabrera:
Once the example gets trimmed down, can move under Navigation so it matches Angular-PatternFly. Was based off of the Notification Drawer behavior/example

andresgalante:
Hi @cdcabrera I haven't seen the code yet, but to me there are design issues.

@jennyhaines @LHinson 
- The hover state is inconsistent with other hovers, check for example the drop menu on the right next to the app launcher.
- Tooltip is useless, each link has an icon and a label, do we really need a tooltip?
- The component is too large IMO, comparing to the scale of any other element in PatternFly.
- What exactly is the "home" link?

src/less/application-launcher.less
```diff
@@ -0,0 +1,141 @@
...
+.applauncher-pf {
+
+  display: inline-block;
+  overflow: visible;
+
+  .applauncher-pf-title {
+    display: none;
+  }
+
+  .dropdown-menu {
```
srambach: Can you put the px values that are sizing this component into variables to make it easier to adjust if needed?

src/less/application-launcher.less
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  .applauncher-pf-list-default {
+
+    .applauncher-pf-entry {
+      display: inline-block;
+      min-width: 0;
+      text-align: center;
+      vertical-align: top;
+      width: 49%;
```
srambach: You can probably use calc() to use 50% and subtract off the padding so that this is more precise.

src/less/application-launcher.less
```diff
@@ -0,0 +1,141 @@
+//
+// Application Launcher
+// --------------------------------------------------
+
+.applauncher-pf {
+
+  display: inline-block;
```
andresgalante, line 7: Why does it have to be `inline-block`?

src/less/application-launcher.less
```diff
@@ -0,0 +1,141 @@
...
+  overflow: visible;
```
andresgalante, line 8: Why are you using `overflow` to hide/show the drop menu?

src/less/application-launcher.less
```diff
@@ -0,0 +1,141 @@
...
+
+  .applauncher-pf-title {
```
andresgalante, line 10: What is the title? and why are you hiding it?

src/less/application-launcher.less
```diff
@@ -0,0 +1,141 @@
...
+    display: none;
+  }
+
+  .dropdown-menu {
+    padding: 9px;
```
andresgalante: this padding doesn't seem to do anything

src/less/application-launcher.less
```diff
@@ -0,0 +1,141 @@
...
+    min-width: 240px;
```
andresgalante: Why do we set a height? If the menu is empty then we don't want to display it anyway

src/less/application-launcher.less
cdcabrera: for sure, are you thinking these vars need to be within the same file at the top?

src/less/application-launcher.less
cdcabrera: taking a look

src/less/application-launcher.less
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  .applauncher-pf-link {
+    border: 1px solid @color-pf-white;
+    padding: 6px 10px;
+
+    &:hover {
+      background-color: transparent;
+      border: 1px solid @color-pf-black-400;
```
andresgalante: just `border-color: @color-pf-black-400` will do it.

src/less/application-launcher.less
cdcabrera: There are two aspects to the behavior, standalone and within the nav itself. The inline-block is aimed at giving it the stand-alone ability outside of the navigation. We can look at updating if the standalone ability is considered to much extra.

src/less/application-launcher.less
cdcabrera: sure thing

src/less/application-launcher.less
```diff
@@ -0,0 +1,141 @@
...
+  .applauncher-pf-link-title {
+    display: inline-block;
+    width: auto;
+  }
+
+  .applauncher-pf-link-icon {
+    padding-right: 4px;
+  }
+
+  .applauncher-pf-list-default {
```
andresgalante: What is `.applauncher-pf-list-default` doing? If it's a reponsive class, then it should be scoped with mediqueries.

src/less/application-launcher.less
cdcabrera: On a related note for the Angular-PF aspect, good catch on the dimensions. The list of items/links is currently the requirement for the Angular-PF component, we can look at making sure it deactivates itself if it isn't provide a list

jennyhaines:
Hi @andresgalante @LHinson @cdcabrera 

In response to the design issues,

- I see that the launcher hover state is inconsistent with other dropdown menu hover, but am not concerned with this fact, since the launcher menu is more visual with the use of icons. I don't think it would look right to have this same light blue background highlight style. I'm curious what the rest of you think!

- I can see that we might not need the tooltip. I'm thinking that if the titles are descriptive enough, we wouldn't need the tooltip. 

- I can see that about the icon size being too large in this "tile view" launcher menu. In the original PSD design, the masthead icons are larger, so there is less of a difference. I'd recommend to reduce the size of the Launcher Menu icons to a maximum height of 30px.

- The "home" link would link the user to the Launcher home screen so that they they can see the the function descriptions, and access the additional links such as "Getting Started", "Learn More", etc.

andresgalante:
@jennyhaines, thanks for the feedback! The hover state is also different from cards and list view and anywhere else we use icons.

jennyhaines:
@andresgalante @LHinson @cdcabrera

I like the update to the icon sizing and the removal of the tooltips- looks better, for sure! Good ideas, Andres. As for the styling to tie it to other hover instances, I propose that we do one of the following. Let me know what you think.: 

1-
We visually mirror the hover style used on list view compound expansions. This would mean that on hover, the container that hold the particular link would turn grey, and the icon would be blue. (see [http://www.patternfly.org/pattern-library/content-views/list-view/#/design](url) - I much prefer this option!) 
listview-compoundexpansion


or 2-
We visually mirror the card hover style (I believe this is too subtle)
[http://www.patternfly.org/pattern-library/content-views/card-view/#/design](url)
cardview-hover

cdcabrera:
@jennyhaines @andresgalante @LHinson 

I'll give option 1 you presented a go... ```"We visually mirror the hover style used on list view compound expansions."```

cdcabrera:
@jennyhaines 

Update is applied

- https://rawgit.com/cdcabrera/patternfly/Feature-AppLauncher-3x-dist/dist/tests/application-launcher-nav.html

jennyhaines:
I think it looks great, @cdcabrera - the only inconsistency with our existing PatternFly style is having the title stay that dark grey color, rather than turn blue. That's my only suggestion! @andresgalante @LHinson , thoughts?

andresgalante:
@cdcabrera where can we preview the version without icons?

cdcabrera:
@andresgalante I'll look at creating another header within the same sample file

dtaylor113:
> I'll look at creating another header within the same sample file

Pls take a look at List View where we have different toggles/options to affect the ngdoc example.   I can see a dropdown for 'Grid' and 'List', checkbox for Disabled, etc.

cdcabrera:
@andresgalante the sample should be updated now... 
- https://rawgit.com/cdcabrera/patternfly/Feature-AppLauncher-3x-dist/dist/tests/application-launcher-nav.html

bleathem:
Today is release day.  @cdcabrera @andresgalante is this PR to be merged before the release?

cdcabrera:
@bleathem still awaiting the final "all clear" on the related 4x PR #636 . 

PRs in [AngularPF 3x](https://github.com/patternfly/angular-patternfly/pull/446) & [AngularPF 4x](https://github.com/patternfly/angular-patternfly/pull/453) are dependents.

andresgalante:
Thanks @cdcabrera

PR #636, opened 2017-04-25
Feature Application Launcher 4x

cdcabrera:
Styling for Angular-PF Application Launcher, ported from 3x PR

## Description
A few sentences describing the overall goals of the pull request's commits.
Include related PRs or Issues

## Todos
- [x] cross browser test
- [x] Is it reponsive?

https://rawgit.com/cdcabrera/patternfly/Feature-AppLauncher-4x-dist/dist/tests/application-launcher-nav.html

@LHinson @andresgalante @cshinn

### DIFF
--- a/src/less/application-launcher.less
+++ b/src/less/application-launcher.less
@@ -1,0 +1,193 @@
+//
+// Application Launcher
+// --------------------------------------------------
+
+.applauncher-pf {
+
+  display: inline-block;
+  overflow: visible;
+
+  .applauncher-pf-title {
+    .sr-only;
+  }
+
+  .dropdown-toggle {
+    &.disabled {
+      cursor: not-allowed;
+    }
+  }
+
+  .dropdown-menu {
+    padding: @applauncher-pf-dropdown-menu-padding;
+    min-width: @applauncher-pf-dropdown-menu-width;
+  }
+
+  .applauncher-pf-link {
+    border-style: solid;
+    border-width: @applauncher-pf-menu-link-border-width;
+    border-color: transparent;
+    padding: @applauncher-pf-menu-link-padding;
+
+    &:hover {
+      background-color: @applauncher-pf-menu-link-background-color-hover;
+      border-color: @applauncher-pf-menu-link-border-color-hover;
+      color: @applauncher-pf-menu-link-color-hover;
+      text-decoration: none;
+
+      .box-shadow(0 0 2px 0 @applauncher-pf-menu-link-shadow);
+    }
+  }
+
+  .applauncher-pf-link-icon {
+    font-size: @applauncher-pf-menu-link-icon-font-size;
+  }
+
+  .applauncher-pf-link-icon,
+  .applauncher-pf-link-title {
+    display: inline-block;
+    width: auto;
+    vertical-align: middle;
+  }
+
+  &.applauncher-pf-block-list {
+
+    .applauncher-pf-item {
+      display: inline-block;
+      min-width: 0;
+      text-align: center;
+      vertical-align: top;
+      width: calc(50% - (@applauncher-pf-menu-link-border-width * 2));
+    }
+
+    .applauncher-pf-link {
+      @media (min-width: @screen-sm-min) {
+        padding: @applauncher-pf-menu-link-padding-block-list;
+      }
+    }
+
+    .applauncher-pf-link-icon,
+    .applauncher-pf-link-title {
+      display: block;
+      padding: 0;
+    }
+
+    .applauncher-pf-link-title {
+      white-space: normal;
+      word-wrap: break-word;
+    }
+  }
+}
+
+.navbar-pf,.navbar-pf-alt {
+
+  .navbar-utility .applauncher-pf,
+  .applauncher-pf {
+    display: block;
+    overflow: hidden;
+
+    > a:hover {
+      background-color: @navbar-pf-navbar-utility-hover-bg-color;
+    }
+
+    &.open > a,
+    &.open > a:focus {
+      background-color: @navbar-pf-navbar-nav-active-bg-color;
+      color: @navbar-pf-active-color;
+
+      @media (min-width: @screen-sm-min) {
+        background-color: @navbar-pf-navbar-utility-open-bg-color;
+        border-color: @navbar-pf-navbar-utility-border-color;
+        color: @navbar-pf-navbar-utility-color;
+      }
+    }
+
+    .dropdown-menu {
+      border-width: @applauncher-pf-menu-link-border-width !important;
+
+      @media (min-width: @screen-sm-min) {
+        margin-top: 11px;
+      }
+    }
+
+    &.dropdown > .dropdown-toggle,
+    .dropdown-toggle {
+      background-color: inherit;
+      color: @applauncher-pf-menu-link-color;
+      text-align: left;
+      text-decoration: none;
+      border-width: 0;
+      display: block;
+      padding-left: 20px;
+
+      &.disabled {
+        color: @applauncher-pf-menu-link-color-disabled !important;
+      }
+
+      @media (min-width: @screen-sm-min) {
+        border-left: 1px solid @navbar-pf-navbar-utility-border-color;
+        padding: @applauncher-pf-dropdown-toggle-padding;
+        line-height: 1;
+      }
+    }
+
+    .applauncher-pf-icon {
+      padding-right:@applauncher-pf-icon-padding-right;
+
+      @media (min-width: @screen-sm-min) {
+        padding: 0;
+      }
+    }
+
+    .applauncher-pf-title {
+      display: inline;
+      position: relative;
+    }
+
+    .applauncher-pf-item {
+      display: block;
+      min-width: inherit;
+      text-align: left;
+      width: auto;
+    }
+
+    .applauncher-pf-link {
+      overflow: hidden;
+      width: 100%;
+
+      &:hover {
+        background-color: transparent;
+        border-color: transparent;
+        .box-shadow(none);
+      }
+    }
+
+    .applauncher-pf-link-icon {
+      display: none;
+    }
+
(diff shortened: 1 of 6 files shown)


tests/pages/_includes/widgets/navigation/application-launcher.html
```diff
@@ -0,0 +1,414 @@
...
+      <li class="applauncher-pf applauncher-pf-block-list dropdown dropdown-kebab-pf">
+        <a class="dropdown-toggle drawer-pf-trigger-icon" data-toggle="dropdown" href="#">
+          <span class="fa fa-th applauncher-pf-icon"></span>
+          <span class="applauncher-pf-title">
+            Application Launcher
+            <span class="caret" aria-hidden="true"></span>
+          </span>
+        </a>
+        <ul class="dropdown-menu dropdown-menu-right" role="menu">
+          <li class="applauncher-pf-entry" role="menuitem">
```
andresgalante: Can we change `applauncher-pf-entry` to `applauncher-pf-item` to be more consistent with other list item class names on bootstrap?

cdcabrera: Good catch, sure thing

tests/pages/_includes/widgets/navigation/application-launcher.html
```diff
@@ -0,0 +1,414 @@
...
+            <a class="applauncher-pf-link" href="#" title="Launch the Function User Interface">
+              <i class="applauncher-pf-link-icon fa fa-2x pficon-storage-domain" aria-hidden="true"></i>
```
andresgalante: When we are using a pficon we should not call `fa`. 

It's `pficon pficon-storage-domain` instead of `fa pficon-storage-domain`

also, I think that since fa-2x is a modifier for font awesome, we can include that font size change on the CSS targeting `applauncher-pf-link-icon`. 
What do you think?

cdcabrera: Going for something like?
```css
.applauncher-pf-link-icon {
...
font-size: [the 2x goes here]
...
}
```

andresgalante: yes!

cdcabrera: sure, can do on the pficon and the font-sizing

src/less/application-launcher.less
```diff
@@ -0,0 +1,188 @@
...
+
+      .box-shadow(0 0 2px 0 @applauncher-pf-menu-link-shadow);
+    }
+  }
+
+  .applauncher-pf-link-icon,
+  .applauncher-pf-link-title {
+    display: inline-block;
+    width: auto;
+    vertical-align: middle;
```
andresgalante, line 49: Is this doing anything?

andresgalante:
@cdcabrera This is not an easy pattern to work. I love how you resolve the mobile view, great work!

I would have use less variables, but after talking to @srambach I am ok with them.

Thanks a lot David and sorry for taking me so long to review it. Let's make those small changes and merge it, please remember to also update the 3x PR

src/less/application-launcher.less
cdcabrera: Indeed, combined with the class ```applauncher-pf-link-icon``` above it, it applies to the list view of icon and title side by side

src/less/application-launcher.less
```diff
@@ -0,0 +1,192 @@
+//
+// Application Launcher
+// --------------------------------------------------
+
+.applauncher-pf {
+
+  display: inline-block;
+  overflow: visible;
+
+  .applauncher-pf-title {
```
srambach, line 10: Do we need to nest the .applauncher-pf-* classes inside .applauncher-pf? One extra level of specificity we can do away with, right?

tests/pages/_includes/widgets/navigation/application-launcher.html
```diff
@@ -0,0 +1,414 @@
...
+    </a>
+  </div>
+  <div class="collapse navbar-collapse navbar-collapse-1">
+    <ul class="nav navbar-nav navbar-utility">
+
+      <li class="applauncher-pf applauncher-pf-block-list dropdown dropdown-kebab-pf">
+        <a class="dropdown-toggle drawer-pf-trigger-icon" data-toggle="dropdown" href="#">
+          <span class="fa fa-th applauncher-pf-icon"></span>
+          <span class="applauncher-pf-title">
+            Application Launcher
```
jgiardino: This text would be nice to include for screenreaders (via `aria-label` attribute on the `.applauncher-pf-icon` element) as well as in a tooltip widget on the `<a>` element. Currently it only appears to be visible on smaller screen sizes.

tests/pages/_includes/widgets/navigation/application-launcher.html
```diff
@@ -0,0 +1,414 @@
...
+            <span class="caret" aria-hidden="true"></span>
+          </span>
+        </a>
+        <ul class="dropdown-menu dropdown-menu-right" role="menu">
+          <li class="applauncher-pf-item" role="menuitem">
+            <a class="applauncher-pf-link" href="#" title="Launch the Function User Interface">
```
jgiardino: Should we use the [tooltip widget](http://www.patternfly.org/pattern-library/widgets/#tooltip) instead of title text? The tooltip widget is nice in that it's keyboard accessible. But it's not screenreader accessible, and in this case, I'm not sure what the right solution is (maybe an `aria-describedby` attribute that references the text inside a span that's only visible to screenreaders?).

cdcabrera: Tooltip widget was discussed earlier on the 3x version, see https://github.com/patternfly/patternfly/pull/632

src/less/application-launcher.less
```diff
@@ -0,0 +1,192 @@
...
+    display: none;
+  }
+
+  .dropdown-toggle {
+    &.disabled {
+      cursor: not-allowed;
```
jgiardino, line 16: Is this needed here? I think `cursor: not-allowed` is already provided by bootstrap for `.disabled`

src/less/application-launcher.less
cdcabrera: Indeed, first round it wasn't nested, however it's used as a mixin around [Line 173](https://github.com/cdcabrera/patternfly/blob/Feature-AppLauncher-4x/src/less/application-launcher.less#L173)

Up for recommendations though...

jgiardino:
This looks really nice, David! I just included a few unexciting comments :-)

tests/pages/_includes/widgets/navigation/application-launcher.html
cdcabrera: Good call, the Angular PF version makes use of the sr-only setup... I can make sure this version is also updated

srambach:
In the narrow screen mode, as I tab through the applications, the current one's text is a dark gray on dark gray.

src/less/application-launcher.less
cdcabrera: Can double check on this one. There were a few instances of cascade overrides needing to be more aggressive

cdcabrera:
There was a modification earlier today regarding the classes for smaller screen, this may be related, will take a look

tests/pages/_includes/widgets/navigation/application-launcher.html
jgiardino: Ah, ok. Thanks for the cross-reference! So should the `title` attribute just be removed altogether?

cdcabrera: @jennyhaines @cshinn your thoughts?

jgiardino:
That gray on gray effect is happening on the Brian Johnson menu, too, and outside of this PR (for example: https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/horizontal-navigation.html#). This might be a separate issue.

tests/pages/_includes/widgets/navigation/application-launcher.html
cdcabrera: For the AngularPF component, this is treated as a further description since there may be limited display area

andresgalante: Hi @cdcabrera we don't really care about angular at this point. If this attribute is not necessary then let's remove it.

@LHinson Can we please change the design docs to reflect this?

src/less/application-launcher.less
srambach: OK, I see what you are doing there. I don't have any better answer. I guess it could call all the individual classes as mixins within the media query. That reduces specificity but is messy. Thanks for explaining. :-)

tests/pages/_includes/widgets/navigation/application-launcher.html
jennyhaines: @jgiardino @cdcabrera @andresgalante - I agree with Andres in not needing the tooltip here.

@LHinson - I can take out the tooltip from the images on the PatternFly design during next PF sprint.

dtaylor113:
> we don't really care about angular at this point.

@andresgalante Why do we not care about angular?  OpenShift is using angular-patternfly 3x and is specifically waiting for this component.

src/less/application-launcher.less
cdcabrera: Investigated, override for a cascade around ```navbar-pf```  [notifications-drawer.less line 162](https://github.com/patternfly/patternfly/blob/branch-4.0-dev/src/less/notifications-drawer.less#L162)

andresgalante:
@dtaylor113 we don't care about angular **at this point**.

At PatternFly core we build the best mark up and CSS possible. We do not add extra markup or useless structures just for the sake of any JS framework implementation.

dtaylor113:
@andresgalante, ah ok, I get your meaning now -thanks :-)

PR #637, opened 2017-05-02
Update to jQuery 3

dabeng:
## Description
Upgrade PatternFly and all patternfly dependencies to use jquery 3 APIs .

## Dependencies List on gemnasium
https://gemnasium.com/github.com/dabeng/patternfly

@bleathem @priley86 , please review :)


bower.json
```diff
@@ -22,20 +22,20 @@
...
     "c3": "~0.4.11",
     "d3": "~3.5.17",
-    "datatables": "~1.10.12",
+    "datatables": "~1.10.15",
     "datatables-colreorder": "~1.3.2",
     "datatables-colvis": "~1.1.2",
-    "font-awesome": "~4.6.3",
+    "font-awesome": "~4.7.0",
     "google-code-prettify": "~1.0.5",
     "jquery": "~2.1.4",
```
bleathem: The jQuery dependency needs to be updated to jQuery 3 here.

package.json
```diff
@@ -5,7 +5,7 @@
   "license": "Apache-2.0",
   "dependencies": {
     "bootstrap": "~3.3.7",
-    "font-awesome": "~4.6.3",
+    "font-awesome": "^4.7.0",
     "jquery": "~2.1.4"
```
bleathem: The jQuery dependency needs to be updated to jQuery 3 here too.

bleathem (changes requested):

bleathem (approved):

bleathem:
This PR looks good to me.  @dabeng I pushed some additional commits addressing the jQuery dependency itself, please review.

bleathem:
This PR looks good to me.  @dabeng please review the commits I added to your PR.

package.json
```diff
@@ -6,7 +6,7 @@
   "dependencies": {
     "bootstrap": "~3.3.7",
     "font-awesome": "^4.7.0",
-    "jquery": "~2.1.4"
+    "jquery": "~3.2.1"
```
dabeng: @bleathem , I didn't change this line in my commit because I'm not sure the complete source code of current patternfly can work well with jquery 3. I think it will need a considerable efforts to test the compatibility.

jeff-phillips-18:
@dabeng Could you please explain "we need to migrate Patternfly to jQuery 3 asap if this RP merged." ?

Gruntfile.js
```diff
@@ -48,20 +48,20 @@ module.exports = function (grunt) {
...
+          'src/js/patternfly-functions-sidebar.js',
+          'src/js/patternfly-functions-popovers.js',
+          'src/js/patternfly-functions-data-tables.js',
+          'src/js/patternfly-functions-navigation.js',
+          'src/js/patternfly-functions-count-chars.js',
+          'src/js/patternfly-functions-colors.js',
+          'src/js/patternfly-functions-charts.js',
+          'src/js/patternfly-functions-fixed-heights.js',
+          'src/js/patternfly-functions-tree-grid.js',
+          'src/js/patternfly-functions-vertical-nav.js'],
```
jeff-phillips-18, line 64: Is it worth modifying this file due to indentation?

bower.json
```diff
@@ -22,24 +22,21 @@
   "dependencies": {
     "bootstrap": "~3.3.7",
     "bootstrap-datepicker": "~1.6.4",
-    "bootstrap-select": "~1.10.0",
-    "bootstrap-switch": "~3.3.2",
+    "bootstrap-select": "~1.12.2",
+    "bootstrap-switch": "3.3.3",
```
jeff-phillips-18: Are we purposefully making this strictly 3.3.3? Is 3.3.3 required if migrating to JQuery 3?

jeff-phillips-18:
Can we use constants for the CDN includes so that when we update versions we don't need to modify so many test pages?

bower.json
bleathem: As per the error introduced in this issue: https://github.com/Bttstrp/bootstrap-switch/issues/612, using the latest release brings in a jQuery 1.x dependency.

bleathem:
> Can we use constants for the CDN includes so that when we update versions we don't need to modify so many test pages?

This is a great idea, let's track that effort with a new story/PR rather than delaying this one.

bower.json
jeff-phillips-18: OK, agree with the version bump, but being stringent to an exact version causes dependency conflicts for applications.

bleathem: Yeah, it's unfortunate.  Can you suggest an alternate solution?  I could specify a range as in:
`"bootstrap-switch": "<=3.3.3"`

jeff-phillips-18: OK, gotcha. no good solution.

jeff-phillips-18 (approved):

bleathem:
To test this PR I ran through all test pages with the https://jquery.com/upgrade-guide/3.0/#jquery-migrate-plugin active.  The only errors I came across stemmed from application code, and I resolved them.
(cc: @dabeng)

jeff-phillips-18:
@bleathem Could you add a link to the story/issue to use constants for CDN includes?

bleathem:
@jeff-phillips-18:
> Could you add a link to the story/issue to use constants for CDN includes?

https://patternfly.atlassian.net/browse/PTNFLY-2373

Gruntfile.js
dabeng: If don't modify indentation, grunt build will get error
`error  Expected indentation of 10 spaces but found 14  indent`

bleathem:
@dabeng can you please indicate in this PR whether you are ok with the commits I amended?

PR #645, opened 2017-05-14
Remove @supports flex for ie11

priley86:
## Description
* removes @supports flex for wizard-pf-steps-indicator and list-view-pf-additional-info-item
* eases duplicate rule restriction based on values [only](https://stylelint.io/user-guide/rules/declaration-block-no-duplicate-properties/#ignore-consecutive-duplicates-with-different-values)

Include related PRs or Issues
https://github.com/openshift/origin-web-catalog/pull/223#issuecomment-301068315

## Todos
- [x] cross browser test
- [x] Is it reponsive?

cc @andresgalante @rhamilto

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -27,7 +27,9 @@
     "value-no-vendor-prefix": true,
     "shorthand-property-no-redundant-values": true,
     "property-case": "lower",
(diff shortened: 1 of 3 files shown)


andresgalante:
👍  I am ok with this

PR #649, opened 2017-05-18
Moves flex on list content to an additional class

srambach:
## Description
This fixes a backwards compatibility issue with the list-pf-contents being flexed by default.

## Todos
- [x] cross browser test - tested on Chrome, Firefox, Safari *Needs IE
- [x] Is it responsive?

Outline the steps to test or reproduce the PR here.

```sh
git pull --prune
git checkout <feature_branch>
bundle; script/server
```

Also, you may want to please any @mentions you'd like to review.

### DIFF
--- a/src/less/list-pf.less
+++ b/src/less/list-pf.less
@@ -64,7 +64,7 @@
 }
 
 // add this class to manage flexed contents in the list item content
-.list-pf-flex-content {
+.list-pf-content-flex {
   align-items: flex-start;
   display:flex;
   flex-grow: 1;
--- a/src/less/list-pf.less
+++ b/src/less/list-pf.less
@@ -59,6 +59,12 @@
 }
 
 .list-pf-content {
+  flex-grow: 1;
+  min-width: 0;
+}
+
+// add this class to manage flexed contents in the list item content
+.list-pf-flex-content {
   align-items: flex-start;
   display:flex;
   flex-grow: 1;
--- a/tests/pages/list-pf.html
+++ b/tests/pages/list-pf.html
@@ -164,7 +164,7 @@ resource: true
       <div class="list-pf-select">
         <input type="checkbox">
       </div>
-      <div class="list-pf-flex-content">
+      <div class="list-pf-content list-pf-content-flex">
         <div class="list-pf-left">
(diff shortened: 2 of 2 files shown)


andresgalante:
@bleathem be aware that we are creating a new bug for any users using pf-list with the example. They will have to add the new `list-pf-content-flex` class. Make sure you mention this on the release notes.

andresgalante:
@srambach I've just test it and it works great, thanks a lot for fixing it.

@rhamilto @jwforres Sorry for introducing this problem, does this PR fix it on your side?

src/less/list-pf.less
```diff
@@ -59,6 +59,12 @@
 }
 
 .list-pf-content {
+  flex-grow: 1;
```
rhamilto, line 62: This is duplicated in .list-pf-content-flex.  Shouldn't it only live there?

src/less/list-pf.less
```diff
@@ -59,6 +59,12 @@
...
+  min-width: 0;
```
rhamilto, line 63: This wasn't part of the original, so I think it's best not to include it.

rhamilto (commented):
A couple of questions.

bleathem:
@andresgalante can you please update the PR description with details of this enhancement (it's not a _bug_ if we are doing it on purpose), and what steps the user has to take to make this enhancement work with their application?  This way it will get included in the release notes.

src/less/list-pf.less
andresgalante: @srambach I can't see to make any difference if we leave that there or not on the empty list. Can we move it?

srambach: Those two styles apply to the behavior of the container itself (.list-pf-content) not the contents of that container. There's no visible difference in the simple examples, but I think once you fill that container with things, it will be needed for the container to flex properly. I could be wrong... @rhamilto have you tried it with in what you are putting into the list content? Does it still behave at different screen widths?

rhamilto: My thoughts are this makes list-pf-content opinionated.  What we originally requested (and was provided) was something agnostic.  This doesn't make a difference in our context, but it could for others?

bleathem:
This PR is obsolete with #652

PR #650, opened 2017-05-19
Merge branch 4.0 dev changes into master

bleathem:
The changes in the 4.0 dev branch are not significant enough to warrant a major version change.  This PR brings those changes into the master branch to be made available with the next PatternFly 3 minor release.


dtaylor113 (approved):

Gruntfile.js
```diff
@@ -45,7 +47,21 @@ module.exports = function (grunt) {
     },
     concat: {
       js: {
-        src: ['src/js/patternfly-settings.js', 'src/js/patternfly-functions.js'],
+        src: ['src/js/patternfly-settings.js',
+          'src/js/patternfly-settings-colors.js',
+          'src/js/patternfly-settings-charts.js',
```
jeff-phillips-18: I'm a bit concerned that this change now requires the inclusion of patternfly-settings.js patternfly-settings-colors.js and patterfly-settings-colors.js whereas the contents of all 3 were in patternfly-settings.js. This will break the usage if the application does not update to include the two new files as well. This change is not backwards compatible.

src/js/patternfly-functions.js
```diff
@@ -1,1224 +1,7 @@
...
-        $chevron.removeClass('fa-angle-down');
-      }
-    }
-
-    init(this);
-
-    return this;
-  };
-}(jQuery));
+}(jQuery));
```
jeff-phillips-18: Breaking this out into several files will break anyone that used to simply include this file. Not backwards compatible.

src/js/patternfly-settings.js
```diff
@@ -516,5 +12,5 @@
   };
 
   window.patternfly = patternfly;
-})(window);
 
+})(window);
```
jeff-phillips-18: Breaking this into separate files will require inclusion of multiple file rather than just patternfly-settings.js. This is not backwards compatible.

jeff-phillips-18 (changes requested):
I think we should concatenate the functions files into patternfly-functions.js and settings files into patternfly-settings.js in order to keep backwards compatibility.

bleathem:
Good point @jeff-phillips-18. I've re-created the patternfly-settings and patternfly-functions files at build time.  Does this address your concerns?

jeff-phillips-18 (approved):
👍

dlabrecq (approved):
LGTM

PR #651, opened 2017-05-20
Fix #634: Header link inside panel-info is almost unreadable

### LINKED ISSUE #634: Header link inside panel-info is almost unreadable
Follow-up of #511
JSFiddle: https://jsfiddle.net/8jp7ep25/1/
The other type of panels, having a much lighter background, don't show this problem.

nireno:
This adds a new variable that modifies @brand-info to try to find a more
suitable background color for info panels in a panel-group.

The style of other panel types had to be supplemented to ensure that
"higher priority" panel borders take precedence.

#634

### DIFF
--- a/src/less/panels.less
+++ b/src/less/panels.less
@@ -28,12 +28,24 @@
     }
   }
 
+  .panel-info {
+    border-color: @panel-info-border;
+    .panel-heading {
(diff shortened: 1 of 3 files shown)


andresgalante:
@nireno thanks a lot for fixing it 👍

PR #655, opened 2017-05-31
Fix app-launcher alignment with no icon

### LINKED ISSUE #648: Application Launcher's List View display issues
In list view:

image

- Please add some padding between the icon and label.
```
.applauncher-pf-link-icon {
      @media (min-width: @screen-sm-min) {
        padding-right: 10px;
      }
    }
```
- Please align all labels regardless of size/width of icons, or if there some apps which do not have icons defined.  (Image below has padding-right applied)
image

Similar to using 'fa-fw'  http://fontawesome.io/examples/  "Fixed Width Icons".  https://github.com/FortAwesome/Font-Awesome/blob/v4.7.0/less/fixed-width.less

cshinn:
## Description
Adds a simulated fa-fw (fixed width) style to the icons in the app launcher to ensure that text lines up properly. Adds an example with a placeholder element (with no icon specified) to correctly align an item with no icon. Closes #648.

screen shot 2017-05-31 at 1 52 24 pm
screen shot 2017-05-31 at 1 52 17 pm
screen shot 2017-05-31 at 1 52 32 pm
 

## Todos
- [x] cross browser test
- [X] Are you sure it works on IE9?
- [X] Is it reponsive?

### DIFF
--- a/src/less/variables.less
+++ b/src/less/variables.less
@@ -23,6 +23,7 @@
 @applauncher-pf-menu-link-background-color-hover:                   @color-pf-black-150;
 @applauncher-pf-menu-link-shadow:                                   @color-pf-black-300;
 @applauncher-pf-menu-link-icon-font-size:                           2em;
+@applauncher-pf-menu-link-icon-width:                               ((9 * @applauncher-pf-menu-link-icon-font-size) / 14); //Numbers to imitate fa-fw class
 @bootstrap-switch-handle-default-bg-color:                          @color-pf-black-100;
 @bootstrap-treeview-highlight-color:                                @color-pf-blue-300;
 @btn-default-bg-img-start:                                          @color-pf-black-100;
--- a/src/less/application-launcher.less
+++ b/src/less/application-launcher.less
@@ -40,12 +40,17 @@
 
   .applauncher-pf-link-icon {
     font-size: @applauncher-pf-menu-link-icon-font-size;
+    text-align: center;
+    width: @applauncher-pf-menu-link-icon-width;
+    min-height: @applauncher-pf-menu-link-icon-width;
+    @media (min-width: @screen-sm-min) {
+      padding-right: @applauncher-pf-icon-padding-right;
+    }
   }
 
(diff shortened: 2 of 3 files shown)


dtaylor113:
Hi, LGTM, although I see `@applauncher-pf-icon-padding-right: 4px;`  I would still like to see more space between the icon and label in list view, perhaps padding-right: 10px?

dtaylor113 (approved):

src/less/variables.less
```diff
@@ -23,6 +23,7 @@
 @applauncher-pf-menu-link-background-color-hover:                   @color-pf-black-150;
 @applauncher-pf-menu-link-shadow:                                   @color-pf-black-300;
 @applauncher-pf-menu-link-icon-font-size:                           2em;
+@applauncher-pf-menu-link-icon-width:                               ((9 * @applauncher-pf-menu-link-icon-font-size) / 14); //Numbers to imitate fa-fw class
```
andresgalante, line 26: We've talked this before with @srambach, but here goes again: Do we actually need everything to be a variable? 😄

andresgalante:
Hi, I'll be out tomorrow and Friday, I am ok merging this PR, if you guys are ok with it too please go ahead and merge it, thanks!

@cshinn great work, thanks a lot 😄

src/less/variables.less
srambach: I'm still a fan of the variables. :-) In particular here, I suspect the icon size could have to deviate from the standard so it's good to have it as a variable. Even better would be to know what the magic 9 and 14 are :-)

cshinn: I guess I didn't make my comment clear enough. Those numbers come from the font awesome fixed width class, and I guess that's the magic font size/width ratio to get things to display correctly

srambach: I'm sorry, no, I understand it's matching fa. It was a poorly-worded offhand comment about fa using magic numbers. :-) I'm good with this as is, nicely done.

cdcabrera (approved):

PR #660, opened 2017-06-02
feat(canvas): Add Canvas and Canvas Containers

priley86:
## WIP: Canvas and Canvas Container components
Starts the HTML/CSS contribution to PF Core for the [Patternfly Canvas View](http://www.patternfly.org/pattern-library/content-views/canvas-view/)

More work is needed to be done:
* review inline canvas styles and determine if any new classes can be added 
* test selected and hover states for canvas nodes
* add canvas toolbox example
* validate the CSS works in Angular-PF and PF-React. This LESS/CSS most closely mimics the current CSS in Angular-PF.
* include any subset of behaviors we'd like to have in this repo
* review [canvas-containers UX](https://github.com/patternfly/patternfly-design/pull/284) in patternfly-design 

## Todos
- [ ] cross browser test
- [ ] Is it responsive? 
I will need some feedback on any potential responsive states we'd like to support

Canvas:
https://rawgit.com/priley86/patternfly/canvas-view-dist/dist/tests/canvas.html

Canvas Containers:
https://rawgit.com/priley86/patternfly/canvas-view-dist/dist/tests/canvas-containers.html

```sh
git pull --prune
git checkout <feature_branch>
bundle; script/server
```

cc @andresgalante @dtaylor113


bleathem:
To what degree is this PR aligned with the angular-patternfly canvas?
http://www.patternfly.org/angular-patternfly/#/api/patternfly.canvas.directive:pfCanvas

dlabrecq (commented):
I tried the rawgit demo on Chrome and Firefox, but nothing seems to be working. I cannot drag, the add item button does not function, zoom does not work, etc. Unfortunately, I don't see any errors in the console.

Is the intention to just add styles to Patternfly or should this test page be functional?

dtaylor113:
> nothing seems to be working. I cannot drag, the add item button does not function, zoom does not work, etc.

I think this is just static html and less/css contribution.  I don't see any javascript files, so I don't think anything dynamic is going to happen.

dlabrecq:
Gotcha. Thanks!

src/img/aws.svg
```diff
@@ -0,0 +1,91 @@
+<?xml version="1.0" encoding="utf-8"?>
```
dlabrecq, line 1: Do we want to add 3rd party logos to the Patternfly repo? We already have the OpenShift logo, but that's Red Hat.

dlabrecq (dismissed):
I noticed that patternfly.org has a placeholder for a canvas example. If we plan to add this, the example will need to be located under tests/pages/_includes to be imported properly.

priley86:
@dlabrecq @dtaylor113 sorry for the late response...this somehow passed my github filter while on vacation ;(

I would very much like to share CSS where possible across the repos (definitely the idea here). No JS behaviors implemented on purpose...

@dtaylor113 can you confirm if this would work in angular-pf (if sourced from core in this way)?

@dlabrecq i'm happy to swap any images we'd like. also fine w/ moving to the _includes folder. I'll update this PR to account for that soon..

dtaylor113:
> @dtaylor113 can you confirm if this would work in angular-pf (if sourced from core in this way)?

Do you mean if the css was in pf core, would it work for angular-patternfly?   Did you change style names?

I imagine we'd have to do a little integration changes in angular-patternfly to use the new canvas pf core html/css.

priley86:
@dtaylor113 No class names were changed (and I attempted to mimic angular-pf as close as possible). However, I was pulling a subset of the angular-pf css at the time (so would like to ensure that these are *all* of the classes associated w/ canvas in angular-pf). Would also like to ensure that I mapped the colors/less variables correctly in `canvas.less`. https://github.com/patternfly/patternfly/pull/660/files#diff-a8e7e74153bbc764e423215a35bf592b

Now would also be a good time to add any additional CSS classes you'd like (i still see that some css is inline, however this appears to be desired when calculating widths/heights etc.). I will try to mimic this structure as closely as possible in pf-react once we are thru...

priley86:
@dtaylor113 It seems that at the time i did this, `canvas.less` was not its own file (and the css was in one file before `angular-patternfly.css`). Now that it is split, i'd be happy to adopt what's in [here](https://github.com/patternfly/angular-patternfly/blob/master/src/canvas-view/canvas/canvas.less) .

I'll update the PR to reflect this soon...

dtaylor113:
@priley86, ok, sounds good.  In general, I'd say go with what you have here, make it as close as possible to a-pf, then we can do a follow up PR in a-pf to replace the current canvas less/css to use what's in core.
Thanks!

priley86:
@dtaylor113 i've cross checked the less w/ a-pf. Everything should be the same now except the following:

* `.right-aligned-controls`: removed vertical-align:middle as it does not apply to block level elements
* `.canvas-editor-toolbox-container`: now uses autoprefixer here to prefix `::placeholder` styles
* `.read-only` and `.canvas-in-connection-mode`: removed `!important`. these shouldn't be necessary
`.canvas` : added a default `background-image` (canvas-dot-grid.png), background-repeat and background-size. This is for display purposes in this repo. It should easy enough to override downstream in a-pf.

@dlabrecq I've moved the templates to `pages\_includes\widgets\canvas-view`. Let me know if this works...

priley86:
@dtaylor113 @dlabrecq anything else here?

src/img/aws.svg
andresgalante: We can't add any logo, they are copyrighted

src/less/canvas-editor.less
```diff
@@ -0,0 +1,196 @@
+.canvas-editor-container {
```
andresgalante, line 1: Every patternfly related class has a `-pf` postfix with this format:
`block-pf-element-modifier`

src/less/canvas-editor.less
```diff
@@ -0,0 +1,196 @@
...
+  background-color: @color-pf-white;
+  height: 100%;
```
andresgalante, line 3: 100% of what?

andresgalante:
We'll have to rework the CSS. @LHinson this is not a simple task, you would like to create a story for it?

src/less/canvas-editor.less
priley86: +1

jeff-phillips-18 (changes requested):
Just marking as 'request changes' so as not to get merged until the CSS is reworked.

LHinson:
After talking with @priley86, we are going to close this one. If/when work picks back up, a new PR will be generated. Thanks everyone.

PR #675, opened 2017-06-30
Implement Notification Drawer Updates

jeff-phillips-18:
Implementation of
https://github.com/patternfly/patternfly-design/pull/297

## Description
This PR add implementations for the notification drawer updates:
  - Close button for the notifications drawer
  - Mark All Read and Clear All in one horizontal area
  - Hide Mark All Read button when there are no unread notifications
  - Use the empty state pattern when there are no notifications

## View and test at:
  - http://rawgit.com/jeff-phillips-18/patternfly/notifications-dist/dist/tests/notification-drawer-horizontal-nav.html
  - http://rawgit.com/jeff-phillips-18/patternfly/notifications-dist/dist/tests/notification-drawer-vertical-nav.html

@beanh66 @serenamarie125 @jennyhaines

### DIFF
--- a/tests/pages/_includes/widgets/notification-drawer.html
+++ b/tests/pages/_includes/widgets/notification-drawer.html
@@ -1,7 +1,7 @@
 <div class="drawer-pf hide drawer-pf-notifications-non-clickable">
   <div class="drawer-pf-title">
-    <a class="drawer-pf-toggle-expand"></a>
-    <a  class="drawer-pf-close"></a>
+    <a class="drawer-pf-toggle-expand fa fa-angle-double-left"></a>
+    <a  class="drawer-pf-close pficon pficon-close"></a>
     <h3 class="text-center">Notifications Drawer</h3>
   </div>
   <div class="panel-group" id="notification-drawer-accordion">
--- a/tests/pages/_includes/widgets/notification-drawer.html
+++ b/tests/pages/_includes/widgets/notification-drawer.html
@@ -1,6 +1,7 @@
 <div class="drawer-pf hide drawer-pf-notifications-non-clickable">
   <div class="drawer-pf-title">
     <a class="drawer-pf-toggle-expand"></a>
+    <a  class="drawer-pf-close"></a>
     <h3 class="text-center">Notifications Drawer</h3>
   </div>
   <div class="panel-group" id="notification-drawer-accordion">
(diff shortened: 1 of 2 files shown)


beanh66:
Thanks @jeff-phillips-18 the badge, mark as read, clear, and empty state are working great. A few things I noticed were (1) in the expanded state, the close x and the title are not in the correct places. The drawer title should be centered and the close should still remain in the far right corner. (2) When clicking on an individual cell that is "new" (bold font) it should become marked as read and unbolded.

jeff-phillips-18:
@beanh66 I will take a look at the close button issue. The Mark read was not implemented before this PR. I can add the JS to make it so.

jeff-phillips-18:
@beanh66 Updated to address both issues.

jennyhaines:
The implementation looks great. It seems that @beanh66 's comments were addressed, as well! bitmoji

beanh66:
@jeff-phillips-18 thanks the centering looks great now. Just wondering about the cursor change for the clicking to mark as read. Changing to the text editor cursor makes it seem like the rows are not clickable. Should that stay as a pointer or change to the click pointer? It is a weird case since it doesn't actually link to anything, but IMO it should still look clickable. Separate nit, can the expand arrow change to the blue on hover since we do that for expanding list view items as well?

jeff-phillips-18:
@beanh66 Changed to the default pointer cursor and added the link color on hover of the drawer expander.

beanh66:
Thanks, LGTM!

jeff-phillips-18:
@beanh66 @andresgalante @bleathem  Updated removing changes for the unread notifications indicator which were done in #678. Can you please please review this please :)

tests/pages/_includes/widgets/notification-drawer.html
```diff
@@ -1,6 +1,7 @@
 <div class="drawer-pf hide drawer-pf-notifications-non-clickable">
   <div class="drawer-pf-title">
     <a class="drawer-pf-toggle-expand"></a>
+    <a  class="drawer-pf-close"></a>
```
andresgalante: Is there a reason why we are using generated icon with css instead of adding it on the html?

andresgalante:
@jeff-phillips-18 Other than the generated icons I think everything is ok. Thanks a lot Jeff!

tests/pages/_includes/widgets/notification-drawer.html
jeff-phillips-18: Using pficon-close sets the line height to 1, and we would need to re-override it back to 1.66667 to align correctly. I can make that change if desired.

matthewcarleton:
This looks good to me, nice work Jeff!

tests/pages/_includes/widgets/notification-drawer.html
andresgalante: That sucks :( Since the icons on the masthead are not generated I think we should do the same here. I'd go with a real icon on the html.

jeff-phillips-18: Done. Thanks @andresgalante

andresgalante:
👍  Thanks a lot @jeff-phillips-18 I am ok to merge this.

@bleathem when you can please check the JS

bleathem (approved):

bleathem:
The js looks fine to me.  Thanks for the PR @jeff-phillips-18!

PR #677, opened 2017-07-04
Exposed the local grunt command via an npm run script

bleathem:
Exposed the local grunt command via an npm run script.  This allows one to invoke arbitrary grunt tasks without a globally installed grunt.  eg.:
`npm run grunt -- build server`


cdcabrera:
lgtm

PR #678, opened 2017-07-05
Bell icon in navigation

matthewcarleton:
## Description
adds bell icon to vertical and horizontal navigation, with circle indicator. Removes flag icon.

In order to show the empty badge this requires a space, otherwise add a value 

Here is the url for the horizontal navigation ( you'll notice the 1px border to increase visibility at a smaller size :) ):
https://rawgit.com/matthewcarleton/patternfly/notification-drawer-dist/dist/tests/notification-drawer-horizontal-nav.html

Here is the url for the vertical navigation:
https://rawgit.com/matthewcarleton/patternfly/notification-drawer-dist/dist/tests/notification-drawer-vertical-nav.html

## Todos
- [x] cross browser test
- [x] Is it reponsive?

@andresgalante @srambach @jgiardino Can you please please review this please :)

### DIFF
--- a/src/less/navbar.less
+++ b/src/less/navbar.less
@@ -55,6 +55,20 @@
         }
       }
     }
+ .badge {
+    background-color: @notification-badge-color;
+    border-radius: 20px;
+    border:1px solid @navbar-pf-bg-color; //improves visibilty at smaller size
+    color: @navbar-pf-vertical-active-color;
+    cursor: pointer;
+    font-size: (@font-size-base - 3);
+    font-weight: 700;
+    padding: 2px 4px;
+    margin-top:-12px;
+    margin-left:-9px;
+    min-height: 10px;
+}
+
     @media (max-width: @grid-float-breakpoint-max) {
       .active .dropdown-menu,
       .active .navbar-persistent,
--- a/src/less/navbar-vertical.less
+++ b/src/less/navbar-vertical.less
@@ -46,7 +46,7 @@
           color: @navbar-pf-vertical-active-color;
         }
       }
-
+      
       .badge {
         background-color: @notification-badge-color;
         border-radius: 20px;
@@ -54,9 +54,11 @@
         cursor: pointer;
         font-size: (@font-size-base - 3);
         font-weight: 700;
-        margin: 0 0 -11px -12px;
-        min-width: 0;
+        margin-top: -15px;
+        margin-left: -12px;
         padding: 2px 4px;
+        min-width: 10px;
+        min-height: 10px;
       }
 
       .caret,
--- a/src/less/variables.less
+++ b/src/less/variables.less
@@ -461,7 +461,7 @@
 @dropdown-link-focus-bg:                                            @link-color;
 @line-height-computed:                                              floor((@font-size-base * @line-height-base));
 @nav-pf-vertical-icon-size:                                         (@font-size-base + 8);
-@notification-badge-color:                                          @brand-danger;
+@notification-badge-color:                                          @color-pf-blue-400;
 @progress-sm:                                                       (@line-height-computed - 6);
 @progress-xs:                                                       (@line-height-computed - 14);
 @syntax-1:                                                          @btn-danger-bg;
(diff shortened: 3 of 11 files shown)


src/less/navbar.less
```diff
@@ -55,6 +55,20 @@
         }
       }
     }
+ .badge {
+    background-color: @notification-badge-color;
+    border-radius: 20px;
+    border:1px solid @navbar-pf-bg-color; //improves visibilty at smaller size
+    color: @navbar-pf-vertical-active-color;
```
andresgalante, line 62: Using a vertical-nav related variables for the horizotal nav seems strange. What color is this, white?

andresgalante (commented):
Code looks ok, great job! 🎉 
The blue for the ball seems off. @mcarrano @kybaker can you please confirm that's the correct blue please.

src/less/navbar.less
matthewcarleton: ah yes that isn't right. Missed that when I copied it over from the vertical implementation. It is white yes.

mcarrano:
The badge color looks OK to me, @andresgalante But I will defer to @kybaker on that. The only other thing is that the representation of other items in the masthead is inconsistent between the two examples. Can we make these both consistent with http://www.patternfly.org/pattern-library/application-framework/masthead/#/design ? The vertical example is close but should add user name.

andresgalante:
@mcarrano maybe @matthewcarleton needs more guidance, Can you be more specific about exactly you'd like to see?

matthewcarleton:
@mcarrano @andresgalante Ok I can add the username for the vertical example.

mcarrano:
Yes, the Veritcal Nav example should include the user name. In the Horizontal example, the label "Status" should not be there (text labels in the masthead are not defined). I would replace that with the help menu as it is in the vertical example.  So when were done, both examples should be identical except for the sizing.

matthewcarleton:
@mcarrano ok, makes sense I will do that.

matthewcarleton:
I've updated the pages to reflect all the same icons. For the Nav page I've removed the status and added the help dropdown. Also changed the general navigation pages to not include the notification icon/drawer.

mcarrano (approved):
Looks great @matthewcarleton !  I'm good to merge.

andresgalante:
@kybaker the "help" icon looks fuzzy on a non retina screen, is that ok?
screen shot 2017-07-07 at 4 10 03 pm

src/less/navbar.less
jeff-phillips-18: Why the border on the badge for horizontal nav but not for vertical nav? I see it says for visibility at smaller sizes but I think it may help for the vertical navbar size as well.

jeff-phillips-18 (approved):
LGTM, one question below but I'm OK either way.

matthewcarleton:
@jeff-phillips-18 I did implement it on the larger version (vertical nav) but didn't find it necessary.

beanh66:
@matthewcarleton I was wondering why the black border as well. I don't think @jeff-phillips-18 had the border at all in his implementations. @jennyhaines thoughts on this from the visuals side?

jennyhaines:
@matthewcarleton @beanh66 - I think the 1px border looks nice when it matches with the background color. This would mean that by default, the border color is #030303 ... on hover, the border color is #232323 ... and while active, the border color should be #363636 . If this is too much, I would think that scrapping the border would be fine, though it does look nicer with the border, in my opinion. It would be good to have consistency in the border treatment between both horizontal and vertical implementations.

andresgalante:
Hi @jennyhaines , #363636 is not part of our color pallet:
https://www.patternfly.org/styles/color-palette/

We should stick to colors from the color pallet so we can use less variables and keep consistency.

srambach:
FWIW, I was recently trying to match colors in the masthead and navigation and there are a bunch of variables defined as lightening other colors, resulting in colors that aren't defined in the palette. I saw #171717, #232323, and #363636 and there are actually over 25 variables for the navbar that are defining lightened colors.

matthewcarleton:
@jennyhaines thanks for that catch :) I've updated the :hover, :active states and added the border to the vertical nav icon as well. I've used the variables that were being used to avoid the possibility of adding new colours.

jennyhaines:
@matthewcarleton - The updates look great! Thanks!

jeff-phillips-18:
@matthewcarleton Is this all set then?

andresgalante:
@jeff-phillips-18 yeap, done 👍 

@matthewcarleton thanks a lot!

PR #679, opened 2017-07-06
ci(releasing): Add travis config and semantic-release

dabeng:
This PR introduces [semantic-release](https://github.com/semantic-release/semantic-release) for automating releases of patternfly to NPM via Travis. Semantic-releases requires a strict [commit message format](https://github.com/semantic-release/semantic-release#default-commit-message-format) which can be ensured by running npm run commit with [commitizen](https://www.npmjs.com/package/commitizen). This process ensures release automation of patternfly-react with every merge to master and proper semantic versioning at all times.

@priley86 @bleathem , please review :)


release.sh
```diff
@@ -0,0 +1,42 @@
...
+  ssh-add deploy_key
+}
+
+if [ "${TRAVIS_REPO_SLUG}" != "${TRIGGER_REPO_SLUG}" -o "${TRAVIS_BRANCH}" != "${TRIGGER_REPO_BRANCH}" ]; then
+  echo -e "${RED}Exiting, this is not a production release.${NC}"
+  exit 0;
+fi
+
+getDeployKey
+npm run build-storybook
```
priley86: We will want to modify the way we do test pages here (based on existing build process in PF). React Repo is setup a bit differently and uses storybook which does not exist in PF Core.

priley86 (changes requested):
This is pretty close...just need to setup the test page bits for PF Core...

release.sh
dabeng: Do I just need to remove the line "npm run build-storybook" ?

cdcabrera: Indeed, we're going to need to remove that line @dabeng along with any following "storybook" references,  and aim for what PatternFly focuses on with its ```-dist``` branch, which can be extrapolated from the [PatternFly-Eng-Release repo](https://github.com/patternfly/patternfly-eng-release). May take a bit to pick apart.

@priley86 additional recommendations for replacements?

package.json
```diff
@@ -72,7 +75,12 @@
     "build": "grunt build",
     "jekyll": "ruby --version || echo \"Ruby required\"; bundler -v || gem install bundler; bundle check || bundle install; set -e",
     "start": "npm run build; grunt server & open http://localhost:9000",
-    "grunt": "grunt"
+    "grunt": "grunt",
+    "commit": "git-cz",
```
cdcabrera, line 80: Not necessarily a big deal, but if we're replacing a Git process with a single NPM script alias maybe think about adding a ```git add .``` to this ```commit```, like:

```javascript
"commit": "git add .; git-cz"
```

cdcabrera (dismissed):

package.json
cdcabrera: Caught up with @priley86 on this one, ignore this comment

priley86:
@bleathem thoughts? I'm not entirely sure yet @cdcabrera ...

bleathem (changes requested):
We are missing the build step and commit to dist, I'll add some line comments describing where to put them.

release.sh
```diff
@@ -0,0 +1,41 @@
...
+  ssh-add deploy_key
+}
+
+if [ "${TRAVIS_REPO_SLUG}" != "${TRIGGER_REPO_SLUG}" -o "${TRAVIS_BRANCH}" != "${TRIGGER_REPO_BRANCH}" ]; then
+  echo -e "${RED}Exiting, this is not a production release.${NC}"
+  exit 0;
+fi
+
+getDeployKey
+cd .out
```
bleathem: This is the storybook build folder, we can drop this.

release.sh
```diff
@@ -0,0 +1,41 @@
...
+git config --global user.email $COMMIT_AUTHOR_EMAIL
+git config --global user.name $COMMIT_AUTHOR_USERNAME
+git init
```
bleathem: We don't need this line either

release.sh
```diff
@@ -0,0 +1,41 @@
...
+git add .
```
bleathem: Change this
```
git add .
```
 to:
```
npm run grunt build
npm add . -f
```

bleathem (dismissed):
We are missing the build step and commit to dist, I added some line comments describing where to put them.

dabeng:
@bleathem , awesome commits. I have no doubt about it :+1:

priley86:
@bleathem can we ensure this doesn't break our current Travis build process in a dev fork? I would <3 confirmation on this...just know that a lot of folks are still relying heavily on this process. We ended up doing this a bit differently in PF-R as we opted to go the CLI route (and allow the consumer to push a rawit branch in an automated fashion). 

I have been referring people who ask how to setup Travis in their fork to this guide: https://github.com/patternfly-webcomponents/patternfly-webcomponents/blob/master/TESTS.md#enabling-travis-builds. But it's not ideal b/c it's not PF specific...

Can we add some docs alongside this?

.travis.yml
```diff
@@ -10,6 +11,17 @@ cache:
...
+  - COMMIT_AUTHOR_USERNAME: "patternfly-build"
+  - TRIGGER_REPO_SLUG: "patternfly/patternfly"
+  - TRIGGER_REPO_BRANCH: "master"
+
+notifications:
+  email: false
+
 before_install:
   - rvm install 2.3.1
   - if [[ `npm -v` != 3* ]]; then npm i -g npm@3; fi
```
dlabrecq: We should consider removing this. I don't think this is needed anymore? We took it out of patternfly-ng because of issues with the symantic release.

scripts/release.sh
```diff
@@ -0,0 +1,40 @@
...
+
+getDeployKey
+git config --global user.email $COMMIT_AUTHOR_EMAIL
+git config --global user.name $COMMIT_AUTHOR_USERNAME
+npm run grunt build
+git add . -f
+git commit -m "Added dist folder"
+
+git remote add ssh-origin $SSH_REPO
+git push --force --quiet ssh-origin master:master-dist
```
dlabrecq, line 40: We need to merge the master-dist branch. Cannot simply force push because you'll wipe out commits, tags, etc. 

Note that we're still using the patternfly-eng-release scripts to build and push to master-dist. This particular script is duplicating that, so your push may not do anything at all -- nothing new to push.

This looks like a script from the patternfly-react repo? They're not using patternfly-eng-release, which is fine, but there is far more that you need to handle here.

Although we don't need to perform an npm release, patternfly-eng-release still builds, runs tests, etc. It also ensures that Travis only pushes to master-dist from the master branch. It also ensures this runs on a merge, not a pull request -- most importantly, not from someone's forked repo.

dlabrecq: Please take a look at patternfly-ng to see how the semantic release and patternfly-eng-release are used together.

dlabrecq (changes requested):

scripts/release.sh
dabeng: @dlabrecq , I need your advice about how to write option "publish-travis" in package.json file.

dlabrecq: You can read up about how to use the publish-travis piece here: https://github.com/patternfly/patternfly-eng-publish

FYI, @bleathem wrote those instructions.

bleathem: > We need to merge the master-dist branch. Cannot simply force push because you'll wipe out commits, tags, etc.

Force-pushing to the master-dist branch will not wipe out tags.  We will lose references to old un-tagged commits, but IMO that's an advantage, not a disadvantage.  Tagged commits (eg. releases) will persist as off-shoots from the master branch.  The master-dist branch will then exist as always 1 commit ahead of master.  If this is unclear I can create and share a graphic showing what the commit history would look like with this approach.

dlabrecq: Force push was also my first choice, but it didn't work out as expected. I believe we're publishing from and tagging master-dist (to include the generated files), so we cannot simply wipe out its history. We need the generated files because bower installs run off of the master-dist branch tag.

Semantic releases are also based on based on commit differences. That means you cannot add a commit for a push to master-dist, and then run the symantic release. It will error because the commit for the generated files is not in master.

We encountered that with patternfly-ng, so we adjusted the symantic release to run on master-dist. It may be possible to run the symantic release before pushing to master-dist, but we probably still want to include those generated files.

priley86: I'm fine w/ adopting the patternfly-ng approach (semantic-releases + the build scripts for automated -dist branches). Should we choose to explore Storybooks in the future, this may become moot though as we've implemented a command in the storybook-deployer for deploying test pages to gh-pages branch or a rawgit branch in a dev's fork. It may still be desired to have automated Travis builds in a dev's fork for deploying test pages, but I don't think it should be required (as a command may be less setup overhead than setting up Travis). I think supporting both options might be nice...

Thanks for providing some helpful feedback on this @dlabrecq @bleathem

dlabrecq: Agreed. We just have to ensure master dist is merged, regardless of how we deploy test pages or whether we adopt the patternfly-ng approach.

My point is that we must include the generated files in a tagged branch. If they're added to master-dist, then that's where we must add our tags. It matters for patternfly and angular-patternfly because these repos still support bower installs.

priley86:
@dlabrecq @dabeng @bleathem i'm guessing this PR can be closed now? We've already added semantic-releases here...

dlabrecq:
Yes, @dabeng asked me about closing this. Thought it was closed a while ago.

PR #690, opened 2017-07-21
Added browsersync for live reload of the jquery showcase.  Implemented handlebars templates for including html

bleathem:


amarie401:
@bleathem thanks!

Issue #691, opened 2017-07-24
API to hide/show main menu removed?


gashcrumb:
Hi!  I had added a little API to hide and show the main navigation widget in this pr:

https://github.com/patternfly/patternfly/pull/582

For this issue:

https://github.com/patternfly/patternfly/issues/580

But I just updated the version of patternfly to the 3.25.1 and had this issue reported:

https://github.com/syndesisio/syndesis-ui/issues/638

From what I can tell in the latest code none of my changes were carried forward?  Is there some other way to achieve the desired behavior?

gashcrumb:
@bleathem any word on this one?  This regression messes up the main views in our app...

bleathem:
Sorry abuot this @gashcrumb.  your contribution seems to have gotten lost with the patternfly-functions.js split.  I've created a new PR (#695) re-introducing your contribution.

gashcrumb:
Awesome, thanks!

LHinson:
@bleathem @gashcrumb can we close this issue since #695 has been merged?

gashcrumb:
Yep, sorry, thanks!

LHinson:
No worries! Just didn't want to prematurely close anything @gashcrumb

PR #582, opened 2017-02-03
Return a handle to allow js to toggle the collapsed state of the vertical navigation bar

gashcrumb:
## Description
A few sentences describing the overall goals of the pull request's commits.
Include related PRs or Issues

Fixes https://github.com/patternfly/patternfly/issues/580

Now calling `$.setupVerticalNavigation()` returns an object with 3 functions:

```
var myNav = $.setupVerticalNavigation();

// collapse the vertical nav
myNav.collapseMenu();

// expand the vertical nav
myNav.expandMenu();

// Check if the nav is already collapsed:
console.log("Nav collapsed? : ", myNav.isCollapsed());
```

## Todos
- [ ] cross browser test
- [ ] Are you sure it works on IE9?
- [ ] Is it reponsive?

## Steps to test or reproduce and link to rawgit
Outline the steps to test or reproduce the PR here.

```sh
git pull --prune
git checkout <feature_branch>
bundle; script/server
```

I tested against `dist/tests/vertical-navigation-primary-only.html` and `dist/tests/vertical-navigation-with-secondary.html` though the former is how we're currently using the nav bar.

Also, you may want to please any @mentions you'd like to review.
@LHinson @andresgalante FYI, lemme know what changes are required or if there's any questions.  Thanks!

PR #695, opened 2017-07-26
Return a handle to allow js to toggle the collapsed state of the vertical navigation bar

bleathem:
This PR reapplies #582 that was lost when the patternfly-functions.js file was split.  Adresses #691.

@jeff-phillips-18 @gashcrumb would you please review this PR since you were the original reviewers/contributors?


gashcrumb (approved):
Awesome, you're the best!  It looks good to me at least...

jeff-phillips-18 (approved):
👍

PR #696, opened 2017-07-27
Add css generated icon for drawer-pf-toggle-expand

jeff-phillips-18:
Removal of this caused a backward compatibility issue where applications would have to add ‘fa fa-angle-double-left’ when not expanded, expanded still worked fine.

@bleathem Please consider this for patch release


andresgalante (approved):
👍 thanks a lot Jeff

bleathem (approved):

PR #702, opened 2017-08-08
Add .less suffix to @import application-launcher for SASS compat

skateman:
The absence of the suffix is making the automated conversion cry. It's also inconsistent with the other import calls.


jeff-phillips-18 (approved):

andresgalante:
@skateman thanks a lot

Issue #713, opened 2017-08-16
Notifications drawer in horizontal layout is misaligned on mobile and doesn't match the design guides


gregnolle:
The notifications drawer implementation does not match the design guide screenshots under "Responsive State". In particular:

- Bell icon isn't visible in the toolbar in mobile mode (it gets put in the hamburger menu)
- When the hamburger menu is hidden, the drawer gets hidden, and when you then reopen the hamburger menu the notifications drawer is visible again. I guess this is because the drawer has been put in the navbar collapse div.
- The full width of the screen isn't used
- The top of the drawer isn't aligned with the bottom of the nav bar.

Screenshot attached for reference (taken from the tests page).

![screen shot 2017-08-16 at 17 14 22](https://user-images.githubusercontent.com/6773397/29373751-0eb315a2-82a7-11e7-85af-c32d157aa70f.png)

LHinson:
Hey @gregnolle I've just linked to the PR that is up to address some of these issues for vertical navigation. We have upcoming work queued to address this for horizontal.

PR #717, opened 2017-08-22
updates notification drawer to support mobile

matthewcarleton:
## Description
The purpose of this is to add support for the notification drawer on vertical nav for mobile.

We will still need to address the issue with the jQuery where to drawer does not close when the hamburger menu is selected. **This was fixed with Jeff's PR below.**

This is related to this story:

https://patternfly.atlassian.net/browse/PTNFLY-1461

* On smaller screens I've added a full height.
* I removed the expansion option on the drawer on smaller screens.
* In order to have the drawer sit above the menu on smaller screens I've introduced a unique z-index of 1035.

Vertical Example
https://rawgit.com/matthewcarleton/patternfly/notification-drawer-mobile-dist/dist/tests/notification-drawer-vertical-nav.html

Horizontal Example
https://rawgit.com/matthewcarleton/patternfly/notification-drawer-mobile-dist/dist/tests/notification-drawer-horizontal-nav.html

## PR checklist (if relevant)


src/less/navbar-vertical.less
```diff
@@ -18,7 +18,7 @@
     position: fixed;
     right: 0;
     top: 0;
-    z-index: @zindex-navbar-fixed;
+    z-index: 1035;
```
jeff-phillips-18: This is a little concerning.  Lots of the vertical-nav.less classes use  @zindex-navbar-fixed as a base and add to that ( https://github.com/patternfly/patternfly/blob/master/src/less/vertical-nav.less#L293-L298 ). Adopters may also be using this to be able to show items over the navbar.

Is there sufficient reason to change this?

src/less/notifications-drawer.less
```diff
@@ -201,6 +201,12 @@
   .drawer-pf {
     height: calc(~"100vh - 80px");//to create a 20px offset bottom
     top: 58px; //menu height;
+
+      @media (max-width: @screen-xs-max) {
+        width:100%;
+        height: 100vh; // override to show use full height
+      }
```
jeff-phillips-18: Should the top value (58px) be subtracted?

tests/pages/_includes/widgets/notification-drawer.html
```diff
@@ -1,6 +1,6 @@
 <div class="drawer-pf hide drawer-pf-notifications-non-clickable">
   <div class="drawer-pf-title">
-    <a class="drawer-pf-toggle-expand fa fa-angle-double-left"></a>
+    <a class="drawer-pf-toggle-expand fa fa-angle-double-left hidden-xs"></a>
     <a  class="drawer-pf-close pficon pficon-close"></a>
```
jeff-phillips-18, line 4: We don't show the expander at xs resolution but the drawer could still have the drawer-pf-expanded class (if expanded prior to shrinking the window). Is this something we should handle?

src/less/navbar-vertical.less
matthewcarleton: @jeff-phillips-18 we can revert it back now that we have the hamburger menu closing the drawer. This was changed to ensure the drawer sits about the open menu but that shouldn't be a problem.

src/less/notifications-drawer.less
matthewcarleton: @jeff-phillips-18 I'll add it in, it didn't see any visible difference when I had it but it can't hurt.

tests/pages/_includes/widgets/notification-drawer.html
matthewcarleton: @jeff-phillips-18 good question. How concerned are we with these types of scenarios. They seem unlikely to me.

jeff-phillips-18: I'm OK with leaving it out. The JS for this is not shipped as an available function anyhow. We'll just need to remember to handle this in the JS repos.

src/less/navbar-vertical.less
matthewcarleton: @jeff-phillips-18 ok, the issue is now. If I change it back to 1030 then the menu can be open and if you open the drawer it's under the menu. Does it make sense to have the menu close if you open the drawer?

jeff-phillips-18: Yeah, it should close

src/less/notifications-drawer.less
jeff-phillips-18: it is required for the drawer to expand/collapse correctly and still show the last group or actions/notifications if there is more than fit on the page.

matthewcarleton: ah ok

matthewcarleton:
@jeff-phillips-18 I've updated this to reflect the changes requested in https://github.com/patternfly/angular-patternfly/issues/576 by @benjaminapetersen. Also added the js provided.

tests/pages/_includes/widgets/notification-drawer.html
matthewcarleton: with the change I just pushed I added in a media query to remove the left:270; that is added when expanded. It now will just take up full width. I think this is a good solution, what do you think?

jeff-phillips-18: Agreed.

benjaminapetersen:
Relevant to web console: https://github.com/openshift/origin-web-console/pull/1326

src/js/patternfly-functions-vertical-nav.js
```diff
@@ -3,7 +3,7 @@
 (function ($) {
   'use strict';
 
-  $.fn.setupVerticalNavigation = function (handleItemSelections) {
+  $.fn.setupVerticalNavigation = function (handleItemSelections, ignoreDrawer) {
```
benjaminapetersen, line 6: Do you follow a convention when adding methods to the `jQuery.prototype`?  For example, I've seen elsewhere `pfDonutTooltipContents`, `pfPieTooltipContents`, etc.

benjaminapetersen: I realize you didn't create the function in this PR & are just modifying, but wanted to point this out just in case.

src/js/patternfly-functions-vertical-nav.js
```diff
@@ -259,6 +261,15 @@
               // Always start at the primary menu
               updateMobileMenu();
               navElement.addClass('show-mobile-nav');
+
+              // If the notification drawer is shown, hide it
```
benjaminapetersen, line 265: I'd suggest breaking this out into a `handleDrawer()` kind of function, rather than adding complexity to `bindMenuBehavior`, but totally up to your team style guidelines.

src/js/patternfly-functions-vertical-nav.js
```diff
@@ -259,6 +261,15 @@
...
+              if (!ignoreDrawer) {
+                $drawer = $('.drawer-pf');
+                if ($drawer.length) {
+                  $('.drawer-pf-trigger').removeClass('open');
+                  $drawer.addClass('hide');
+                }
+              }
             }
           } else if (navElement.hasClass('collapsed')) {
             window.localStorage.setItem('patternfly-navigation-primary', 'expanded');
```
benjaminapetersen, line 275: I wonder if `sessionStorage` would be better for things like `patternfly-navigation-primary`.  Is it essential to keep this information between browser restarts?  I would tend to expect it to return to default `collapse` state between.  

Also not part of your PR so feel free to ignore :)

src/less/notifications-drawer.less
```diff
@@ -42,7 +42,9 @@
     display: block;
   }
   &.drawer-pf-expanded {
-    left: 270px;
+    @media (min-width: @screen-xs-max) {
```
benjaminapetersen: Should it go full screen on mobile rather than partial?  Orig comment: @bean66

Also bringing my gif back, If I expand the drawer, then shrink the page down, the drawer will go off screen (and stay off screen) as I click the toggle:

benjaminapetersen: Reference screenshot re: should it go full width on mobile:

pf-drawer-not-full-mobile

matthewcarleton: You are looking at the horizontal example which hasn't been fixed yet, this PR started out as just for the vertical example. I am working on the horizontal and will push shortly.

matthewcarleton:
@jeff-phillips-18 @benjaminapetersen 
I've updated this to support the horizontal menu as well. I'm wondering if showing the notification icon while showing the drawer on mobile makes sense. It will offer context when screen is sized down as well as show update when all notifications are read.

src/less/notifications-drawer.less
```diff
@@ -220,6 +235,11 @@
   .drawer-pf {
     height: calc(~"100vh - 46px");
     top: 26px;
+    @media (max-width: @screen-xs-max) {
+      width:100%;
+      height: calc(~"100vh - 58px");
+      top:34px; // this pushed the drawer below the notification icon in the menu
```
benjaminapetersen: Curious if the `top: 34px` could be a variable that we could build off of.  I know we have had conversations in the web console about having things pile up in that top left corner, once you open the drawer & have a toast or two... @rhamilto thoughts?

src/less/notifications-drawer.less
```diff
@@ -201,6 +210,12 @@
   .drawer-pf {
     height: calc(~"100vh - 80px");//to create a 20px offset bottom
     top: 58px; //menu height;
+
+      @media (max-width: @screen-xs-max) {
+        width:100%;
+        height: calc(~"100vh - 58px");
```
srambach, line 216: I'd love to see the px measurements in variables, especially since they are used multiple times and height relates to top.

rhamilto:
>Curious if the top: 34px could be a variable that we could build off of. I know we have had conversations in the web console about having things pile up in that top left corner, once you open the drawer & have a toast or two... @rhamilto thoughts?

I think all of these positioning values should be variables so they can easily be overridden downstream.  Otherwise entire rules have to be duplicated...

src/less/notifications-drawer.less
benjaminapetersen: Yup, esp if there is relationships between these values across components.

matthewcarleton:
@srambach @benjaminapetersen @rhamilto That makes a lot of sense. @LHinson should we pull @andresgalante into this when he is back?

src/js/patternfly-functions-vertical-nav.js
jeff-phillips-18: This was done specifically for an application that needs this state maintained between restarts.

benjaminapetersen: Hmm. possible to consider making these kinds of things configurable so that the needs of one app don't have to affect another?

jeff-phillips-18: Sure, this seems separate from this PR though. Please file an issue and consider contributing a PR.

LHinson:
Thanks @srambach @rhamilto @benjaminapetersen for the reviews. After syncing with @matthewcarleton, we are in agreement with your recommendation to make the position valuables variables. 

I've filed two new issues for this. One for the notification drawer #725 and the other for the toast notifications #726.  

Considering this update out of scope for this particular PR however will address that update next.

src/js/patternfly-functions-vertical-nav.js
benjaminapetersen: Yup, not relevant for this PR.  Will do!

jeff-phillips-18 (approved):

src/less/navbar-vertical.less
```diff
@@ -46,7 +46,7 @@
           color: @navbar-pf-vertical-active-color;
         }
       }
```
benjaminapetersen, line 48: Quick nit, you prob could eliminate this file from the PR as its just a whitespace change, or make sure you are using the `patternfly/~.editorconfig`.  If everyone contributing is using it little whitespace shuffles like this should disappear.

benjaminapetersen:
Pulled these updates into my local running `origin-web-console`.  Noticed the following issue:



I believe you are switching to a mobile layout too soon:

screen shot 2017-08-29 at 10 37 43 am

So it seems 1,2 or 3 columns of content should work (the beauty of using flex), with varying lengths of text.  I'd definitely like to see the date/time block collapse a bit as size decreases.  There is a lot of opportunity to reclaim lost whitespace.

matthewcarleton:
@benjaminapetersen curious to see how this is built. Can you show me the markup for one of the notifications?

spadgett:
@matthewcarleton We're using angular-patternfly, which is handling the markup.

@jeff-phillips-18 Are any changes are needed downstream in angular-patternfly for this PR?

jeff-phillips-18:
@spadgett I think it is the notification body include that Matt is interested in. Angular Patternfly does not set the html layout of the notification body.

spadgett:
Ah right, understood. Here is the markup structure:

https://github.com/openshift/origin-web-console/blob/master/app/views/directives/notifications/notification-body.html

matthewcarleton:
@spadgett ah ok. So from what I can see you've changed the structure quite a bit and that's going to effect how the areas react at the breakpoint. I can create a quick html prototype to show what you'd need to get it to work if that is helpful.

benjaminapetersen:
I am somewhat interested in the reason to move away from a `container-fluid` with rows/cols inside that use `col-md*`, `col-sm*`, etc to [adjust the layout](https://getbootstrap.com/docs/3.3/css/#grid-example-mixed-complete).  Seems like that might be a good, generic but bootstrap consistent way to solve this?

matthewcarleton:
@benjaminapetersen I wasn't here when that decision was made but I would assume it's taken from bootstrap itself. If you look at their components they don't use page layout within theirs either. It could be to avoid verbose code. @andresgalante would have some more insight on this I'm sure.

spadgett:
@matthewcarleton What is the normal structure? We can try to align if that's the right thing to do and it accommodates the design.

This might break other downstream projects, though, since angular-patternfly allows any markup for the notification body. It seems like the CSS is assuming a lot about what the content is. Chaning flex-direction from row to column is significant :)

We've had similar problems in the past upgrading Patternfly with components like list-view.

LHinson:
addresses comments seen in #713

matthewcarleton:
@spadgett @benjaminapetersen have a look at the following: http://recordit.co/Lf1wsgXM4Y

I believe this is what you are trying to accomplish.

I've followed the example code here http://www.patternfly.org/pattern-library/communication/notification-drawer/#/code

I've added an additional 
`<span class="drawer-pf-notification-message">Your additional content here</span>`
after the initial:
`<span class="drawer-pf-notification-message">A New Event! Huzzah! Bold!</span>`

This will allow you to make a simple change but stick to what patternfly suggests.

LHinson:
Thanks @matthewcarleton for providing that. 

Related to this issue and from a PF perspective, I think that an expanded view of the notification drawer (with additional content) is a new requirement for PatternFly and requires design review and docs. 

Looping in @serenamarie125 as FYI who will help determine the priority for this. I've filed an issue against the patternfly-design repo regarding this issue to kick start a conversation: https://github.com/patternfly/patternfly-design/issues/405

Since the above it out of scope, I'm going to merge this one and we will continue to follow up on the wizard as needed.

benjaminapetersen:
Cross referencing some further updates @sg00dwin has done in the openshift web console to help improve readability with the type of content we are handling.

PR includes some discussion to upstream a few items.

https://github.com/openshift/origin-web-console/pull/1997

PR #722, opened 2017-08-24
Wizard mobile update

matthewcarleton:
## Description
The purpose of this is to add a temporary simplified version of the wizard for mobile. This will be updated to offer a more robust option that will include all the lost functionality.

Related to this story:

https://patternfly.atlassian.net/browse/PTNFLY-1472

## Changes

This mostly just disables functionality for mobile so you'll see a few instances of display:none; 
The rest of the changes are purely styling adjustments to meet the design requirements.

https://rawgit.com/matthewcarleton/patternfly/wizard-dist/dist/tests/wizard.html
## PR checklist (if relevant)


cshinn:
@matthewcarleton is it easy enough to add support for substeps in this implementation? With the current example, it's a bit confusing that it stays on (1) when you hit the next button

matthewcarleton:
@cshinn this is the extent of what I can gracefully do with just the CSS. @jeff-phillips-18 what do you think?

cshinn:
@matthewcarleton I was thinking just the part where it shows the name of the current substep after the main step name. Does that require additional JS?

matthewcarleton:
@cshinn ya, in the desktop version it's only shown on the sidebar and we hide that for mobile.

cshinn (approved):
@matthewcarleton I wouldn't worry about it too much then. Looks good!

matthewcarleton:
@cshinn another reason why this will be better once the extended version is implemented.

jeff-phillips-18:
Without showing hte substep it does not match the design at https://github.com/patternfly/patternfly-design/blob/64ad53e5f555c0489b3c008186a7309b10f82f4b/pattern-library/communication/wizard/design/design.md#simplified-mobile-wizard

matthewcarleton:
@jeff-phillips-18 ya, I wasn't able to go that far with the CSS and from what I understood with the requirements it wouldn't be needed. Should we discuss the JS requirements to get the substeps working?

jeff-phillips-18:
I don't believe there would be any JS requirements just some HTML in the header area that is shown only in mobile mode.

jeff-phillips-18:
Well, I suppose for the example to work correctly we would need to set the text based on the current substep.

matthewcarleton:
@jeff-phillips-18 ya, do you think we should consider? For this example we could just remove the sub steps, I just don't know if that's the best approach.

jeff-phillips-18:
If you want to style the header to have the substep, I can do the JS work to set the text appropriately.

matthewcarleton:
ok, will do.

benjaminapetersen:
"lost functionality" sounds like a potential breaking change?  Any chance screenshots can detail the change?  We consume the wizard downstream, just curious how things will affect us.

jeff-phillips-18:
The design is at https://github.com/patternfly/patternfly-design/blob/64ad53e5f555c0489b3c008186a7309b10f82f4b/pattern-library/communication/wizard/design/design.md#simplified-mobile-wizard

The lost functionality at this point is that in mobile mode you cannot step by clicking on the step/substep titles, you must Next / Back in order to navigate. Mobile navigation is a follow on.

benjaminapetersen:
Ok great, that doesn't sound like it will break us.

matthewcarleton:
noticed an issue with height 
screen shot 2017-08-28 at 10 37 25 am
The most reasonable fix I can see is to have the window adjust to the height of the content. The alternative could be to have a scroll inside the modal for longer content but that doesn't seem like a good approach. The design for this doesn't show it in a modal so not sure how flexible we are with it. @cshinn what do you think? Take a look at the rawgit and you'll see what I've come up with.

benjaminapetersen:
@matthewcarleton if the window grows beyond the height of the browser, what then?  I would prob suggest the modal scrollbar approach in this case.  I don't recall seeing many modal windows that are so long that you have to scroll the entire page to see all content (and the nav buttons get pushed below the fold).  

It would def be good to test.  If the modal is long and the user has to scroll to find the buttons, do you then have to implement an auto scroll to scroll the page back to the top to see the beginning content whenever the "next" or "prev" buttons are clicked?

matthewcarleton:
@benjaminapetersen Thanks for the feedback. 
The issue I see with a scrollbar in a modal is it may be not completely obvious that they need to scroll where as the action at the bottom forces them to scroll.

With your second point I'm not sure I know what you mean? When the "next" "prev" are clicked the modal reloads and shows the top by default doesn't it?

cshinn:
@matthewcarleton I like your solution to the scrolling issue! Though it seems like the step/substep display is not updating correctly. it shows "details" twice instead of changing to "settings"

matthewcarleton:
@cshinn great!  Ya @jeff-phillips-18 still needs to Implement the JS for that to work properly but the html/css is ready to go for it.

benjaminapetersen:
@matthewcarleton prob easiest to see what I mean by making a modal with multiple steps, make sure each step is really long (so you have to scroll down), then click next/prev & see if the experience makes sense.

matthewcarleton:
@benjaminapetersen You are making some good points. The thing to keep in mind with this PR is it will be replaced by a more extensive option that is shown in the design. I think we should consider those options for it.

benjaminapetersen:
Sounds good, no worries, one or two of my comments were off topic anyway. Thanks!

src/less/wizard.less
```diff
@@ -18,76 +23,82 @@
 .wizard-pf-body {
     padding: 0;
     position: static;
+    background:#fff;
```
jeff-phillips-18: This should be @color-pf-white

src/less/wizard.less
```diff
@@ -335,24 +410,31 @@
...
 .wizard-pf-footer {
   border-top: 1px solid @color-pf-black-300;
-  bottom: 0;
-  left: 0;
   margin-top: 0;
   padding-bottom: 17px;
-  position: absolute;
-  right: 0;
-
+  background:#fff;
```
jeff-phillips-18: nit: @color-pf-white

jeff-phillips-18 (approved):
LGTM. Just a couple of nits. Thanks @matthewcarleton

src/less/wizard.less
matthewcarleton: ah good catch, I meant to change that.

jeff-phillips-18:
@LHinson Do we need further CSS review on this or is this good to go?

LHinson:
@jeff-phillips-18 I think this PR is good steps forward, I'm going to merge it and we will continue to iterate to improve this wizard.

@benjaminapetersen
Regarding your comments above about scroll, I wanted you to be aware of a test that we did previously on the topic of how modals scroll. See:  https://blog.patternfly.org/how-we-scroll/ 

Full disclosure though, mobile was not tested so I'm going to talk to Sara (our researcher) about getting a test lined up to do so including your comment below. 

>It would def be good to test. If the modal is long and the user has to scroll to find the buttons, do you then have to implement an auto scroll to scroll the page back to the top to see the beginning content whenever the "next" or "prev" buttons are clicked?

benjaminapetersen:
Sounds good, thanks!

Issue #723, opened 2017-08-25
Vertical navigation is not keyboard accessible


rhamilto:
[Vertical navigation](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/vertical-navigation-with-secondary.html) is not [keyboard accessible](http://webaim.org/techniques/keyboard/).  The addition of href attributes to the &lt;a&gt; in the test pages will partially resolve this issue.  However, secondary and tertiary nav menus will remain inaccessible to keyboard navigation at desktop resolutions without additional changes.  

In [adopting PatternFly's vertical nav in OpenShift](https://github.com/openshift/origin-web-console/pull/1932), this issue results in a regression from our current nav which is fully keyboard accessible.

cc: @spadgett, @ncameronbritt, @serenamarie125, @jeff-phillips-18

LHinson:
Thanks @rhamilto. @matthewcarleton has a story to address the href attributes first. See: https://patternfly.atlassian.net/browse/PTNFLY-2533

We will look at the secondary and tertiary after the above is complete.

LHinson:
@jgiardino will be picking up this story to add further enhancements related to the secondary and tertiary menus. https://patternfly.atlassian.net/browse/PTNFLY-2537

FYI @rhamilto

andresgalante:
@LHinson the CSS part of PR #766 is ready to merge, we need to include the Javascript for it.

LHinson:
@andresgalante I see some open questions on the PR. Are you sure it's ready to merge?

matthewcarleton:
#766 Relates to this and another open issue #839  
Should we consider prioritizing it to get it merged @LHinson? 
@jgiardino suggested breaking it out to deal with these specific issues so that could be an option as well.

jgiardino:
I think the css updates in #766 addresses the issues in #839 
I'm hesitant to really do anything more in the html other than just making the <a> keyboard accessible.

jgiardino:
This is primarily a design and JS issue. I suggest pairing :pear: up a designer and JS dev to investigate a solution that's possible in all the JS repos.

I removed the `css` label since css updates related to keyboard navigation can be included for issue #839.

LHinson:
Closing per a conversation with Robb. Thanks for taking a look at those.

PR #733, opened 2017-08-31
adds href to vertical nav

matthewcarleton:
## Description
This addresses the issue of vertical nav accessibility issues raised here:

https://github.com/patternfly/patternfly/issues/723#event-1226278712

https://patternfly.atlassian.net/browse/PTNFLY-2533

This PR is only for the primary this story exist for secondary/tertiary:

https://patternfly.atlassian.net/browse/PTNFLY-2537

## Changes

This adds href="#" to all <a> tags. 

https://rawgit.com/matthewcarleton/patternfly/vertical-nav-href-dist/dist/tests/vertical-navigation-primary-only.html

@rhamilto further testing may be required on this to ensure it meets accessibility requirements.

### DIFF
--- a/src/less/vertical-nav.less
+++ b/src/less/vertical-nav.less
@@ -72,7 +72,6 @@
       font-weight: @nav-pf-vertical-font-weight;
       height: @nav-pf-vertical-link-height;
       line-height: 26px;
-      outline: 0;
       padding: @nav-pf-vertical-link-padding;
       position: relative;
       white-space: nowrap;
--- a/tests/pages/_includes/widgets/navigation/secondary-nav-amet.html
+++ b/tests/pages/_includes/widgets/navigation/secondary-nav-amet.html
@@ -1,11 +1,11 @@
 <div id="amet-secondary" class="nav-pf-secondary-nav">
   <div class="nav-item-pf-header">
-    <a class="secondary-collapse-toggle-pf" data-toggle="collapse-secondary-nav"></a>
+    <a href="#0" class="secondary-collapse-toggle-pf" data-toggle="collapse-secondary-nav"></a>
     <span>Amet</span>
   </div>
   <ul class="list-group">
     <li class="list-group-item {% if page.nav-tertiary %}tertiary-nav-item-pf{% endif %}" data-target="#amet-detracto-tertiary">
-      <a>
+      <a href="#0">
         <span class="list-group-item-value">Detracto Suscipiantur</span>
         {% if page.nav-tertiary %}{% else %}
         {% if page.nav-badges %}
@@ -27,7 +27,7 @@
       {% endif %}
     </li>
     <li class="list-group-item {% if page.nav-tertiary %}tertiary-nav-item-pf{% endif %}" data-target="#amet-mediocrem-tertiary">
-      <a>
+      <a href="#0">
         <span class="list-group-item-value">Mediocrem</span>
         {% if page.nav-tertiary %}{% else %}
         {% if page.nav-badges %}
@@ -45,7 +45,7 @@
       {% endif %}
     </li>
     <li class="list-group-item {% if page.nav-tertiary %}tertiary-nav-item-pf{% endif %}" data-target="#amet-corrumpit-tertiary">
-      <a>
+      <a href="#0">
         <span class="list-group-item-value">Corrumpit Cupidatat Proident Deserunt</span>
         {% if page.nav-tertiary %}{% else %}
         {% if page.nav-badges %}
@@ -67,7 +67,7 @@
     </li>
     {% if page.nav-tertiary %}
     <li class="list-group-item">
-      <a>
+      <a href="#0">
         <span class="list-group-item-value">Urbanitas Habitant Morbi Tristique</span>
         {% if page.nav-badges %}
         <div class="badge-container-pf">
--- a/tests/pages/_includes/widgets/navigation/secondary-nav-ipsum.html
+++ b/tests/pages/_includes/widgets/navigation/secondary-nav-ipsum.html
@@ -1,11 +1,11 @@
 <div id="-secondary" class="nav-pf-secondary-nav">
   <div class="nav-item-pf-header">
-    <a class="secondary-collapse-toggle-pf" data-toggle="collapse-secondary-nav"></a>
+    <a href="#0" class="secondary-collapse-toggle-pf" data-toggle="collapse-secondary-nav"></a>
     <span>Ipsum</span>
   </div>
   <ul class="list-group">
     <li class="list-group-item active {% if page.nav-tertiary %}tertiary-nav-item-pf{% endif %}" data-target="#ipsum-intellegam-tertiary">
-      <a>
+      <a href="#0">
         <span class="list-group-item-value">Intellegam</span>
       </a>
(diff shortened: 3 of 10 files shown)


rhamilto:
The primary nav focus state is inconsistent with the other items on the page.

screen shot 2017-08-31 at 9 39 06 am

vs.

screen shot 2017-08-31 at 9 39 59 am

Presumably these should be consistent so as not to confuse the user.

jeff-phillips-18:
I don't see the secondary navigation menus being available

src/less/vertical-nav.less
```diff
@@ -72,7 +72,7 @@
       font-weight: @nav-pf-vertical-font-weight;
       height: @nav-pf-vertical-link-height;
       line-height: 26px;
-      outline: 0;
```
priley86, line 75: yes..this is correct.

priley86:
not positive what @jeff-phillips-18 is referring, but wanted to comment specifically on the a11y standards...

The use of an `a href="#"` anchor is OK per WAVE as long as anchor text or content is specified within the link (you can see that WAVE still flags this inside our `.card-pf-aggregate-status-notification` spans which are outside the scope of this PR). This is required for all anchors and buttons. The [old standard](https://www.webaccessibility.com/best_practices.php?best_practice_id=946) disallowed this for browsers that disable javascript, but this is outside the scope of WCAG 2 and is no longer flagged by WAVE if link text is specified.

For the specific WAVE rule in question, see the following links which also describe ways to hide anchor text if desired (though there are several ways to do this):
https://www.w3.org/WAI/WCAG20/quickref/#qr-navigation-mechanisms-refs
https://www.w3.org/TR/WCAG20-TECHS/C7.html  (shows how to use a valid `a href="#"`)

screen shot 2017-08-31 at 2 46 18 pm

priley86:
also agree w/ @rhamilto's mention of consistency of the focus state...

tests/pages/_includes/widgets/navigation/secondary-nav-amet.html
```diff
@@ -1,11 +1,11 @@
 <div id="amet-secondary" class="nav-pf-secondary-nav">
   <div class="nav-item-pf-header">
-    <a class="secondary-collapse-toggle-pf" data-toggle="collapse-secondary-nav"></a>
+    <a href="#" class="secondary-collapse-toggle-pf" data-toggle="collapse-secondary-nav"></a>
```
jeff-phillips-18: Should these just be href="" ?

matthewcarleton: I wasn't sure about that so I just followed bootstraps example. They use the "#"

matthewcarleton:
@jeff-phillips-18 I've updated the description with the link to the story that refers to secondary/tertiary levels.

dlabrecq:
Using an href attribute becomes an accessibility issue if the link doesn't go anywhere. A screen reader may not understand that the link is not used for navigation and attempt to bring the user to the top of the page. You may consider using href="javascript:void(0);" instead?

Also ensure the anchor contains text in its body or use a tooltip to describe what the link does.

priley86:
@dlabrecq I don't think `javascript:void(0);` is necessary any more. I would trust this w3 standard doc as the latest guideline (see Ex:1) . Not sure if there is another a11y specialists we can consult? 
https://www.w3.org/TR/WCAG20-TECHS/C7.html

jeff-phillips-18:
From what I've seen href="" works OK

priley86:
So this one seems somewhat controversial across different browsers and screen readers (and has changed over time)... I read a bit further b/c we don't have consensus. The downside of adding `a href="#"` is it will get added to the history (not ideal) even though this is listed in the spec as a viable option. What's not great about `href=""` is some screen readers will mistakingly read "Empty" or potentially something else if `javascript:void(0)`. It also appears some old browsers will reload with empty hrefs.

After reading this [article](https://yuiblog.com/blog/2008/01/23/empty-links/), it seems that adding a `title` attribute alongside should aid most screen readers. Most however should be OK without `title` if there is something readible inside the anchors link text.

So I don't really know what a perfect solution looks like here and would have to defer to an expert on this. However, it seems that as long as an `href` attribute is present, it will pass a WAVE check (regardless what's inside).

matthewcarleton:
@priley86 Thanks for the research on this. I've found out essentially the same thing. It's hard to get a clear answer as to what is best practice. I think you are right about deferring to an expert here.

jgiardino:
According to [WebAIM](http://webaim.org/techniques/keyboard/):

> Note: An <a> element is only keyboard accessible when it has a non-empty href attribute. `<a>` and `<a href="">` should not be used for links.

Regarding `title` text, for a link to be keyboard accessible for sighted users, the link needs to visibly have focus and the text should be visible on focus. [Don't rely on title text for this](https://www.paciellogroup.com/blog/2013/01/using-the-html-title-attribute-updated/). (As a side note, the bootstrap tooltip widget is keyboard accessible, since it the text displays on focus.)

It does look like most of the <a> elements include text by wrapping a <span> with text, except in these cases:

```
<div class="nav-item-pf-header">
    <a href="#" class="secondary-collapse-toggle-pf" data-toggle="collapse-secondary-nav"></a>
    <span>Ipsum</span>		      
</div>
```

jeff-phillips-18:
I believe OpenShift uses href="" a lot and it is keyboard accessible. @spadgett Am I correct here?

spadgett:
`<a href="">` is specifically for AngularJS. They have an `a` directive to handle it.

https://docs.angularjs.org/api/ng/directive/a

andresgalante:
both `<a href="">` or `<a href="#0">` works since an ID cannot start with a number. I personally like it more with something inside `<a href="#0">`

spadgett:
Any reason not to use `href="#"`?

andresgalante:
Hi @spadgett  when we use `href="#"` the browser will scroll to the top of the page when the link is clicked.

On the other hand, since `#0` is invalid, it just stays there as if the href was empty.

spadgett:
Ah I see. `<a tabindex="0" role="button">` might be sufficient if the `href` causes a problem.

spadgett:
I think you can also use `return false` from the click handling to prevent `href="#"` from going back to the top of the page. (I haven't tested, though.)

andresgalante:
I don't think we should use `tabindex`, anchors are in the tab order already. An anchor without an href is a wrong anchor. having the href empty or `#0` is the same for me, but I am against using tabindex on an achor or using js to deactivate a link when the solution is in html itself and is so simple

src/less/vertical-nav.less
```diff
@@ -72,7 +72,7 @@
...
+      // outline: 0; removing this for accessability.
```
andresgalante: Maybe consider just removing this line instead of commenting it out?

tests/pages/_includes/widgets/navigation/secondary-nav-amet.html
andresgalante: change them all to `#0` please.

andresgalante (changes requested):

matthewcarleton:
@andresgalante updated, thanks!

andresgalante (approved):

Issue #734, opened 2017-09-01
Wizard classes set the sidebar width based on resolution but does not update the content margin accordingly.


jeff-phillips-18:
Since the sidebar width changes but the content margin does not, extra space is used unnecessarily in the content area. The test pages use javascript to update the margin, this is not listed as being a necessary step for adopters.

Would be nice to use flexbox and remove the left-margin all together.

LHinson:
Closing this issue here since #735 addresses it.

PR #735, opened 2017-09-05
fix(wizard): Updates wizard to use flex box

matthewcarleton:
## Description
This changes the Wizard layout to use flex box instead of positioning. It removes the need for JS for page layout.

https://github.com/patternfly/patternfly/issues/734

https://rawgit.com/matthewcarleton/patternfly/wizard-flexbox-dist/dist/tests/wizard.html

## PR checklist (if relevant)

### DIFF
--- a/src/less/wizard.less
+++ b/src/less/wizard.less
@@ -2,12 +2,6 @@
 // Wizard
 // --------------------------------------------------
 .wizard-pf {
-  @media (max-width: @screen-xs-min) {
-    min-height: 100vh;
-  }
-  @media (min-width: @screen-xs-min) {
-    height:100%;
-  }
   margin: 0 auto;
   max-height: 900px;
   padding: 10px;
@@ -33,12 +27,9 @@
   @media (min-width: @screen-xs-min) {
     background: @color-pf-black-100;
     border-right: 1px solid @color-pf-black-300;
-    bottom: 0;
-    left: 0;
+    flex:0 0 auto;
     overflow-x: hidden;
     overflow-y: auto;
-    position: absolute;
-    top: 0;
     .list-group {
       border-top: 0;
       margin-bottom: 0;
@@ -266,17 +257,13 @@
   height: 100%;
   overflow: auto;
   vertical-align: top;
-
   @media (min-width: @screen-xs-min) {
     padding:3em;
-    margin-left: 253px; /* this value is updated by js */
+    flex:1 1 auto;
   }
   @media (max-width: @screen-xs-min) {
     padding:1em;
     width:100%;
-    &[style] {
-      margin-left: 0!important; // this is to override any margin from JS
-    }
   }
   // width: 10000px;
   .blank-slate-pf {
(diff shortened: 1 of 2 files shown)


src/less/wizard.less
```diff
@@ -2,13 +2,18 @@
 // Wizard
 // --------------------------------------------------
 .wizard-pf {
-  height:100%;
+  @media (max-width: @screen-xs-min) {
```
srambach: Any reason not to rearrange both of these into mobile-first (min-width) media queries?

srambach:
Initialization state is very tall with buttons only halfway down - they should be at the bottom, correct?

Also, the width of the left column doesn't seem to quite line up with the current version, creating extra space that takes away from the right column. 
image

There are lots of px measurements that break if, for example, the step names are long. But I'm guessing that's out of scope for this PR. :-/

src/less/wizard.less
matthewcarleton: I can actually take that altogether. That was something that I forgot to remove  :(

matthewcarleton:
@srambach fixed side column width on smaller screens. Not sure about the "Initialization state is very tall with buttons only halfway down - they should be at the bottom, correct?"

Can you provide a screenshot?

srambach:
You got me. I can't reproduce it now. I must have messed with something in the inspector that broke it.

matthewcarleton:
@srambach could it have been cached? There was positioning on the button bar that I was able to remove.

jeff-phillips-18 (approved):
Need to remove the updateWizardLayout function and references to it in tests/pages/_includes/widgets/communication/wizard.html

jeff-phillips-18 (commented):
Having some troubles pulling this into Angular Patternfly. Still checking it out.

src/less/wizard.less
```diff
@@ -427,14 +408,11 @@
...
 
     @media (min-width: @screen-xs-min) {
-      bottom: 58px; /* this value is updated by js */
-      position: absolute;
-      overflow: hidden;
-      top: 172px; /* this value is updated by js */
       width:100%;
-
+      display: flex;
+      height: 900px;
```
jeff-phillips-18, line 413: Does this need to be set? I don't think we need a default height here.

jeff-phillips-18:
Other than the one comment (setting default height) this works really well for angular-patternfly. No more need to set margins is great! Thanks Matt!

jeff-phillips-18 (changes requested):
Just the one required change and then the change for the test page.

src/less/wizard.less
matthewcarleton: If we remove the fixed height it sets the height based on the content. That 900px was from the original implementation of the wizard.

jeff-phillips-18: I would let applications set the value (900px is random anyhow). If we feel strongly about leaving it in, I can adjust but to me it is pretty random and likely to be overridden by applications.

jeff-phillips-18: @matthewcarleton you can leave the height setting, could cause backwards compatibility issues

jeff-phillips-18 (approved):
LGTM 👍

matthewcarleton:
@jeff-phillips-18 great thanks Jeff! @srambach or @andresgalante any issues you see?

jeff-phillips-18:
@srambach Are you OK with this now? This is necessary for the A-PF work which is required upstream.

srambach:
Yeah, I talked to @matthewcarleton and my concerns about the media queries can be done another time.

PR #739, opened 2017-09-07
ci(travis): Added the `npm install` command to the `install` hook of the travis config

bleathem:


priley86:
lgtm 👍

dlabrecq (changes requested):
The npm install is skipped because the build scripts need to run it. This has to be done after removing the npm shrinkwrap or you get old package versions.

The scripts also run nsp after creating the shrinkwrap

jeff-phillips-18:
@bleathem working on master is bad 😟

bleathem:
@dlabrecq running `npm install` in the travis _install_ hook does not preclude the release script from running it.  The release script is free to remove shrinkwrap and re-run the npm install if it needs to do that for a release.  (On a side note, we really should think about breaking down our monolithic install script into something more modular the can be customized for the individual release notes, but that is a matter for another PR :) )

@jeff-phillips-18 I'm not sure I understand the context nor the relevancy of your comment about "working on master".  This PR is from my fork, and is subject to a PR review before being merged to PatternFly master.  What exactly is your concern?

dlabrecq:
See: https://github.com/patternfly/patternfly/pull/741

dlabrecq:
I wouldn't want to run nsp against old package versions -- that's not what we will be releasing? Reinstalling packages later, only to obtain the right versions, is unnecessary and increases the length of the build.

bleathem:
Thanks for updating the release script @dlabrecq! :+1:

PR #741, opened 2017-09-08
Bump patternfly-eng-release version

dlabrecq:
Backed out the recently added nsp check, which requires an npm install. Instead, the nsp check has been added back to the build scripts, which already handle the npm install.


dtaylor113 (approved):

bleathem:
Thanks @dlabrecq!

PR #743, opened 2017-09-08
vertical nav: fix close-on-click hover bug

### LINKED ISSUE #742: vertical navigation doesn't open the just-selected primary option on hover
Vertical navigation doesn't open the just-selected primary option on hover.

Secondary level nav won't open on the currently selected primary option. For example, if you select Option One > Sub-Option One, and then hover again over Option One, the secondary menu doesn't open. Hovering over Option Two and then Option One is a workaround.

Likely a side effect of #673 

See also: https://bugzilla.redhat.com/show_bug.cgi?id=1480134

gregsheremeta:
Fix close-on-click hover bug. Previously, secondary level
nav wouldn't open on the currently selected primary option.
For example, if you selected Option One > Sub-Option One,
and then hovered again over Option One, the secondary menu
wasn't opening. Hovering over Option Two and then Option One
was a workaround.

The cause is that the navUnHoverTimeout was being started up
after a menu was closed on click. navUnHoverTimeout not being
'undefined' then prevented the secondary menu from opening.

## Description
Fixes #742 

## Changes
Only run the navUnHoverTimeout function if a secondary
menu is visible.

### DIFF
--- a/src/js/patternfly-functions-vertical-nav.js
+++ b/src/js/patternfly-functions-vertical-nav.js
@@ -48,6 +48,10 @@
         navElement.find('.mobile-nav-item-pf').each(function (index, item) {
           $(item).removeClass('mobile-nav-item-pf');
         });
+
+        navElement.find('.is-hover').each(function (index, item) {
+          $(item).removeClass('is-hover');
+        });
       },
 
       hideTertiaryMenu = function () {
@@ -60,6 +64,10 @@
 
         navElement.find('.mobile-nav-item-pf').each(function (index, item) {
           $(item).removeClass('mobile-nav-item-pf');
+        });
+
+        navElement.find('.is-hover').each(function (index, item) {
+          $(item).removeClass('is-hover');
         });
       },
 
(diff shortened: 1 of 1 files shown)


gregsheremeta:
please review @bleathem @priley86 @jeff-phillips-18 

cc @vojtechszocs

src/js/patternfly-functions-vertical-nav.js
```diff
@@ -48,6 +48,11 @@
         navElement.find('.mobile-nav-item-pf').each(function (index, item) {
           $(item).removeClass('mobile-nav-item-pf');
         });
+
+        navElement.find('.is-hover').each(function (index, item) {
+          $(item).removeClass('is-hover');
+        });
```
jeff-phillips-18, line 54: Should the same be done in the hideTertiaryMenu function?

gregsheremeta: In theory, maybe, but it doesn't seem to have any bad effects to not remove it. I tested the different menu configs and all seems to work.

jeff-phillips-18: I think it makes sense to add it. It would be more consistent and who knows, some day maybe we have a fourth level 😖

gregsheremeta: done

jeff-phillips-18 (approved):
LGTM.

Note we are using semantic releasing so please take a look at https://github.com/patternfly/patternfly#git-commit-guidelines

gregsheremeta:
@jeff-phillips-18 never heard of it, but ok. Do I need to do something else?

jeff-phillips-18:
Yes, could you squash and reword your commits to:

fix(pfVerticalNavigation):  fix close-on-click hover bug

gregsheremeta:
@jeff-phillips-18 done

Issue #276, opened 2016-05-05
Move Datatable To Angular UI


sherl0cks:
The current datatable implementation uses [jquery datables](https://www.datatables.net/), which is at odds with the other angular components in the framework. It seems to me that patternfly should be using [Angular UI Grid](http://ui-grid.info/) instead.

jeff-phillips-18:
@sherl0cks Did you mean for this issue to be against the angular-patternfly repo?

dtaylor113:
Hi,
patternfly/patternfly uses HTML/CSS/Javascript and jQuery.  patternfly/angular-patternfly contains angular directives for corresponding patternfly layouts and widgets.  We will evaluate using Angular UI Grid when we implement PF datatables in angular-patternfly.  It is in our backlog: https://patternfly.atlassian.net/browse/PTNFLY-883

sherl0cks:
@jeff-phillips-18 not sure, it's possible I got this wrong

@dtaylor113 just so I'm clear, patternfly/patternfly continue to use jQuery datatable, patternfly/angular-patternfly would use angular UI (assuming it is chosen), and end users would be responsible for managing which dependencies get brought in?

dtaylor113:
Hi @sherl0cks , yes that is correct.

sherl0cks:
@dtaylor113 thanks. Closing the issue.

PR #750, opened 2017-09-13
fix(toast notifications): Updates toast notification to offer a variable for position

matthewcarleton:
updates toast notification to sit below nav bar for both vertical and horizontal navigation. A variable was created to offer more flexibility.

This relates to issue #276

https://rawgit.com/matthewcarleton/patternfly/toast-notification-variable-dist/dist/tests/notification-drawer-vertical-nav.html

https://rawgit.com/matthewcarleton/patternfly/toast-notification-variable-dist/dist/tests/notification-drawer-horizontal-nav.html

### DIFF
--- a/src/less/variables.less
+++ b/src/less/variables.less
@@ -164,6 +164,7 @@
 @nav-pf-vertical-tertiary-badge-color:                              @color-pf-white;
 @nav-pf-vertical-tertiary-badge-bg-color:                           @color-pf-black-900;
 @nav-pf-vertical-width:                                             200px;
+@toast-notification-top:                                            80px;
 @pagination-padding-small-vertical:                                 0;
 @panel-default-border-color:                                        @color-pf-black-400;
 @pficon-var-add-circle-o:                                           "\e61b";
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.sublime-workspace
 *.sublime-project
 *.idea
+package-lock.json
 
 # Ignore vim files
 *.swp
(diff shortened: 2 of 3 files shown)


src/less/variables.less
```diff
@@ -164,6 +164,7 @@
 @nav-pf-vertical-tertiary-badge-color:                              @color-pf-white;
 @nav-pf-vertical-tertiary-badge-bg-color:                           @color-pf-black-900;
 @nav-pf-vertical-width:                                             200px;
+@toast-notification-top:                                            80px;
```
srambach, line 167: Just asking - is 80px actually something meaningful, like the navbar height + grid-gutter-width or something? Or just what happens to work? Is the navbar height a variable so that you can calculate this value from it?

matthewcarleton: @srambach Thanks, I should have explained that the 80px is based on the design provided plus what will work for both vertical and horizontal. I don't believe there are heights set on the navbar unless I am missing something.

srambach: There's @navbar-pf-height and @navbar-pf-vertical-height that are both set to 60px but I don't know if they are used consistently. What do you think, is it worth trying?

srambach: So if those aren't actually used, then no, just stick with the px! Otherwise looks fine to me :-)

matthewcarleton: ok great!

andresgalante: I think it can go as is for now.

andresgalante (approved):

andresgalante:
@matthewcarleton remove package lock file from this PR please.

src/less/variables.less
matthewcarleton: ok @srambach if you are happy with it can you merge? Thanks!

jeff-phillips-18:
package-lock.json is due to an npm package update. We should probably add it to .gitignore

jeff-phillips-18 (approved):

PR #755, opened 2017-09-18
fix(wizard): Fixes responsive bug at 480px

### LINKED ISSUE #754: Wizard steps layout broken at exactly 480px wide viewport
screen shot 2017-09-18 at 8 45 54 am

cc:  @matthewcarleton, @jeff-phillips-18, @spadgett

### LINKED ISSUE #757: Usability issues with wizard in landscape orientation on mobile
On a small mobile screen (e.g., iPhone SE) at landscape orientation, the contents of the wizard are cut off.

screen shot 2017-09-18 at 11 21 55 am

When using the iOS simulator in landscape orientation, depending on where you try to scroll, you may or may not be able to view the wizard footer buttons at the bottom or the content in the right-hand column.  In my iOS simulator testing, there appear to be several compounding issues.  It is possible to scroll body (the content behind the modal) in order to see all of the modal.  But is also possible to scroll the inner right-hand column contents of the within modal to see all of them, but not all of the modal.  Couple that with the fact that the URL bar and browser buttons come and go, changing the height of the viewport, and it's maddening experience to try and interact with the modal.  It appears that in practice, [the use of vh units is impractical](https://adactio.com/journal/11690).

More importantly, when using an iOS device directly, my testing shows is no way to scroll to reveal the wizard footer buttons at all.

cc: @spadgett

matthewcarleton:
This resolves issue where the steps don't display properly at 480px. It also fixes an overflow issue
on smaller screens for the main content of the wizard.

This closes #754, closes #757  

https://rawgit.com/matthewcarleton/patternfly/wizard-754-dist/dist/tests/wizard.html

## PR checklist (if relevant)

### DIFF
--- a/src/less/wizard.less
+++ b/src/less/wizard.less
@@ -2,9 +2,9 @@
 // Wizard
 // --------------------------------------------------
 .wizard-pf {
+  padding: 10px;
   margin: 0 auto;
   max-height: 900px;
-  padding: 10px;
   width: auto;
   .modal-content {
     min-height:100%;
@@ -15,18 +15,17 @@
 // Top section of the wizard w/ title and dismiss
 
 .wizard-pf-body {
+    background:@color-pf-white;
     padding: 0;
     position: static;
-    background:@color-pf-white;
 }
 /* styles the sidebard containing the sub-steps */
 .wizard-pf-sidebar {
-  @media (max-width: @screen-xs-min) {
-    display:none;
-  }
-  @media (min-width: @screen-xs-min) {
     background: @color-pf-black-100;
     border-right: 1px solid @color-pf-black-300;
+    display:none;
+  @media (min-width: @screen-sm-min) {
+    display:inherit;
     flex:0 0 auto;
     overflow-x: hidden;
     overflow-y: auto;
@@ -68,8 +67,8 @@
           }
           // line to left side showing active substep
           > a:before {
-            background: @color-pf-blue-300;
             content: " ";
+            background: @color-pf-blue-300;
             height: 100%;
             left: 0;
             position: absolute;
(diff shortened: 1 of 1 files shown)


rhamilto:
Using https://rawgit.com/rhamilto/patternfly/matthewcarleton-wizard-754/dist/tests/wizard.html on my iPhone 6s running iOS 10.3.3, I am still seeing the issue where I am unable to scroll to the bottom of the modal when switching to landscape mode (or vice versa) on step three.

andresgalante:
I've just tested it on a real iphone 6 and on android device and it works fine, what are you seen @rhamilto ?

matthewcarleton:
@rhamilto ok, I'm not having the same experience on a iPhone 6 10.3.3. Do you experience this in the desktop simulator?

rhamilto:
> @rhamilto ok, I'm not having the same experience on a iPhone 6 10.3.3. Do you experience this in the desktop simulator?

I am not seeing this in the simulator, and I suspect that's because of how the translation of the mouse pointer is being handled. 

Steps to reproduce:

1.  Visit https://rawgit.com/rhamilto/patternfly/matthewcarleton-wizard-754/dist/tests/wizard.html on an iPhone with the phone held in portrait orientation
1.  Click the "Launch Wizard" button under "Complete Wizard"
1.  Click the "Next" button four times until Step 3 is visible
1.  Rotate the phone to landscape orientation
1.  Attempt to scroll the modal in order to see all the modal contents

Note:  if you scroll the window at all while it's in portrait orientation, the bug will not occur.

matthewcarleton:
@rhamilto I'm not seeing this on an iPhone 6 following these instructions. @andresgalante can you confirm?

andresgalante:
Same here, I've tested it following the steps and I can scroll it

spadgett:
I was able to reproduce, but it doesn't happen always. Sometimes the page behind the dialog scrolls instead of the dialog. (iPhone 7)

rhamilto:
> I was able to reproduce, but it doesn't happen always. Sometimes the page behind the dialog scrolls instead of the dialog.

I suspect https://stackoverflow.com/questions/41594997/ios-10-safari-prevent-scrolling-behind-a-fixed-overlay-and-maintain-scroll-posi may be part of the issue.

src/less/wizard.less
```diff
@@ -121,17 +120,17 @@
   display: inline-block;
   display: flex;
   list-style: none;
-    @media (min-width: @screen-xs-min) {
+  padding: 15px 0;
+  margin-bottom: 0;
+  background: @color-pf-black-200;
+  border-top: 1px solid @color-pf-black-300;
+    @media (min-width: @screen-md-min) {
```
rhamilto: @spadgett and I were looking at this in the context of the OpenShift catalog and think this breakpoint may be too large for our specific instance and @screen-sm-min would probably be a better fit.  If that won't work for the default, may I suggest this value be configurable via a variable (e.g., @wizard-pf-steps-indicator-breakpoint) so that it could be determined on a per-app basis?

rhamilto: A mockup of the OpenShift catalog at a resolution of 991px:
localhost-9001-home- 1

matthewcarleton: I think it's a reasonable breakpoint, I've looked at the different landscape views and it still works well. That change works for me.

matthewcarleton: Updated to reflect this request.

andresgalante:
Hi! I'll merge this PR and we'll track the iOS scroll issue here #765

Thanks a lot @rhamilto and @spadgett for the help!

PR #762, opened 2017-09-19
Bumped patternfly-eng-release version

dlabrecq:
Bumped patternfly-eng-release version so showcases-dost may be created


bleathem (approved):
Thanks @dlabrecq!

PR #766, opened 2017-09-22
fix(vertical nav) improve accessibility support

jgiardino:
## Description
Update css to apply hover styles on menu items with focus
Update css to prevent keyboard Tab key from shiftin focus to hidden elements
Update html test page example to include aria attributes for better screenreader support

Also includes additional text for screen readers for the following elements:
- the arrow toggle that displays in submenus
- indicating the active menu items
- identifying the menu level (e.g. "Primary Navigation Menu" vs "Secondary Navigation Menu")

**The following updates are needed but not included in this PR:**

- JS updates 
   - to capture click events when a nav item with a submenu has focus and Enter key is pressed (NOTE: click events are currently captured for the menu that displays for mobile devices)
   - to modify aria attributes like aria-hidden and aria-expanded
- providing visual indication to the hamburger when it has focus
- enabling keyboard focus on the menus in the masthead

Jira stories: 
https://patternfly.atlassian.net/browse/PTNFLY-2537
https://patternfly.atlassian.net/browse/PTNFLY-2611

## Changes

* CSS updates:
  * removed underline from nav item text on focus
  * applied hover styles to nav item on focus
  * modified collapse-toggle arrow icon so that if not visible, it cannot receive focus via keyboard or take up space
  * set display none to <a> elements that are hidden, so that they cannot receive  focus via keyboard
* HTML updates:
  Updates are mostly modeled after [Adobe's accessible mega menu](https://adobe-accessibility.github.io/Accessible-Mega-Menu/):
  * Role attributes for `menu` are not added. After researching this, I based this decision on the following reasons:
    * Standard markup is accessible by default, as [noted in this article about aria-select](https://www.stefanjudis.com/blog/aria-selected-and-when-to-use-it/). Changing default roles for elements can alter how assistive technologies interact with the controls and ultimately result in worse UX if you don't know what you're doing.
    * The [w3c example for menus](https://www.w3.org/WAI/GL/wiki/Using_ARIA_menus) recommends using the arrow keys for navigating the menu items.  Since we are not using arrow keys, and instead only using Tab and Enter to navigate menus, then we should not partially follow this example.
    * The `menu` role is not intended for navigation, as noted here: https://webaim.org/blog/three-things-voiceover/
  * Role attributes for `list` are being added. Even though this is redundant, there were a couple of issues I noticed that were resolved by including the default role to `<ul>` and `<li>` elements:
    * aria-label attributes were ignored if a role was not explicitly defined
    * In voiceover on mac, simple nested lists are described as expected without role, but with the additional DOM, this seemed to affect the ability of Voiceover to understand the list structure. However, the results of Voiceover were inconsistent.
  * Aria attributes are defined as shown in the sample html below. For these, I'm mostly following the pattern shown for [Adobe's accessible mega menu](https://adobe-accessibility.github.io/Accessible-Mega-Menu/)
    * The adobe example uses [aria-expanded](https://www.w3.org/TR/wai-aria/states_and_properties#aria-expanded) with aria-controls. I did not include this as I noticed no benefit when testing with Chromevox or Voiceover, and my concerns that that screen readers that support the aria attribute have expectations about what key commands are enabled for the widget.
      * `aria-expanded` is applied to the element that controls the display of the submenu, but not to the submenu itself. 
      * `aria-hidden` is applied to the submenu
    * [aria-selected](https://www.w3.org/TR/wai-aria/states_and_properties#aria-selected) is not supported for links. The best solution I could find for indicating the current active item is to use `        <span class="sr-only">(Active Menu Item)</span>`
    * `aria-label` is applied to the `<ul>` of the submenu

```
NOTE: only aria attributes are included in this example

<div> 
  <ul> 
    <li>
      <a aria-expanded="false"> 
         ...
      </a>
      <div aria-hidden="true"> <!-- submenu -->
        <div>...</div> <!-- submenu header and collapse-toggle -->
        <ul>...</ul>
    </li>
  </ul>
</div>  
```

To be provided

## PR checklist (if relevant)

### DIFF
--- a/tests/pages/_includes/widgets/navigation/secondary-nav-amet.html
+++ b/tests/pages/_includes/widgets/navigation/secondary-nav-amet.html
@@ -1,11 +1,16 @@
-<div id="amet-secondary" class="nav-pf-secondary-nav">
+<div id="amet-secondary" class="nav-pf-secondary-nav" aria-hidden="true">
   <div class="nav-item-pf-header">
-    <a href="#0" class="secondary-collapse-toggle-pf" data-toggle="collapse-secondary-nav"></a>
+    <a href="#0" class="secondary-collapse-toggle-pf" data-toggle="collapse-secondary-nav">
+      <span class="sr-only toggle-label-expanded-pf">Pin Secondary Navigation Menu</span>
+      <span class="sr-only toggle-label-collapsed-pf" aria-hidden="true">Unpin Secondary Navigation Menu</span>
+      <span class="sr-only toggle-label-back-pf" aria-hidden="true">Back to Primary Navigation Menu</span>
+    </a>
     <span>Amet</span>
   </div>
-  <ul class="list-group">
+  <ul class="list-group" role="list" aria-label="Secondary Navigation Menu">
     <li class="list-group-item {% if page.nav-tertiary %}tertiary-nav-item-pf{% endif %}" data-target="#amet-detracto-tertiary">
-      <a href="#0">
+      <a href="#0"
+        {% if page.nav-tertiary %}aria-expanded="false"{% endif %}>
         <span class="list-group-item-value">Detracto Suscipiantur</span>
         {% if page.nav-tertiary %}{% else %}
         {% if page.nav-badges %}
@@ -26,8 +31,9 @@
       {% include widgets/navigation/tertiary-amet-detracto.html %}
       {% endif %}
     </li>
-    <li class="list-group-item {% if page.nav-tertiary %}tertiary-nav-item-pf{% endif %}" data-target="#amet-mediocrem-tertiary">
-      <a href="#0">
+    <li class="list-group-item {% if page.nav-tertiary %}tertiary-nav-item-pf{% endif %}" role="listitem" data-target="#amet-mediocrem-tertiary">
+      <a href="#0"
+        {% if page.nav-tertiary %}aria-expanded="false"{% endif %}>
         <span class="list-group-item-value">Mediocrem</span>
         {% if page.nav-tertiary %}{% else %}
         {% if page.nav-badges %}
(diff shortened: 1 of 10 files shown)


jgiardino:
I have several questions/comments about these updates, and didn't want to note all of them above...

@andresgalante , these are some of the html/css related questions/comments:
 - I removed the underline from the primary nav menu items on focus, but it looks like the [underline for submenus is intentional](https://github.com/patternfly/patternfly/pull/766/files#diff-1aa93921ee93cbf778c0315f5c8c3f3bL584), so I left that.
- When the primary nav is not visible (e.g. when a submenu is pinned), the menu items can still have focus via the keyboard. I'm not sure what the right solution for this is. Should we do something like `.collapsed-tertiary-nav-pf > .list-group > .list-group-item > a, [and similar path to secondary list items] {display: none}` in the case where a tertiary menu is visible? I'd need to test this, and also look at the classes for the mobile case.

@jeff-phillips-18, can you clarify what the purpose of `data-target` is in the vertical nav?

andresgalante:
Closes #723

jeff-phillips-18:
@jgiardino I don't believe the data-target is used any more, it was originally used to determine what submenu to open.

tests/pages/_includes/widgets/navigation/secondary-nav-amet.html
```diff
@@ -1,11 +1,12 @@
...
+<div id="amet-secondary" class="nav-pf-secondary-nav" aria-hidden="true" aria-expanded="false">
   <div class="nav-item-pf-header">
     <a href="#0" class="secondary-collapse-toggle-pf" data-toggle="collapse-secondary-nav"></a>
     <span>Amet</span>
   </div>
-  <ul class="list-group">
+  <ul class="list-group" role="menu" aria-labeledby="amet-menuitem">
     <li class="list-group-item {% if page.nav-tertiary %}tertiary-nav-item-pf{% endif %}" data-target="#amet-detracto-tertiary">
-      <a href="#0">
+      <a id="amet-detracto-menuitem" href="#0" role="menuitem"
```
spadgett: @andresgalante Is there any chance to reconsider the `href="#0"`? It breaks the browser back button among other problems.

See https://github.com/patternfly/angular-patternfly/issues/641

andresgalante: Yeap, will an empty `herf` have tab index?

andresgalante:
@LHinson we are waiting to do the JS part of this PR to merge it. Is there anyone able to do it?

tests/pages/_includes/widgets/navigation/secondary-nav-amet.html
jgiardino: I tested `href=""` in the browser that PatternFly supports, and it does receive keyboard focus in all of them.

LHinson:
per @jgiardino: 

> I think the CSS is in good shape. I think the html could just be scrapped except for cases where we’re enhancing what’s there but not changing the overall structure.  So mostly just a review of what’s there, and pruning out what we don’t want. I don’t see it as a huge effort.

So, we are going to close this PR and @jgiardino will create a new one after completing the above work + addressing #839

PR #776, opened 2017-10-09
chore(tests): Add visual regression tests

cliffpyles:
## Description
Adds a base for running visual regression tests locally.

https://patternfly.atlassian.net/browse/OSUX-171

To create the reference images for tests run `npm run regressions:reference`. If you would like 
to create the reference image for a specific scenario, pass the name of the scenarios as
an argument. For example, you could run `npm run regressions:reference alerts badges`

To execute the tests run `npm run regressions:test`. If you would like 
to run the test for a specific scenario, pass the name of the scenarios as
an argument. For example, you could run `npm run regressions:test alerts badges`

## Changes

* adds a BackstopJS dependency
* adds configurations for BackstopJS
* adds configurations for the widgets, charts, and layouts sections of the documentation
* adds npm script `regressions:reference` which generates the
   base images used to compare test images against.
* adds npm script `regressions:test` which tests changes against
   the reference images


jeff-phillips-18:
I don't see the dependency for BackstopJS added.

cliffpyles:
Thanks @jeff-phillips-18. Just added it.

dgutride (approved):

jeff-phillips-18:
If I were to add a new test page, what changes would I need to make?

jeff-phillips-18:
When I run 

`npm run regressions:test`

a) it fails
b) it leaves me with a modified file: backstop_data/html_report/config.js

cliffpyles:
for A – Did you already run the command for generating the references?

for B - I'll add it to the gitignore

jeff-phillips-18:
A) I did not,
B) Thanks

cliffpyles:
The issues with the config.js should be resolved now. I did a `git rm` on the html_report directory. Those files should not have been in the repo. They are generated each time you run the tests.

Regarding the reference images – My thought was that we could require the reference images to be generated on the dev environment, since we haven't worked out the details of the external storage yet. So the workflow would look something like this:

- clone the project
- install the dependencies `npm install`
- run the dev server `npm start` (Do you think we should eliminate this step?)
- generate the reference images `npm run regressions:reference`
- run the regression tests `npm run regressions:test`

\*Note: The reference images and the tests can be filtered by passing the name of a scenario as an argument. Ex. `npm run regressions:reference alerts` and `npm run regressions:test alerts`

dtaylor113:
Hi, I picked a bad one to test out.  It seems the About Modal test page doesn't actually contain the modal.  It has a button which launches the modal, which this test harness doesn't trigger:

image

Switched to testing Alerts, and that worked as expected.  Changed the less file and this test caught the change.  Nice!

dtaylor113 (approved):

jeff-phillips-18 (approved):
Looks like a great start for this. Thanks @cliffpyles!

PR #780, opened 2017-10-17
chore(tests): Add regression test optimizations

cliffpyles:
## Description
This commit iterates on the initial regression test implementation. The changes are meant to provide a better developer experience for working with regression tests, and to prepare the code for integration with Travis.

This is an iteration for: https://patternfly.atlassian.net/browse/OSUX-171
And preparation for: https://patternfly.atlassian.net/browse/OSUX-172

## Changes

- Add express server for running tests, to remove need for npm start
- Add approve npm script for confirming the new reference images should be generated
- Added all viewports for all states that Bootstrap accounts for (see: https://getbootstrap.com/docs/3.3/css/#grid-media-queries)
- Optimized diff settings for better visibility
- Updated script naming
- Updated instructions for usage
- Add pdf generation for approval step
- Add npm script for cleaning up hung chrome instances
- Automatically load all scenarios
- Add a disabled feature for scenarios
- Add reference images for regressions


priley86:
@cliffpyles nice job on this... I like the concept of being able to approve changes in our visual ui tests and auto-update them (this is very similar to how Jest does updates of Snapshot tests in React, only a different paradigm here 😸 ). I'm all for automating this process like you've done here.

re: our discussion. It would be cool to see Backstop adopt [Puppeteer](https://github.com/GoogleChrome/puppeteer) eventually, but I agree that IE is much more pressing!

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - '4'
+  - '6'
```
priley86, line 3: I'm not positive if this will impact other downstream projects (i.e. pf.org), but we'd need to confirm that...also need to confirm if all the other grunt plugins still work in Node6.

cliffpyles: I haven't seen any issues with grunt plugins. I'll take a second look though. What would be the best way to find out about the impact of downstream projects?

Also – The Node LTS is currently 6 and will be 8 on the 31st of this month, and the LTS is what we support according to the README. Would that have any affect on this?

For reference this is our statement in the README:
> PatternFly stays up to date with the Node LTS Release Schedule. If you're using PatternFly downstream, we suggest the use of an actively supported version of Node/NPM, although prior versions of Node may work.

backstop/config/scenarios/about-modal.js
```diff
@@ -1,4 +1,5 @@
 module.exports = [{
+  removeSelectors: ['.page-header + .alert'],
   label: 'about-modal',
-  url: 'http://localhost:9000/about-modal.html'
+  url: 'http://localhost:4200/about-modal.html'
```
jeff-phillips-18, line 4: Can we put the port in a config file somewhere? Would be nice to only have to change in one place. Also, it would allow to be easily overridden if someone per chance was already using the port on their device.

README.md
```diff
@@ -242,6 +242,18 @@ or
...
+Visual regression tests provide a way to detect if unintended visual changes have
+occured as a result of changes in the code. They work by taking screenshots of
+what components or pages should look like in a browser (known as references), and then comparing the references to screenshots of those components or pages with your code changes applied. Once the tests are complete, you will be a shown the test results in a browser.
+
+You can run all of the test scenarios with `npm run regressions`.
+
+You can run specify test scenarios with `npm run regressions <scenario-name>`.
+(Ex. `npm run regressions alerts buttons`)
+
+To approve differences run: `npm run approve`
```
jeff-phillips-18: Can you explain when/if this should be done?

jeff-phillips-18 (changes requested):

backstop/config/scenarios/about-modal.js
cliffpyles: In the next PR for CI integration, I was going to implement the environment variables and command line arguments. Can the configuration file be done with those also?

README.md
cliffpyles: Yep.

.travis.yml
priley86: Yes - I seem to remember adding that bit a while ago during the Node 4 upgrade ;)
At the time pf.org had some older dependencies that weren't ready. Since then we've changed the built a lot though...

@bleathem @dlabrecq can you speak to this? Does pf.org sharing any pages impact the Node 6 upgrade here?

README.md
```diff
@@ -242,6 +242,23 @@ or
...
+
+You can run all of the test scenarios with `npm run regressions`.
+
+You can run specific test scenarios with `npm run regressions <scenario-name>`. This
+will probably be most useful while you are doing active development and only want
+to check a few scenarios without running the entire suite.
+(Ex. `npm run regressions alerts buttons`)
+
+To approve differences run: `npm run approve`. This is the command you want to run
+when the tests find conflicts, but the changes/conflicts are intended. This command
```
jeff-phillips-18: Should we call this approve-images or something indicating what you are approving?

cliffpyles: What about approve-regressions? or approve-conflicts?

jeff-phillips-18: approve-conflicts makes more sense, thanks.

jeff-phillips-18 (approved):

jeff-phillips-18:
@bleathem @dlabrecq Do you have any concerns with bumping to Node 6 in .travis.yml?

bleathem:
+1 for updating to node 6

PR #781, opened 2017-10-17
fix(wizard): fixes wizard close button style

### LINKED ISSUE #775: Wizard close button is different than modal close button
screen shot 2017-10-06 at 3 08 16 pm
screen shot 2017-10-06 at 3 08 28 pm

Presumably these should be the same since they offer the same functionality?  The larger click target on the modal is better IMO.

cc:  @sg00dwin, @spadgett

matthewcarleton:
wizard close button now matches the modal button style. I also added the aria label tag from the
wizard to the modal.

In the issue you'll see the modal has the larger "x". I went with this because it uses the pf-icon and it makes more sense to use a larger icon.

closes #775 

## Here is the modal
https://rawgit.com/matthewcarleton/patternfly/wizard-close-button-dist/dist/tests/modals.html

## Here is the wizard
https://rawgit.com/matthewcarleton/patternfly/wizard-close-button-dist/dist/tests/wizard.html

### DIFF
--- a/src/less/close.less
+++ b/src/less/close.less
@@ -2,6 +2,7 @@
 // Close icons
 // --------------------------------------------------
 
+//this should no longer be needed and be replaced with pficon-close. This is here for legacy. For further info see https://github.com/patternfly/patternfly/pull/781
 
 .close {
   text-shadow: none;
--- a/tests/pages/_includes/widgets/communication/wizard.html
+++ b/tests/pages/_includes/widgets/communication/wizard.html
@@ -3,8 +3,9 @@
   <div class="modal-dialog modal-lg wizard-pf">
     <div class="modal-content">
       <div class="modal-header">
-        <button type="button" class="close wizard-pf-dismiss" aria-label="Close"><span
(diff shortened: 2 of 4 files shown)


andresgalante (approved):
👍  It works great here, thanks Matt. @jeff-phillips-18 @dlabrecq @priley86 Will this generate a problem on angular or react?

jeff-phillips-18:
@andresgalante This shouldn't cause any issues since it's just test pages that have changed. Any adopters will need to update their HTML (as will angular-patternfly, patternfly-react, patternfly-ng).

Do we consider it a fix when only updating test pages?

andresgalante:
The test pages are the HTML we provide to use our components, so yes, I'd say it's a fix.

jeff-phillips-18 (approved):

dlabrecq (approved):
LGTM

Thanks for the heads up. I'll create an issue to update the wizard for patternfly-ng

mcarrano (approved):
Looks good to me.

PR #794, opened 2017-10-19
feat(icons): PTNFLY-2589-add-new-icons

andresgalante:
This commit is adding 7 new icons to the icon set following https://patternfly.atlassian.net/browse/PTNFLY-2589


jennyhaines (changes requested):
Hi @andresgalante - thanks for doing this! I just noticed that these icons are not in the html file but should be: "integration", "process-automation", "fa-cogs", and "fa-line-chart". Would also like to see them in rawgit to check if they are rendering properly/ have the correct code strings.

andresgalante:
Thanks for the review @jennyhaines 😄

PR #795, opened 2017-10-19
fix(navbar): href on <a>

### LINKED ISSUE #697: Clickable items in masthead (and possibly elsewhere) don't always get the pointer cursor
The cursor:pointer attribute is being set in quite specific cases such as `.navbar-pf .drawer-pf-trigger-icon`, which has led to some inconsistencies when not explicitly set.
e.g. the Help/About dropdown on [Horizontal Nav](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/horizontal-navigation.html).
It may be better set on something more universal, like `.dropdown-toggle` or `.nav-item` and those classes should be applied consistently to masthead (or other clickable) components.

matthewcarleton:
## Description
There were multiple <a> without an href="" throughout the navbar so I've added it to them all. This ensure consistency with cursors on each.  I've added ```href="#"``` based on previous discussions but open to feedback on this approach.

closes #697

https://rawgit.com/matthewcarleton/patternfly/697-dist/dist/tests/notification-drawer-horizontal-nav.html#

### DIFF
--- a/tests/pages/_includes/widgets/layouts/nav-horizontal-notification-drawer.html
+++ b/tests/pages/_includes/widgets/layouts/nav-horizontal-notification-drawer.html
@@ -13,63 +13,63 @@
   <div class="collapse navbar-collapse navbar-collapse-1">
     <ul class="nav navbar-nav navbar-utility">
       <li class="drawer-pf-trigger dropdown">
-        <a class="nav-item-iconic drawer-pf-trigger-icon">
+        <a href="#0" class="nav-item-iconic drawer-pf-trigger-icon">
           <span class="fa fa-bell" title="Notifications"></span>
           <span class="badge badge-pf-bordered"> </span><!-- in order to show the empty badge this requires a space, otherwise add a value -->
         </a>
       </li>
       <li class="dropdown">
-        <a class="nav-item-iconic" id="dropdownMenu1" data-toggle="dropdown" aria-haspopup="true" aria-expanded="true">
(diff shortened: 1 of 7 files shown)


jeff-phillips-18:
I see href="#" rather than href="0" as you suggest above. I believe the conversation resolved in using href="#0". However, that has proven problematic as well as it does change the URL and break the browser history.

matthewcarleton:
@jeff-phillips-18 ah you are right, I've corrected above. Ok this may be opening this up again for discussion then :)

andresgalante:
Can we have `href="#"` for the test pages on core and then you guys change it for angular or react?

dlabrecq:
If you don't want the link to jump to the top of the page, you could use href="#0" for PatternFly.

There will never be anything on the page with the ID 0. Since "#0" doesn't exist, the click event won't bring you back to the top of page.

The JavaScript equivalent is href="javascript:void(0);" -- we currently use this in patternfly-ng. Although, href="#0" doesn't require JavaScript.

I can say that href="#0" will generate a router navigation error with Angular 4, while  href="javascript:void(0);" does not.

matthewcarleton:
@dlabrecq That's a good point. I've replace all with href="#0". @jeff-phillips-18 are you happy with this? I believe both methods have the same impact on browser history.

jeff-phillips-18:
@matthewcarleton works for me.

jeff-phillips-18 (approved):

PR #800, opened 2017-10-23
fix(accessibility): skip to content link

### LINKED ISSUE #787: Add a skip link to horizontal and vertical nav
To help keyboard only users, lets add a "skip link" to vertical and horizontal nav

matthewcarleton:
## Description
This adds a skip to content link. This is most useful when skipping navigation. As you tab through the page you will see the link appear and disappear when you tab off it. 

I've added this to the navigation pages as well, which can be seen here:

https://rawgit.com/matthewcarleton/patternfly/nav-787-dist/dist/tests/horizontal-navigation.html

Note: If you are using safari you must enable tabbing in your preferences.

closes #787 

## Link to rawgit

https://rawgit.com/matthewcarleton/patternfly/nav-787-dist/dist/tests/skip-to-content.html

### DIFF
--- a/src/less/patternfly-additions.less
+++ b/src/less/patternfly-additions.less
@@ -54,6 +54,7 @@
 @import "notifications-drawer.less";
 @import "search.less";
 @import "sidebar.less";
+@import "skip-to-content.less";
 @import "spinner.less";
 @import "syntax-highlighting.less";
 @import "table-view.less";
--- a/src/less/skip-to-content.less
+++ b/src/less/skip-to-content.less
@@ -1,0 +1,12 @@
+//
+// Skip to content
+// --------------------------------------------------
+.skiplink-pf {
+position: absolute;
+top: 12px;
+left: 100%; /* moves off screen */
+z-index: @zindex-modal;
+}
+.skiplink-pf:focus {
+  left: 10px;
+}
--- a/tests/pages/_includes/head.html
+++ b/tests/pages/_includes/head.html
@@ -34,6 +34,7 @@
     {% endfor %}{% endif %}
     <script src="{{site.baseurl}}/{{site.url-js}}"></script>
   </head>{% if page.full-page %}
+  <a class="skiplink-pf btn btn-primary" href="#main">Skip to main content</a>
   <div class="toast-notifications-list-pf">
     <div class="toast-pf alert alert-warning alert-dismissable">
       <button type="button" class="close" data-dismiss="alert" aria-hidden="true">
--- a/tests/pages/_includes/widgets/navigation/horizontal-primary-nav-bar-page.html
+++ b/tests/pages/_includes/widgets/navigation/horizontal-primary-nav-bar-page.html
@@ -3,7 +3,7 @@
 {% else %}
   {% include widgets/navigation/horizontal-primary-nav-bar.html %}
 {% endif %}
-<div class="container-fluid container-cards-pf">
+<div class="container-fluid container-cards-pf" id="main">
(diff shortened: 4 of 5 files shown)


srambach:
On the Nav example link above, the blue outline doesn't extend completely around the skip link. Is that intended?

srambach:
I'm seeing the skiplink after the toast notifications rather than first thing on the page on the horizontal nav example above.

matthewcarleton:
@srambach good question about the toast notification. I thought it made sense to have it after, do you think it should go before? Looking at the outline issue.

andresgalante:
@matthewcarleton I think it should be before. The skip link should always be the first element in the tab index.

matthewcarleton:
@andresgalante ok, sounds good to me.

dlabrecq (changes requested):
I was able to see the skip like on Chrome, but it doesn't show up (couldn't find it) on Firefox. Is this intended to only work on Chrome?

matthewcarleton:
@dlabrecq thanks Dan, what version/OS are you using? I've got it working in FF

dlabrecq:
I'm using FF 56.0

matthewcarleton:
@dlabrecq hmmm, I have it working on FF 58. What OS is it?

andresgalante (approved):
CSS is ok.
@dlabrecq FireFox by default only focuses on inputs and buttons. You need to go to system preferences on OSX and change your keyboard accessibility settings.

@mcarrano can you please review this one and see if the design is ok?

dlabrecq (approved):
I see it on Firefox now. Thanks, @andresgalante, I didn't have that turned on.

mcarrano (commented):
I played around with this a bit and have the following design related questions/concerns:

* I'm concerned that the pop-up message appears in the upper right of the window when the user's focus is probably initially on the upper left. It would be easy to miss this I think. Should we consider changing the location of the pop-up?

* Can we change the text to say something like "Press Enter to skip to main content".  Not sure it's obvious what the action is, although I have to admit to not using keyboard navigation so I'm not the best test target.

* Is the message configurable per usage? "Skip to main content" might not be applicable in all cases so would be nice if this could be customized per usage.

It would be great to test this or get feedback from either users who regularly use tabbing and/or someone with more expertise in this area than myself. @LHinson what are your thoughts.

andresgalante:
@dlabrecq me neither 😄  I learned all of that from FF from this PR. Thanks for bringing it up

@mcarrano 
>I'm concerned that the pop-up message appears in the upper right of the window when the user's focus is probably initially on the upper left. It would be easy to miss this I think. Should we consider changing the location of the pop-up?

We can place it on the top left corner, IMO it's not relevant. Should we change the position to the top left?

>Can we change the text to say something like "Press Enter to skip to main content". Not sure it's obvious what the action is, although I have to admit to not using keyboard navigation so I'm not the best test target.

Skip links are standard, they always read either "skip nav" or "Skip to main content". We copy "Skip to main content" from Heydon Pickering's book Inclusive Design Pattern.

> Is the message configurable per usage? "Skip to main content" might not be applicable in all cases so would be nice if this could be customized per usage.

Yes, the text is on the HTML, anyone can easily change it on implementation.

mcarrano:
@andresgalante Per your question, yes I think the pop-up should be positioned somewhere on the left. Not sure if this needs to be different for horizontal or vertical nav. I just know that the first time I saw this I missed the pop-up as that's not where my eye was focused.

Based on what you are saying, I'm fine to leave the text as is.

PR #802, opened 2017-10-23
fix(forms): Fixes static form control

### LINKED ISSUE #752: `.form-control-static` is not styled according to Patternfly
`.form-control-static` CSS class does not seems to be styled according ti Patternfly standards. The class exists, but inherits values from Bootstrap. The problem is that the `min-height` is too high (`32px`). It leads to mismatched rows in case of horizontal forms in two or more columns. To fix this issue the `min-height` rule could be removed or set to `auto`.

`.form-control-static` is already included in [Test Pages](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/forms.html) (see Static Control).

Example with `.control-label`:
```html
<form class="form-horizontal">
    <div class="form-group">
        <label class="control-label col-sm-4">Text:</label>
        <span class="col-sm-6 form-control-static">
            Static Text
        </span>
    </div>
</form>
```

Thank you!

andresgalante:
This PR fixes static form control line-height for a correct alignment to labels

Closes #752

@vkrizan can you please confirm that this fixes the issue?

### DIFF
--- a/src/less/forms.less
+++ b/src/less/forms.less
@@ -93,3 +93,9 @@ span.required-pf {
     width: @font-size-large;
   }
 }
+
+.form-group-lg, .form-group-sm {
+  .form-control-static {
+    line-height: @line-height-base;
+  }
+}


vkrizan:
Unfortunately no.

The form control static should match the height of its label in horizontal form and align with the label. Current fix (dea1a72) only touches sm/lg form groups, but it do not fixes the regular group.

Try adding this to form tests:
```html
<h3 id="controls_static-controls">Static Control</h3>
<form class="form-horizontal" role="form">
  <div class="form-group">
    <label class="col-lg-2 control-label">Email</label>
    <div class="col-lg-10">
      <p class="form-control-static">email@example.com</p>
    </div>
  </div>
  <div class="form-group">
    <label for="inputPassword" class="col-lg-2 control-label">Password</label>
    <div class="col-lg-10">
      <input type="password" class="form-control" id="inputPassword" placeholder="Password">
    </div>
  </div>
  <div class="form-group-sm">
    <label for="inputPassword" class="col-lg-2 control-label">Form Group Small</label>
    <div class="col-lg-10">
      <p class="form-control-static">Static Control</p>
    </div>
  </div>
  <div class="form-group-lg">
    <label for="inputPassword" class="col-lg-2 control-label">Form Group Large</label>
    <div class="col-lg-10">
      <p class="form-control-static">Static Control</p>
    </div>
  </div>
</form>
```

From top:
1. "Email" label has height of 20px and top padding of 3px vs. static control has **height 23px** and padding 3px on top and **bottom**.
2. "Password" (regular form) depends on the input, which height is 26 px in total and the text in input and control label are aligned correctly.
3. "Form Group Small" label has ~18px height with top padding of 3px. The `form-group-sm` **does not behave as `.row` in `.form-horizontal`**, but `.form-group` (without sm/lg) does! The static control has height of **25px** and padding 3px on top and bottom.
4. "Form Group Large" seems to be aligned correctly with your fix (`line-height: @line-height-base;`), but it still does not behave as `.row`.

I am not sure if the sm/lg form groups were designed for horizontal forms, since they do not have correct CSS. If we would fix the first static control for normal form group, I'd be very glad. Basically the `.form-group .form-control.static` should behave similar to the input. I am also not sure if it should also have a side padding as the inner value of the input and as the sm/lg groups have.

Thank you for looking into this!

LHinson:
Just wanted to make a note as I know this one has been sitting her awhile. @andresgalante will be picking this back up to see how we can get this one resolved.

PR #806, opened 2017-10-26
fix(app launcher): rename inclusion variables

### LINKED ISSUE #738: Hyphenated variables and multiple inclusion do not work (application launcher)
The issue seems to be caused by liquid, pages built with jekyll are ok.

Application launcher uses hyphenated variables in include:

    {% include widgets/navigation/horizontal-primary-nav-bar.html launcher-icons=true launcher-grid=true navindex=1 %}

which are re-included in `horizontal-primary-nav-bar.html`:

    {% include widgets/framework/application-launcher.html launcher-icons=include.launcher-icons launcher-grid=include.launcher-grid %}

As far as I can tell on the first page the variable `include.launcher-grid` doesn't exist, instead of it there is `include.grid` so the second inclusion is broken. However even fixing the variable names doesn't work, it looks like re-including a variable turns it into a Promise and the variable is inaccessible at the time of the build.

michpetrov:
Rename variables used in application launcher pages. 
Variable names must not contain a hyphen, must be unique across the whole include chain and an include variable must not be referenced in another include tag (in the latter two cases `grunt build` will short-circuit without any explanation).

Closes #738

### DIFF
--- a/tests/pages/_includes/widgets/framework/application-launcher.html
+++ b/tests/pages/_includes/widgets/framework/application-launcher.html
@@ -1,4 +1,4 @@
-<li class="applauncher-pf {% if include.launcher-grid %} applauncher-pf-block-list {% endif %} dropdown dropdown-kebab-pf">
+<li class="applauncher-pf {% if include.grid %} applauncher-pf-block-list {% endif %} dropdown dropdown-kebab-pf">
   <a class="dropdown-toggle nav-item-iconic" data-toggle="dropdown" href="#">
     <i class="fa fa-th applauncher-pf-icon" aria-hidden="true"></i>
           <span class="applauncher-pf-title">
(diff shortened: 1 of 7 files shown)


andresgalante:
Hi @michpetrov thanks a lot for sending this PR.

I can't build your branch, it stops here:
```
Running "pages" task
Builidng test pages with liquid.js
Starting build...
Reading test pages...
```

If there a liquid thing that makes the build stop?

michpetrov:
@andresgalante that's the bug I was running into. I have adjusted the variable names yet again (and added some pages I missed) now it should work.

andresgalante:
Thanks a lot @michpetrov it works now 😄 

@jeff-phillips-18 @dlabrecq can you please help me to spot if this PR is affecting the website?

dlabrecq (approved):
LGTM

The patternfly-ng repo hasn't ported the app launcher yet, so this would only a potential issue for angular patternfly, right now.

jeff-phillips-18:
This has no effect on consumers only on test pages.

jeff-phillips-18 (approved):
Test out the test pages and LGTM.

PR #810, opened 2017-10-27
fix(vertical-nav): fix list group item width

### LINKED ISSUE #807: Vertical nav secondary nav items truncate earlier in IE11 than other browsers
Chrome 61 on macOS
screen shot 2017-10-27 at 10 05 43 am

IE11
screen shot 2017-10-27 at 10 05 29 am

This is due to the fact that [IE11 doesn't support @support](http://caniuse.com/#search=%40support), so [the CSS declaration that unsets the max width on the secondary nav items](https://github.com/patternfly/patternfly/blob/master/src/less/vertical-nav.less#L134) doesn't get unset for IE11.

andresgalante:
Closes #807 

@jeff-phillips-18 Maybe I am missing something, but it seems like the code I removed it's not necessary. I've tested on IE9, 10 and 11 and it works fine.

@rhamilto when you have time, can you please confirm if this Pr fixes your issue?

Thanks a lot!

### DIFF
--- a/src/less/vertical-nav.less
+++ b/src/less/vertical-nav.less
@@ -129,6 +129,8 @@
       line-height: 25px;
       max-width: 120px;
       // If flexbox is supported, do not set max-width, take all space with just some right padding
+      // This generates a known issue on IE11:
+      // https://github.com/patternfly/patternfly/pull/810
       @supports (display: flex) {
         flex: 1;
         max-width: none;


rhamilto:
> @rhamilto when you have time, can you please confirm if this Pr fixes your issue?

It certainly does.  Thanks.

jeff-phillips-18 (dismissed):

matthewcarleton:
@andresgalante I'm seeing some odd overlays of badges and chevrons in IE 9, 10. I believe this is caused by these changes. I tested the current page and it I see the truncated text which resolves the overlay issues. 
screen shot 2017-10-31 at 4 56 47 pm

jeff-phillips-18:
Ah yes! That's what the max-width was for. Figured there was a purpose. Thanks @matthewcarleton

andresgalante:
@rhamilto Are you ok with closing this PR and keeping a difference on IE?

rhamilto:
> @rhamilto Are you ok with closing this PR and keeping a difference on IE?

Given where we are in the release cycle, I'd already addressed this bug in the OpenShift console via an override.  And since it doesn't seem like this can be fixed in a backward-compatible way, I'm fine with closing the issue, but it seems like it should be documented as a known issue?

andresgalante:
I've reverse the change and added a comment pointing to this PR.
@dlabrecq is there a way to merge this PR without generating a release?

jeff-phillips-18:
@andresgalante Update the commit message to chore to avoid the release (and since this isn't really a fix anymore).

matthewcarleton:
@andresgalante is this ready to merge?

jeff-phillips-18:
LGTM, just needs commit message update

andresgalante:
@jeff-phillips-18 is the commit message better now? it's not a fix, but neither any of the other options 😞

jeff-phillips-18:
@andresgalante Change `fix` to `chore` or even `doc`, those won't generate a release.

jeff-phillips-18 (approved):

rhamilto:
@andresgalante, src/sass/converted/patternfly/_vertical-nav.scss is missing from this PR.  The missing file appears on running a build.

PR #811, opened 2017-10-27
fix(login): login alert

### LINKED ISSUE #804: the login page styles for alerts are implemented differently from the design
The design page shows the notification between the logo and the form while they are style normally:
http://www.patternfly.org/pattern-library/application-framework/login-page/#/design

In the `login.less` there is an override for the default alert styles so they became transparent and with a white text color:
https://github.com/patternfly/patternfly/blob/master/src/less/login.less#L52-L55

This is how the example actually look when i add the alert into the html:
login patternfly

matthewcarleton:
This address the issue of the transparent background and white text that overrides the default styles when used on the login page.

closes #804

## Changes
Removes override style.

### DIFF
--- a/src/less/login.less
+++ b/src/less/login.less
@@ -49,10 +49,6 @@
       position: absolute;
       width: 100%;
     }
-    [class^="alert"] {
-      background: transparent;
-      color: @color-pf-white;
-    }
     .details {
       p:first-child {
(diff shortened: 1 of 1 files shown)


srambach:
LGTM, the intended behavior is to have the normal alert background and text color, correct?

srambach (approved):

Issue #661, opened 2017-06-04
npm module size doubled (19.2 MB)


gregsheremeta:
I noticed that between 3.19 and 3.25 that the npm module size shot up. Looks like the font sizes all grew. Perhaps this is by design and not a real issue, in which case feel free to close.

I have a ton of npm modules on my machine, and patternfly is by far the largest of them at almost 20 MB :)

![selection_229](https://cloud.githubusercontent.com/assets/2248293/26762133/80770cec-490a-11e7-89a8-706f9b7fe018.png)

![selection_230](https://cloud.githubusercontent.com/assets/2248293/26762149/f2039344-490a-11e7-851a-bea7a3bb806a.png)

gregsheremeta:
Introduced in commit https://github.com/patternfly/patternfly/commit/c9800ee1a2876aa9accc351ea3618a9173357db0

andresgalante:
Thanks @gregsheremeta We introduce this to add new characters #377

But is a problem. I wonder if open sans can be an opt in and we can use a CDN instead. What do you think?
 
cc @bleathem

gregsheremeta:
I was thinking opt-in a separate opensans npm package, maybe even two (light and full [all the characters]). I guess CDN works, but feels not as polished.

cliffpyles:
@gregsheremeta thanks for bringing this to our attention. I'm making adjustments now.

AdamJ:
The current version (3.45.0) of PatternFly is 16.1 MB zipped, so I will close this issue. If necessary, please open a new issue with a reference to this one.

gregsheremeta:
It used to be only 7. The issue is not fixed, only very slightly alleviated.

AdamJ:
@gregsheremeta The reason why PatternFly is this size is because the distribution now includes both `less` and `sass` files (including dependencies for both versions), compiled CSS for PatternFly, the full tests and a variety of fonts. All of these additions were done by design.

CC: @cliffpyles @andresgalante @dgutride @LHinson

gregsheremeta:
The extra text from sass is trivial. This was something caused by large binaries -- specifically fonts. My argument was that the fonts were too big. Even if that was intentional.

At any rate -- it's better now. 6.8 gzipped from npm / yarn. What happened between 3.27.5 and 3.30.0? Less huge binaries, like fonts, I assume? Did the fonts just shrink?

yarn-offline-cache % ll patternfly*                               
[...]

PR #812, opened 2017-10-30
build(package): Remove extra files from the npm distribution

cliffpyles:
## Description

Currently we are distributing a bunch of extra files with our
npm package. This has inflated the size of the package being
installed, and this commit is to optimize that. More about
the solution can be read here:

https://docs.npmjs.com/files/package.json#files

To test:

- run `npm pack`
- run `tar -tf patternfly-<version>.tgz`
- verify the folder that was extracted only has the dist folder, package.json, LICENSE, and README.md

Related to JIRA ticket:

https://patternfly.atlassian.net/browse/OSUX-233

Related to GitHub issues:

https://github.com/patternfly/patternfly/issues/661


andresgalante:
@gregsheremeta would you mind taking a look at this PR please?

jeff-phillips-18 (approved):

.gitignore
```diff
@@ -25,6 +25,7 @@ npm.cmd
 
 # Ignore generated files
 patternfly.spec
+*.tgz
```
dlabrecq, line 28: Do we have a files with the .tgz extension or was this a one-off created by your specific test case?

dlabrecq (changes requested):

dlabrecq:
Please also run ‘npm install’ on the built repo. Not sure if pack/tar is an equivalent test, but that’s how we test via the Travis build.

.gitignore
cliffpyles: This is created by `npm pack` when the PR is tested. I added it because it will be a common scenario when debugging npm distributions in the future.

It is a file commonly added to .gitignore.

cliffpyles:
@dlabrecq Can you elaborate? There isn't a need to install dependencies again. `tar` is a file archive utility that is built into Linux like environments.

dlabrecq:
Yup, I understand what tar is -- I use it often. I'm just not certain that looking at looking at npm pack with tar is the equivalent of an npm install? We probably should install what npm packs as well.

Prior to publishing to npm, we test by installing the build. For example, run 'npm install &lt;dir&gt;' on your built patternfly repo. We just need to ensure the install runs without errors. 

It's probably fine, but Travis will fail if there is an issue. Or, we have to wait until the npm is actually published to discover an issue. It's just an extra test.

cliffpyles:
@dlabrecq The original issue is that we were including source files and not just the distribution. The only thing that changes with this PR is the files that get published. The rest of the process is the same as it was previously.

To verify what would get published if you were to do an `npm publish`, you can run `npm pack`. The tarball that is produced is what would normally be published to npm. After you have this tarball you can verify that the correct distribution files are being published. In an automated process this would occur after tests had been executed, so running `npm install` again wouldn't serve a purpose.

Read more here: https://docs.npmjs.com/cli/pack

dlabrecq:
I understand what original issue. I'm simply asking that you run an extra test. Are you refusing to do that? Why?

I'm asking because I built/installed your changes myself and saw a potential error. I'm curious if you see the same?

cliffpyles:
I'm not refusing. I'm not sure what you are asking me to do. Are you wanting me to modify the code or perform a task?

cliffpyles:
I need you to be more specific on what the issue is, how you want to resolve the issue, and why that is the resolution for resolving the issue.

gregsheremeta:
@dlabrecq what error did you see?

cliffpyles:
@dgutride @jeff-phillips-18 @bleathem Maybe there is a communication breakdown? Any thoughts?

cliffpyles:
I've run `npm install` on my local. Is that what you are asking?

image

dlabrecq:
I'm not asking you to modify the code, just run a test. When I ran the install, I saw this:

npm WARN enoent ENOENT: no such file or directory, open '/private/tmp/test/package.json'
npm WARN test No description
npm WARN test No repository field.
npm WARN test No README data
npm WARN test No license field.

It's probably nothing and I'm fine with merging. I just want to ensure the npm installed is what we're expecting.

cliffpyles:
Got it. Thanks for your help with the PR!

dlabrecq (approved):

gregsheremeta:
Before:

npm pack
ll patternfly-0.0.0-semantically-released.tgz
-rw-r--r--. 1 greg greg 43079865 Nov  1 12:09 patternfly-0.0.0-semantically-released.tgz

After:
-rw-r--r--. 1 greg greg 6046439 Nov  1 12:10 patternfly-0.0.0-semantically-released.tgz

PR #820, opened 2017-11-02
chore(tests): enable all regression tests

cliffpyles:
## Description
Enables visual regression tests for all pages.


jeff-phillips-18:
The new images for chart components do not look correct.

cliffpyles:
Good catch. I'll update them.

dgutride:
@cliffpyles - agree with Jeff, not just the chart images, others have overlapping text like the accordion.  Any idea what would cause that?

jeff-phillips-18:
Fixed height accordion is likely due to the window size not being computed and acted upon prior to the screen capture.

cliffpyles:
Updates done.

cliffpyles:
While I debug the CI issues, I'm going to close this PR to eliminate extra CI load and noise.

You can track the progress here: https://github.com/cliffpyles/patternfly/pull/5

PR #826, opened 2017-11-03
Bumped version number to 3.30.0-beta

dlabrecq:
Bumped version number to 3.30.0-beta.

There are two issues regarding this PR:

1. The current build deletes everything under src/saas/converted. Was that was intentional?
2. There were 4 regression test failures. However, the test does not fail with a proper exit code, so the build continues normally instead of stopping upon failure

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,6 +1,6 @@
 {
   "name": "patternfly",
-  "version": "0.0.0-semantically-released",
+  "version": "3.30.0-beta",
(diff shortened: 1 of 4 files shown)


dlabrecq:
Issues have been revolved. 

Files under src/saas/converted are no longer deleted and the regression tests are passing.

jeff-phillips-18 (approved):

dgutride (approved):

PR #827, opened 2017-11-06
fix(build): Only clean converted Sass files on Sass build

TheRealJon:
## Description
Less build task no longer runs `clean:sass` which removes the `src/sass/converted` directory.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -558,7 +558,7 @@ module.exports = function (grunt) {
   });
 
   grunt.registerTask('build:less', [
-    'clean',
+    'clean:build',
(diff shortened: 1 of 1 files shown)


TheRealJon:
@dgutride @dlabrecq @jeff-phillips-18 Here is the Gruntfile fix. The `src/sass/converted` directory should now only be removed as part of the `build:sass` Grunt task.

dlabrecq (approved):
LGTM. I built the less-to-sass-pr branch and the files under src/sass/converted are no longer deleted

jeff-phillips-18 (approved):

Issue #663, opened 2017-06-12
alignment of kebab dropdown menu callout


jgiardino:
The callout (i.e. triangle pointing up) of the dropdown menu is not aligned with the kebab icon when the following conditions are true:

- kebab comes after a button
- buttons and kebab are not in a toolbar
- menu is NOT right-aligned (e.g. `ul.dropdown-menu` does not include the class `dropdown-menu-right`)

An example of this implementation can be seen in the [PF-React repo storybook for the kebab](https://jtomasek.github.io/patternfly-react/?knob-Label=Danger%20Will%20Robinson%21&knob-Right%20aligned=false&selectedKind=DropdownKebab&selectedStory=kebab&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-knobs)

andresgalante:
@jgiardino Do you have time to send a PR to pf to fix this?

PR #835, opened 2017-11-09
fix(kebab): adjust position of kebab callout

jgiardino:
Fixes an issue that occurs when a combination of buttons and kebab display outside the context of a
toolbar and the menu is left-aligned, so that the callout of the menu is better aligned with the
kebab.

#663 
Jira story: https://patternfly.atlassian.net/browse/PTNFLY-2605

## Changes

These changes are based on the following use cases: 

* A combination of buttons and kebab could display in other places besides a list view row or a toolbar
* The html used in the toolbar for the combination of buttons and a kebab would be reused when displaying a combination of buttons and kebab in other locations.

To support these use cases, the following changes were made:

* Move the css that correctly renders the combination of buttons and kebab on the toolbar from toolbar.less to dropdowns.less
* Include an example of a combination of buttons and kebab on the Dropdowns test page.

Nope.

## PR checklist (if relevant)

### DIFF
--- a/src/less/dropdowns.less
+++ b/src/less/dropdowns.less
@@ -170,6 +170,7 @@
 
 // Kebab dropmenu
 .dropdown-kebab-pf {
+  &.btn-group > .btn:first-child,
   .btn-link {
     color: @gray-darker;
     font-size: (@font-size-base + 4);
@@ -183,6 +184,7 @@
       color: @link-color;
     }
   }
+  &.btn-group { margin-left: (@grid-gutter-width/4); }
   .dropdown-menu {
     left: -15px;
     margin-top: 11px;
--- a/src/less/toolbar.less
+++ b/src/less/toolbar.less
@@ -37,11 +37,6 @@
       &:focus,
       &:hover { color: @link-color; }
     }
-    .dropdown-kebab-pf .btn-link {
-        padding: 4px (@grid-gutter-width/4);
-        margin-left: (@grid-gutter-width/(-4));
-        margin-right: (@grid-gutter-width/(-4));
-    }
   }
 }
 .toolbar-pf-actions {
--- a/src/less/toolbar.less
+++ b/src/less/toolbar.less
@@ -37,6 +37,11 @@
       &:focus,
       &:hover { color: @link-color; }
     }
+    .dropdown-kebab-pf .btn-link {
+      padding: 4px (@grid-gutter-width/4);
+      margin-left: (@grid-gutter-width/(-4));
+      margin-right: (@grid-gutter-width/(-4));
+    }
   }
 }
 .toolbar-pf-actions {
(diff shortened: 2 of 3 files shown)


jgiardino:
@dgutride @dlabrecq @jeff-phillips-18 @andresgalante 
Could you review this one and let me know if you see any issues with these changes? I removed css from toolbar.less and included it in dropdowns.less instead, so that the combination of buttons and kebabs would have a similar layout whether inside a toolbar or not.

src/less/dropdowns.less
```diff
@@ -183,6 +184,7 @@
       color: @link-color;
     }
   }
+  &.btn-group { margin-left: (@grid-gutter-width/4); }
```
dlabrecq, line 187: I downloaded the source and ran the test page, which looks fine. However, I'm looking at Angular Patternfly and patternfly-ng and don't see "btn-group" as a position for the kebab. 

The test page uses dropmenuPosition="btn-group", but the only thing in the Angular-based toolbar that uses "btn-group" is the sort menu. The patternfly-ng kebab appears to be using pull-right instead, while Angular Patternfly doesn't use a position?.

It's likely that the Angular-based components are wrong, but this change may break them?

jgiardino: Thanks for reviewing, @dlabrecq! 

I did a quick review of the angular-patternfly toolbar, and like you said, the html and classes applied to the kebab in the toolbar are slightly different than in core patternfly. 

It does appear that the layout would be negatively impacted by the css I pulled out of toolbar.less. I had pulled css out of toolbar.less and moved it to dropdowns.less to make the css more generally available outside the toolbar and less specific to the toolbar. But the angular toolbar is dependent on how the css selectors are written in the toolbar.  

I'll add the css back to toolbar.less and check other angular components for any other instances of the kebab that exist, now that I know that they're a little different. ;-)

dlabrecq: I think it would probably be fine if you referenced btn-link instead of btn-group?

andresgalante: Hi @jgiardino @dlabrecq is this PR ok to go, or does it need further adjustments?

dlabrecq: Were changes were made? I only see one commit, but it could have been squashed?

dlabrecq (dismissed):

jgiardino:
No, I need to make updates to this PR. But I'm still horrible at git and am not entirely certain how to make updates and submit as a single commit so I've been waiting for a day when I have the time to figure that out.

dlabrecq:
Just make a new commit with your changes and push to the same 'jgiardino-kebab-callout' branch. GitHub will automatically show two commits for this PR.

jgiardino:
Oh, ouch. This sounds so harsh:
>  jgiardino dismissed dlabrecq’s stale review via 4b1c686 just now

I was under the impression that I could only have one commit per PR. But it sounds like that isn't the case.

I made the updates that this needed. @dlabrecq Could you take one more look, and let me know if there's something that I missed.

Thank you!

dlabrecq:
All that message means is when you push new changes, GitHub requires a new review before the PR can be merged.

dlabrecq (approved):
LGTM

matthewcarleton:
@dgutride @jeff-phillips-18 can we get one more +1 on this to get it merged :)

jeff-phillips-18 (approved):

matthewcarleton (approved):
👍

Issue #832, opened 2017-11-09
Rubygems and rails asset pipeline support in the SASS version


skateman:
The new native sass doesn't have support for rubygems and the rails asset pipeline. This will require some changes in the conversion:
* Provide a top-level `_patternfly-sprockets.scss` similarly to the [old code](https://github.com/patternfly/patternfly-sass/blob/master/assets/stylesheets/_patternfly-sprockets.scss) mimicking [bootstrap-sass](https://github.com/twbs/bootstrap-sass/blob/master/assets/stylesheets/_bootstrap-sprockets.scss)
* Update the variables that are [referring to a path](https://github.com/patternfly/patternfly-sass/blob/master/assets/stylesheets/patternfly/_variables.scss#L42)
* Update each [path for files](https://github.com/patternfly/patternfly-sass/blob/master/assets/stylesheets/patternfly/_fonts.scss#L11) referenced in SASS
* Add the gemspec and rails engine files

TheRealJon:
Will be closed by PR #840

Issue #834, opened 2017-11-09
Importing bootstrap in the SASS code should not have the `-sass` suffix


skateman:
I noticed that when you're importing bootstrap-related things in the top level SASS file, you're referring to as `bootstrap-sass`. However, in the SASS world the library is called `bootstrap` so we should import it like that.

TheRealJon:
Will be closed by PR #840

Issue #833, opened 2017-11-09
Directory structure in the SASS folder needs a revisit


skateman:
The files and folders in the `dist/sass` folder should look more similar to bootstrap-sass:
* There should be a top level file called `_patternfly.scss` without the `-sass` suffix
* The `partials` folder should be renamed to `patternfly`

TheRealJon:
Will be closed by PR #840

PR #840, opened 2017-11-13
fix(sass): Fix issues #832, #833, #834, and #848

TheRealJon:
## Description
Updated dist file structure to more closely match the original patternfly-sass package. Added Sprockets/Compass path helpers and updated path references in Sass code to use the helpers if they are present. Fix Patternfly variable declarations to use `default!` flag which makes it easier for downstream projects to override them. 

Closes issue #832, #833, #834, and #848.

## Changes

* Break copy task down into fonts, images, less, sass, and js for ease of reference and reusability.
* No longer copy dependencies that provide native sass. These are imported directly from the package instead.
* Remame sass `partials` folders to `patternfly`
* Change `$font-path` and `$img-path` to include trailing slash to be consistent with bootstrap and font-awesome conventions.
* Update relative file paths in Sass code to use Sprockets/Compass path helpers when present.
* Add sass sprockets/compass helper flag to `variables.less` so that a default `false` value is present in the Sass. The variable will be unused in Less, but eliminates the need to prepend the declaration during the conversion process.
* Add font-awesome-sass ^4.7.0 dependency
* Move global variable declarations into `variables.scss` so that they are all centrally located.
* Add `default!` flag to global variable declarations so that they can be overridden by setting their value prior to importing Patternfly Sass.
* Consolidate patternfly.scss and patternfly-additions.scss into _patternfly.scss
* Add build.scss to `src/sass` so that sass can be compiled for testing. The `_patternfly.scss` file can't be compiled because Sass and the grunt-sass package ignore partials (any file name starting with `_`).
* Add less to sass conversion to the main build task so that `dist/sass` is always in sync with current less and dependencies are converted and populated into the appropriate location.
* Move converted dependencies into `dist/sass/patternfly/dependencies`
* Change import paths for converted dependencies to relative paths
* Include sass in default build so that it is executed as part of CI
* Include all sass dependencies (other than bootstrap-sass and font-awesome-sass) in the distribution.
* Fix minor less to sass conversion inconsistency with ampersand interpolation.

## PR checklist (if relevant)

- [x] **Less Visual Regression Tests**: Passes all visual regression tests.
- [x] **Sass Conversion**: Less to Sass conversion step has been completed. 
- [x] **Sass Visual Regression Tests**: Passes all visual regression tests.


skateman:
Almost good, I found two things:
* the top level `patternfly.scss` should be also a partial `_.patternfly.scss` as it will be possibly imported in other projects
* do we really need the `patternfly-additions.scss`, I think the IE9 CSS bug should be fixed on the CSS level and not in SASS, @jeff-phillips-18 and I agreed on merging it with the top level file

TheRealJon:
@skateman I have addressed the issues with the top level broker file.

TheRealJon:
@skateman @jeff-phillips-18 I added the less to sass conversion step to our build. This is necessary so that the dependencies mentioned before get converted and placed in the dist folder. It will also ensure that the sass in dist is always up to date with the less.

TheRealJon:
Dependencies are now in `dist/sass/patternfly/dependencies` and `_patternfly.scss` uses relative paths to import them.

skateman:
Starting to look good, I tried to test with the following command:
```sh
sass dist/sass/_patternfly.scss -I node_modules/bootstrap-sass/assets/stylesheets -I node_modules/font-awesome-sass/assets/stylesheets -I node_modules/
```

It returned:
```
Error: File to import not found or unreadable: bootstrap-touchspin/dist/jquery.bootstrap-touchspin.
```

The problem is that SASS assumes the `.scss` suffix at imports without file a extension. After a quick check I found out that the same error occurs for `c3`. You need to specify the `.css` suffix explicitly OR copy/rename the file to the `dependencies` folder with a `.scss` extension. As there is no internal rubygem support for these packages, I'd suggest the copy/rename solution for all these dependencies (even the ones that have SASS).

skateman:
After resolving the issues, I got the following error:
```
Error: Invalid CSS after "...ropdown-toggle.": expected class name, was ".btn-danger"
        on line 17 of dist/sass/patternfly/_mixins.scss, in `button-variant'
        from line 29 of dist/sass/patternfly/_buttons.scss
        from line 86 of dist/sass/_patternfly.scss
```
Shouldn't be these issues caught on the CI?

TheRealJon:
@dgutride @jeff-phillips-18 @dlabrecq What do you all think about adding a Sass build step to CI so that we can catch Sass compilation errors? I had two ideas:
1. Build from sass first, clean the dist/css folder, then build from less.
2. Build from sass to a separate folder then clean it up.

skateman:
@TheRealJon or you can have a [build matrix](https://docs.travis-ci.com/user/customizing-the-build/#Build-Matrix) and have the whole SASS build as a separate task triggered by a special ENV var. We use this approach in [ManageIQ](https://travis-ci.org/ManageIQ/manageiq-ui-classic/builds/308124866) to run different sets of tests against multiple ruby versions.

martinpovolny:
@TheRealJon : I would like to remind you that ManageIQ is blocked on old version of PF for minimally 6 months.

In that light I don't understand why you started this job by rewriting what worked from Ruby to Javascript.

What is the expected date of the new PF with SASS to be relased?

What is the rationale of prioritizing removing Ruby over actually cutting a release?

Thank you!

jeff-phillips-18:
@TheRealJon I think option 2 makes more sense and seems reasonable.

@martinpovolny We are adding support of SASS to patternfly. Previously this was done by an external repository not created or maintained by the core patternfly team. We have to do this in a  manner that is maintainable by the team we have. We have a few minor hurdles to get over to complete this effort. Once this PR is completed, we will release another Beta version and if all goes well a full release of PF w/ SASS support will be done.

serenamarie125:
@martinpovolny we are aware that you are blocked, thus we have assigned resources to work on this effort.

TheRealJon:
@serenamarie125 @dgutride @jeff-phillips-18 @skateman 

I just pushed up a few changes to address the last few issues that @skateman mentioned. @skateman let me know if this resolves everything and if there is anything else we need to address before merging this PR.

jeff-phillips-18 (approved):

skateman:
Tested against ManageIQ and it seems working, however, there is a deprecation warning. New SASS versions will support percentage values in `0.xx` form and the support for `xx%` will be dropped. I noticed that these are being used inconsistently and I think they should be unified.

I created a [PR against your](https://github.com/TheRealJon/patternfly/pull/1) repo where I'm adding the files required for the rubygem support. It should automatically parse the package version from the `package.json`. Releasing can be automated using [Travis deployment](https://docs.travis-ci.com/user/deployment/rubygems/) triggered by version tagging.

dgutride (approved):

dgutride:
@skateman - https://github.com/patternfly/patternfly/issues/852 created to track deprecation warning

PR #851, opened 2017-11-24
fix(wizard): robust mobile wizard

matthewcarleton:
## Description
This is a more robust version of the current wizard for mobile support. The old version is still supported so I've created an alternative rather than replace the current one. The existing wizard is now marked as wizard(old).  

This version still needs js consideration and I would like to talk about the multiple implementations of the wizard code. 

https://rawgit.com/matthewcarleton/patternfly/advanced-wizard-dist/dist/tests/wizard.html

## PR checklist (if relevant)


andresgalante:
@LHinson are you ok with renaming the old wizard to Wizard(old) and introducing the new one as the recommended version?

andresgalante:
@matthewcarleton there is a problem with the line on the left, it moves and shifts depending on the viewport size

img

That must be due to the subpixel position of the modal, this would not happen if the wizard was outside a modal.

matthewcarleton:
@andresgalante is it actually showing up on different mobile sizes or is that just when you resize your browser window?

andresgalante:
@matthewcarleton when hove over the subitems, they shift a bit to the right to accommodate for the space of the border. Can you add a transparent border to them and them add a color on hover/active please?

matthewcarleton:
@andresgalante good catch. Done :)

andresgalante (dismissed):
@LHinson @jeff-phillips-18 @dlabrecq @dgutride @priley86:

I am ok to merge this one, but I'd like for all of us to be aware of what this PR is doing:

Due to how the design was done, this is not a responsive component. It's adding a mobile version for the steps. Meaning that we'll force to have duplicated information in the DOM and showing or hiding parts according to the viewport.

We have the same case on the vertical nav.

Both designs should be corrected for the next version of patternfly were no component will be merge unless it's responsive.

priley86:
No issues on this PR. Please nudge me though once this is merged please. ⏰

jeff-phillips-18:
@andresgalante Would it be better to take the time to make this a 'responsive component' now? Would doing so break the existing wizard usage?

matthewcarleton:
@jeff-phillips-18 @andres is on vacation, but this might be a question for @LHinson. I believe we would have to revisit design to understand how this translate to desktop, which would add more time. I think the idea is we are going to correct these things in PF-next.

jeff-phillips-18 (dismissed):

LHinson (dismissed):

PR #854, opened 2017-11-29
fix(sass): Fix relative path variables in less and sass

TheRealJon:
## Description
PR #840 introduced a breaking change related to relative paths defined by Less variables. This change was reverted and the Sass conversion rules were adjusted to ensure the necessary changes are only present in Sass.

### Less
* Revert changes to `@font-path`, `@img-path`, and `@icon-font-path` variables in Less
* Revert changes to Less where forward slashes were removed from url's using the above variables.

### Sass
* Add Sass conversion rule that appends a forward slash to these variables as needed.
* Adjust Sass conversion rule for urls that use these Sass variables to remove extra forward slash.
* Adjust Sass conversion rule to add conditional Compass/Sprockets asset helper functions in `url()` statements.

## PR checklist

- [X] **Cross browser**: works in Chrome
- [X] **Less Visual Regression Tests**: Less changes pass all relevant visual regression tests
- [X] **Sass Conversion**: Sass conversion step has been completed and committed
- [X] **Sass Visual Regression Tests**: Sass changes pass all relevant visual regression tests

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -369,8 +369,14 @@ module.exports = function (grunt) {
             },
             {
               //
-              pattern: /^\$(img|font)-path:(\s*)(".*")(.*);$/mgi,
-              replacement: '\$$$1-path:$2if($pf-sass-asset-helper, "", $3)$4;',
+              pattern: /^\$(img|font)-path:(\s*)"(.*)"(.*);(.*)$/mgi,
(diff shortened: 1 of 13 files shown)


serenamarie125:
fyi @ohadlevy

dlabrecq (commented):
LGTM. Although, we should probably test with RCUE as we rebuild the patternfly CSS.

jeff-phillips-18 (approved):

dgutride (approved):

PR #880, opened 2017-12-08
fix(login): Replaces invalid fade for valid sass mixin, rbga

### LINKED ISSUE #879: [SASS] Invalid sass mixin breaks login page, container background-color absent
`fade($color-pf-white, 5.5%)` [is not a valid option](http://sass-lang.com/documentation/Sass/Script/Functions.html), recommend using `rgba($color-pf-white, .055)`

[Offending code is here](https://github.com/patternfly/patternfly/blob/master/src/sass/converted/patternfly/_variables.scss#L88)

AllenBW:
closes #879

this is for da sass

OK SO less variable in place calculates to: `background-color: rgba(255,255,255,.055);`
This work calculates to: `background-color: rgba(255, 255, 255, 0.055);`

close enough right?  😏 

🔪 🍍

### DIFF
--- a/src/sass/converted/patternfly/_variables.scss
+++ b/src/sass/converted/patternfly/_variables.scss
@@ -85,7 +85,7 @@ $list-view-divider:                                                 $color-pf-bl
 $list-view-hover-bg:                                                $color-pf-blue-25 !default;
 $list-group-top-border:                                             $color-pf-black-200 !default;
 $login-bg-color:                                                    $color-pf-black !default;
-$login-container-bg-color-rgba:                                     fade($color-pf-white, 5.5%) !default;
+$login-container-bg-color-rgba:                                     fade_out($color-pf-white, 0.945) !default;
 $modal-about-pf-bg-img:                                             "bg-modal-about-pf.png" !default;
 $modal-title-padding-horizontal:                                    18px !default;
 $modal-title-padding-vertical:                                      10px !default;


AllenBW:
@jeff-phillips-18 goooooood for your review (and all others) at your earliest 🙇‍♀️

TheRealJon:
@AllenBW We will have to add a conversion rule for this. In the Gruntfile, there is a 'lassToSass' config section with a bunch of custom rules defined. A rule will need to be added there to convert less `fade` mixins to Sass `fade_out`. Manually changing `_variables.scss` will not fix the issue because the next time a build happens, your changes will be overwritten.

AllenBW:
@TheRealJon awww 💩 thanks for the heads up, will look into that, throw the work here

AllenBW:
Okie dokie, @TheRealJon for ease of conversion, looks like we should a conversion target of`rgba($color-pf-white, .055)`  (yields the same result as, minimal changes) so swapping `fade` with `rgba` and the decimal with a numerical equivalent, I looking at the config file, this'll take a bitta time to figure out, any chance you could knock it out real quick? (if not, no worries)

cliffpyles:
@AllenBW Did you mean to use `fade-out`? Were you meaning to use the SASS function?

http://sass-lang.com/documentation/Sass/Script/Functions.html#transparentize-instance_method

AllenBW:
@cliffpyles so we CAN use `fade-out` but it will require a recalculation of the percentage....  

`fade($color-pf-white, 5.5%)`👉 ` fade_out($color-pf-white, 0.945)`
alternatively...
`fade($color-pf-white, 5.5%)`👉 `rgba($color-pf-white, .055)`


that second one looks a bit easier, cuz all ya gotta do is swap `fade` for `rgba` and 5.5% into a number, .055

cliffpyles:
Nevermind. Looks like they both work. Nice Work.


https://codepen.io/cliffpyles/pen/KyOzzW

TheRealJon:
So this issue was a little bit bigger than just the one variable. A conversion rule already exists in the grunt tool that I am using which converts less fade to sass rgba. The problem was that it does not handle decimal percentages in the second argument of the fade function (the regex used to find fade functions does not match when a decimal is used). This caused any instance of the less fade function with decimal percentages to be skipped in the conversion (not just the one variable). I am working on a fix for this now.

AllenBW:
@TheRealJon oh oh flippin awesome! by chance can I burden you for a where? 

❤️ 💌

TheRealJon:
Closing this PR. Issue resolved by PR #883.

PR #883, opened 2017-12-11
fix(sass): Fix conversion rule for fade() to rgba()

TheRealJon:
## Description
Add new custom Sass conversion rule that converts less fade() function with equivalent sass rgba() function, and properly handles decimal percentages.

Fixes Issue #879, replaces PR #880

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -384,7 +384,7 @@ module.exports = function (grunt) {
           }
         ],
         options: {
-          excludes: ['variables', 'default'],
+          excludes: ['variables', 'default', 'rgba'],
           replacements: [
             {
               // Customize variable conversion to include newer css reserved words.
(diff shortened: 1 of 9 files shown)


Gruntfile.js
```diff
@@ -416,6 +416,12 @@ module.exports = function (grunt) {
               replacement: '$1@include $2()',
               order: 3
             },
+            // Original fade -> rgba conversion did not account for decimal percentages
+            {
+              pattern: /fade\((.*),\s?([\d\.]+)\%\)/gmi,
+              replacement: 'rgba($1, ($2/100))',
```
jeff-phillips-18, line 422: Will we still handle fade that does not use a percentage?

AllenBW (approved):
Thanks for the quick fix!!

Gruntfile.js
TheRealJon: @jeff-phillips-18 The less fade function signature requires a percentage, so this shouldn't be an issue.

jeff-phillips-18: Ok great.

jeff-phillips-18 (approved):

PR #884, opened 2017-12-12
fix(filter): Use Datatables column().search() API to allow to filter ajax data

### LINKED ISSUE #670: Datatable filter with Ajax data source
I am using the table view with the pfFilter extension. My data is sourced via an ajax call and filtered on the server via DataTables's server side rendering. DataTables handles filtering in this case by setting headers on the request.

Is it possible to make patternfly filters set these headers?

tuxgasy:
## Description
Call Datatables column().search() API before redraw the table.

Fixes: https://github.com/patternfly/patternfly/issues/670

### DIFF
--- a/src/js/patternfly.dataTables.pfFilter.js
+++ b/src/js/patternfly.dataTables.pfFilter.js
@@ -213,6 +213,36 @@
   // Local functions
 
   /**
+   * update filters if server-side processing is enabled
+   * see more https://www.datatables.net/release-datatables/examples/data_sources/server_side.html
+   *
+   * @param {DataTable.Api} dt DataTable
+   * @private
+   */
+  function updateRemoteFilters (dt) {
+    var ctx = dt.settings()[0];
+    var filters = [];
+
+    $.each(ctx._pfFilter.filters, function (index, filter) {
+      if (filters[filter.column] === undefined) {
+        filters[filter.column] = [];
+      }
+
+      filters[filter.column].push(filter.value);
+    });
+
+    dt.search('').columns().search(''); // clear before update
+
+    $.each(filters, function (column, values) {
+      if (values === undefined) {
+        return;
+      }
+
+      dt.column(column).search(values.join('|'), true, false, ctx._pfFilter.filterCaseInsensitive);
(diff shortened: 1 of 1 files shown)


dlabrecq (dismissed):
I built and ran the source locally. Filtering seems to work fine with these changes. It still filters client-side data by lowercase.

Update: need to retest latest changes

jeff-phillips-18:
Please squash commits and use semantic release commit formatting: https://github.com/patternfly/patternfly#git-commit-guidelines

andresgalante:
Hi @jeff-phillips-18 when you have a min, can you please check if we can merge this one?

jeff-phillips-18:
nit: Can we change the commit (and PR title) to Datatable rather than filter?

AdamJ:
@jeff-phillips-18 can this PR be merged, or does the commit message need to be altered?

jeff-phillips-18:
we could alter the commit message on merge once this is approved if @tuxgasy does not update it.

dlabrecq:
I retested the latest changes an there appears to be something wrong. I recall the first iteration of this PR worked when tested, but the commits have been squashed and don't remember why it changed?

The table examples are not showing any rows in the test pages.
screen shot 2018-04-30 at 10 20 03 am

dlabrecq (changes requested):
I retested the latest changes an there appears to be something wrong. 

I recall the first iteration of this PR worked when tested, but the commits have been squashed and don't remember why it changed?

The table examples are not showing any rows in the test pages.
screen shot 2018-04-30 at 10 20 03 am

tuxgasy:
I realized that my PR did not solve the problem completely. If I apply filters on multiple columns, only the last filter is send to the server.

Sorry for my late response!

LHinson:
No worries @tuxgasy. Things get busy! Should I close this PR for now or do you expect to send more updates to address the issue?

tuxgasy:
I do not know how to implement this properly. So I think we can close this PR!

LHinson:
ok thanks for getting back to me @tuxgasy

PR #885, opened 2017-12-12
feat(login page): basic, social, SSO logins

matthewcarleton:
## Description
Updates login to show new login screens with options for basic, SSO and social accounts.

There are slight differences with the spacing for the design but overall very close. This was to ensure a strong base that could be used to implement all the different concepts in the inVision link below.

https://redhat.invisionapp.com/share/2DE6FA656

Here are the test pages:
https://rawgit.com/matthewcarleton/patternfly/login-responsive-dist/dist/tests/login-single-sign-on.html
https://rawgit.com/matthewcarleton/patternfly/login-responsive-dist/dist/tests/login-basic.html
https://rawgit.com/matthewcarleton/patternfly/login-responsive-dist/dist/tests/login-social-account.html
https://rawgit.com/matthewcarleton/patternfly/login-responsive-dist/dist/tests/login-social-account-two-column.html

I've also updated the design doc to include branding guidelines for each logo used: 
https://github.com/patternfly/patternfly-design/pull/512
## PR checklist


src/less/login.less
```diff
@@ -94,3 +94,209 @@
...
+    }
+  }
+  .card-pf {
+    padding: 30px 40px;
+    p {
+      color: @color-pf-black-600;
+    }
+  }
+  .form-control {
+    height: 36px;
```
jgiardino: bootstrap initially defines the height of `.form-control` using this:
`  height: @input-height-base; // Make inputs at least the height of their button counterpart (base line-height + padding + border)`

Should the height in this case also be based on variables? (I'm relying on Sarah to actually follow up with a suggestion on what that would actually be ;-)

src/less/login.less
```diff
@@ -94,3 +94,209 @@
...
+        bottom: 0;
+        left: 50%;
+        transform: translateX(-50%);
+        z-index: 0;
+      }
+      p {
+        position: absolute;
+        z-index: 1;
+        width: 30px;
+        height: 30px;
```
jgiardino: I'm curious how this would work if the translated word for "or" is longer than 30 pixels. Do you think we need to specify height and width here? When I disable it in the inspector, I don't see a change.

Also, the "or" separator is removed in the smaller width layout. Is that intentional in the design?

jgiardino: Oh about that last comment, I found the responsive design example in the invision prototype. It looks like "or" was added by the visual team, so I'm curious if they would include that in the responsive layout. I really like the use of "or" because it makes it obvious that they're two different options/methods, and it might be nice to include that in the responsive layout as well.

jgiardino:
bitmoji
These screens are really nice, @matthewcarleton! 

I included a couple of inline comments about fixed sizes. There are other places with fixed font sizes that I didn't comment on, but I wonder if those should use variable sizes too?

mcarrano (changes requested):
@matthewcarleton These look great. Just a couple of questions/comments.

* I noticed that you left out the PatternFly logotype at the top of the page. Should we keep this here even though it's really a placeholder? It just gives a better sense of what the page in intended to look like.

* I had suggested that the language selector be moved from the footer to the top of the login card, See the updated design here: https://redhat.invisionapp.com/d/main/#/console/12118257/255230615/preview  It seemed off that this would be at the bottom since I'm assuming changing the language will refresh this screen with translated content.

matthewcarleton:
@mcarrano hey Matt,
Thanks for pointing that out, the logo is actually there I'm just having path issues when it gets built. I'll fix that.

I will adjust the language selection as well.

matthewcarleton:
@jgiardino thanks Jenn! Ya you caught me, I meant to go back and update with variables for that stuff and forgot. I'll adjust :)

I think I can get rid of width/height and border radius. I'm not sure why I thought that needed to be a circle. That should deal with the longer string of text flexibility.

@mcarrano what are your thoughts on keeping the "or" in the social login that @jgiardino points out above. I agree it could be useful for all layouts.

mcarrano:
@matthewcarleton @jgiardino Agree that it would be nice to consistently use the "or" separator on social login whether in desktop or mobile layout. But how to you propose that this would respond? Would the vertical separator change to horizontal? @kybaker would also be good to get your opinion.

matthewcarleton:
@kybaker @jgiardino @mcarrano something like this is what I imagine.
screen shot 2017-12-13 at 11 56 19 am

jgiardino:
I always go back and replace fixed sizes with variables last. :-) I figured that's what you were doing as well

mcarrano:
Looks good to me @matthewcarleton . Just a small nit, but shouldn't this read, "Log in using a social account"?

matthewcarleton:
@mcarrano ok thanks. I went back to look at the design for the social logins and noticed now the titles have been removed altogether. Is that intentional?

mcarrano:
@yih-wang Please see @matthewcarleton 's question. Seems like there are differences in the InVision deck between the wireframes and the hi-res mockups. In this case, the specific question includes the instructional text. Was that intentionally removed?

yih-wang:
@matthewcarleton @mcarrano  We just had a meeting yesterday with Keycloak team and some minor updates had been made. I just stick notes on wireframes to clarify which places have been modified (visual design haven't been updated yet) and will keep updating the notes if further updates are made. 

Titles are not removed but have some adjustment. "Log in using your social account" has been modified to "Log in with" because IdPs buttons can also be displayed.

And we also decided to remove the "or" in the middle of the divider line. Because it can be different numbers of characters in different language. See the wireframes to find the latest design. Thanks.

matthewcarleton:
@yih-wang thanks for the update. So if the "or" is being removed is the design remaining the same or have you adjusted the spacing? Also, should it be "Log in with" OR "Log in with :"? @mcarrano what do you think?

mcarrano:
@matthewcarleton Just using "Log in with" is fine. We typically are not using semi-colons at the end of labels.

@yih-wang Thanks for adding the notes into the InVision deck. I reviewed the latest designs and left some comments. Do you know if we can expect further changes? I had thought that the designs were fairly solid, but if we are still making changes then it seems like a moving target. Based on what I saw, the basic and social login look pretty good ( but see my question about the 'More' button for social login). The identity first and social login seem to have more questions. Please do keep us up to date on any changes that you are making.

junezhang:
@matthewcarleton, thanks for your efficient work.
@mcarrano Currently, Basic Login / SSO Login / Social Login, these three login flows almost have been confirmed. 
Our next target is Identity First & other login method design review, and Mobile Re-Design.

zhutaoredhat:
Hi @matthewcarleton After we remove the "or", the padding will change a little. It will look like the mockup below.
padding

These screens look great! In most cases, we want to display all the content in one screen, so there may be some minor adjustments to padding and margin. I'll add marking and discuss with you later. Great thanks ~

yih-wang:
Thanks @matthewcarleton for the great work! @mcarrano Please see @junezhang 's comment to get a general view of our progress. I think most designs are already determined. But there may be great adjustment to the responsive state design. The latest version is already posted to InVision and waiting for the review of Keycloak team. @zhutaoredhat and I will post comments here once any further updates have been made.

src/less/login.less
```diff
@@ -94,3 +94,214 @@
...
+  .login-pf-social-accounts {
+    display: flex;
+    list-style: none;
+    padding: 0;
+    margin: 0;
+    flex-wrap: wrap;
+    //justify-content: space-between; if two columns are required use this
+    &-link {
+      flex-basis: 100%;
+      //flex-basis: 45%; if two columns are required use this
```
srambach: I'm a little confused by this comment and the one just above - is it work in progress or does the consumer of PF have to change this if they are using two columns? Should there be a two-column variant?

src/less/login.less
srambach: Bootstrap uses `@input-height-base: (@line-height-computed + (@padding-base-vertical * 2) + 2);` so you could replicate that. I'm not sure what the +2 is for though.

src/less/login.less
```diff
@@ -94,3 +94,214 @@
...
   }
 }
+
+//
+// Login
+// The following styles are for the new login
+// --------------------------------------------------
+.login-pf-page {
+  .login-pf-brand {
+    margin: 20px 0 10px;
```
srambach: Throughout there are a lot of magic numbers and I see where it's difficult to know whether to make them variables for the one time use. Some maybe would be a fraction of @grid-gutter-width? Others I'm just not sure where they are coming from, whether the design is using "standard" PF spacings as much as we have them, or diverging for this special case.

srambach:
It looks beautiful and the responsive behavior is really nice. I left a couple of comments about fixed sizes, but your code is great and easy to read! Nice job.

src/less/login.less
matthewcarleton: This is where I got stuck, the design calls for a two column break if there are more than four items but one column if there are only 4. Changing the flex-basis will get you a two column layout. I was thinking the smallest change was the best. I commented in the html that the change was available in the CSS. Although, the design has now changed so I could change this back to bootstrap grid and just leave it at that.

src/less/login.less
matthewcarleton: ya, I don't know where the values are coming from. I didn't know if there was much value in creating variables. Are you thinking it would be good to at least use @grid-gutter-width?

src/less/login.less
srambach: What about doing another class that is a variant of the .login-pf-social-accounts-link so that the HTML can just have that class or not, and the CSS doesn't have to change?

src/less/login.less
srambach: If that's why it's spaced that way, then yeah. Maybe the designer can tell you that?

src/less/login.less
jgiardino: I like that approach. You can default to 100% but then have an additional class that switches it to 45%.

mcarrano (dismissed):

matthewcarleton:
@srambach @jgiardino @mcarrano This PR has been updated to reflect the requested changes. Thanks for all the help reviewing @srambach & @jgiardino, @mcarrano we should be good to merge now.

mcarrano:
@matthewcarleton Thanks for all your hard work on this! @srambach and @jgiardino are you OK to merge? We'd like to make these changes available to the Keycloak team prior to shutdown. @amysueg FYI

srambach:
LGTM 👍

andresgalante:
@matthewcarleton great job! 
The name of some classes are not great and there are some things that can probably be achieved with bootstrap structures.

Let's pair code it and merge it when you are back

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
 }
+
+//
+// Login
+// The following styles are for the new login
+// --------------------------------------------------
+.login-pf-page {
+  .login-pf-brand {
+    margin-top:ceil((@grid-gutter-width / 2));
+    width: 70%;
```
andresgalante: Why do you need a `width` here?

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+  }
+  .login-pf-page-header {
+    margin-bottom: ceil((@grid-gutter-width / 2));
+      @media (min-width: @screen-sm-min) {
+        margin-bottom: @grid-gutter-width;
+      }
+    text-align: center;
+    p {
+      color: @color-pf-white;
+      font-weight: 700px;
```
andresgalante: this doesn't exists.

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+      font-size: @font-size-large;
+      max-width: 500px;
```
andresgalante: I don't think that the `max-width` here is needed.

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+    }
+    p {
+      color: @color-pf-black-600;
+    }
+  }
+  .form-control {
+    height: ceil((@input-height-base + 10px ));
+  }
+  .checkbox-label {
+    color: @color-pf-black-600;
```
andresgalante, line 139: I wouldn't change the color here, the form should be a normal pf form.

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+    font-weight: 300;
```
andresgalante, line 140: neither the font weight. If there is a need for a font weight change, it should be global in my opinion.

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+    margin-bottom: 15px;
+    @media (min-width: @screen-sm-min) {
+      margin: 0;
+    }
+  }
+  .btn-primary {
+    margin: @grid-gutter-width 0 0;
+  }
+  .login-pf-header {
```
andresgalante: classnames are so confusing sometimes :(

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+    margin: 0 0 ceil((@grid-gutter-width / 2));
+    @media (min-width: @screen-sm-min) {
+      margin: 0 0 ceil((@grid-gutter-width * .75));
+    }
+    display:flex;
```
andresgalante: Why are we flexing this, the normal direction of the elements are `column` anyway.

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+      text-align: center;
+    }
+    //this is ugly but neccessary to get the desired style :(
+    .bootstrap-select:not([class*=col-]):not([class*=form-control]):not(.input-group-btn) {
+      width: auto;
+      align-self: flex-end;
+      .dropdown-toggle {
+        color: @color-pf-black-700;
+        background: none;
+        border: none;
```
andresgalante: The drop menu needs a focus ring, you can tab on it and open it, but since the border is none there is no visual indication.

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+  }
+  .login-pf-signup {
+    margin: @grid-gutter-width 0 0;
+    font-size: ceil((@font-size-base * 1.25));
+    text-align: center;
+    a {
+      margin-left: 10px;
+    }
+  }
+  .login-pf-settings {
```
andresgalante: This class name is really bad.
If the design must stay as is, then let's think of another name. something related to forms maybe... naming is impossible.

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+    display: flex;
+    justify-content: space-between;
+    flex-wrap: wrap;
+  }
+  .login-pf-page-footer {
```
andresgalante, line 197: Can't we just use `list-inline`?
https://getbootstrap.com/docs/3.3/css/#inline

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+        justify-content: center;
+          li {
+            margin: 0 5px 5px;
+          }
+      }
+    }
+  }//login-pf-footer
+  .login-pf-social {
+    display: flex;
+    justify-content: space-between;
```
andresgalante: if you only have 2 columns and they sit size by side, then I don't think there is a need to `justify-content: space-between` right?

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+  .login-pf-social-section {
+    flex-basis: 100%;
+    padding:0;
+    @media (min-width: @screen-sm-min) {
+    flex:0 1 50%;
+    max-width:50%;// this is for IE10/11 that doesn't consider padding with box-sizing on flex-children.
+    }
+    &-heading {
+      text-align: center;
+      margin: ceil((@grid-gutter-width * .75)) 0;
```
andresgalante: The login page is measuring margins relative to the `@grid-gutter-width`

I see what the intention is to keep consistency. But PF3 has no variable to keep space consistency, and if a user changes the gutter this page will break.

The relationship of these 2 things is by chance and not intentional.

I recommend changing all references to the gutter and creating a new variables.

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+        padding-right: @grid-gutter-width;
+      }
+      &:last-of-type {
+        padding-left: ceil((@grid-gutter-width - 1px));
+        border-left: 1px solid @color-pf-black-300;
+      }
+    }
+  }//social-section
+  .login-pf-social-accounts {
+    display: flex;
```
andresgalante: There is no need to flex this.

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+      &:first-of-type {
+        padding-right: @grid-gutter-width;
+      }
+      &:last-of-type {
+        padding-left: ceil((@grid-gutter-width - 1px));
+        border-left: 1px solid @color-pf-black-300;
+      }
+    }
+  }//social-section
+  .login-pf-social-accounts {
```
andresgalante: you could use `list-unstyled` instead:
https://getbootstrap.com/docs/3.3/css/#unstyled

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+    display: flex;
+    list-style: none;
+    padding: 0;
+    margin: 0;
+    flex-wrap: wrap;
+    @media (min-width: @screen-sm-min) {
+      &-double-col {
```
andresgalante: I couldn't find this on the html. Can you point me out where it's implemented please?

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+        justify-content: space-between;
+        .login-pf-social-accounts-link {
+          flex-basis: 45%;
+        }
+      }
+    }
+    &-link {
+      flex-basis: 100%;
```
andresgalante: do you actually need this? a `li` should be a block level item.

src/less/login.less
```diff
@@ -94,3 +94,210 @@
...
+      &:not(:nth-of-type(4n)) {
+        margin: 0 0 15px;
+      }
+      a {
```
andresgalante: don't flex or justify content the anchor, there is no need, just do `display: block; text-align: center;`

andresgalante (changes requested):
@matthewcarleton overall it's great, I made a few comments and we can pair code it next week. Thera re some indentations and spaces we can fix too.

Thanks a lot!!!

src/less/login.less
matthewcarleton: yup :)

src/less/login.less
matthewcarleton: I'll change, this is left over from the last design  :(

src/less/login.less
matthewcarleton: Ya I talked with Sarah a bit about this b/c I just had fixed values in there. I didn't use a variable at all to begin with, I wonder if we need to create one, will it be used anywhere else?

src/less/login.less
matthewcarleton: yup :)

src/less/login.less
matthewcarleton: The flex is there for when there are two columns. I can remove it and add it to the double column class.

src/less/login.less
matthewcarleton: This is not shown in the build but is in the design.

src/less/login.less
matthewcarleton: nope, good catch this was left over from something else.

src/less/login.less
matthewcarleton: Is there a reason why one is better than the other if they have the same results?

src/less/login.less
matthewcarleton: ugh, removed the px.

src/less/login.less
matthewcarleton: ya good point, I can remove width and leave max-width

src/less/login.less
matthewcarleton: this is just to match the design.

src/less/login.less
matthewcarleton: Yes I would have loved to have you around when I did these, we can talk when I get back and figured out a better naming convention.

src/less/login.less
matthewcarleton: I did this to control the language placement. I believe the default is row isn't it?

src/less/login.less
matthewcarleton: yup good catch.

src/less/login.less
matthewcarleton: :(

src/less/login.less
matthewcarleton: I think list-unstyled works better here, there are spacing issues with list-inline

andresgalante (dismissed):

kybaker:
@matthewcarleton The only feedback I have is could we reduce the size of the logo at the smallest screen size? Other than that looks good.

matthewcarleton:
@kybaker ah yes good catch, @andresgalante that's what that 70% width was for. I'll add it back in unless you can think of a better approach.

matthewcarleton:
@mcarrano this has been updated with the show/hide function. Thanks @jeff-phillips-18 for the help 👍

mcarrano:
The show/hide function looks great. Yes, thanks for your help @jeff-phillips-18 !

mcarrano (approved):
Check all of your screens are they look good to me! Ready to merge. Thanks, @matthewcarleton

srambach:
These look really good. Nice job!

PR #889, opened 2017-12-15
chore(tests): Create scenarios for navigations and enable nav state pages

cliffpyles:
## Description
Create scenarios and reference images for the horizontal and vertical navigations. Enabled the states being demonstrated on the affiliated pages (on page load).

Ticket: https://patternfly.atlassian.net/browse/PTNFLY-2676


jeff-phillips-18 (commented):
Looks good.
Not thrilled with the test pages start with navigation in an invalid state (shown without being hovered), but better to be able to test. Is there any way to add those classes much the way you are removing the alert selectors?

Also, there could be an issue with the collapsed state since we use cookies to determine the initial state. You may need to override the cookie setting.

cliffpyles:
Backstop doesn't have an addSelector property currently. It does have a hover state that should help with some of them though, but the screenshots still weren't showing the expanded state when I used the hover event. I think it has to do with the delay on hover. So I ended up setting the default state in the template. I got the idea from Bulma.

https://bulma.io/documentation/components/dropdown/

I see where the issue with the collapsed state is. I'll make a couple of updates.

cliffpyles:
I'm closing this one out. I found a different approach for the test that doesn't require the states to be set directly in the markup.

https://github.com/patternfly/patternfly/pull/955

PR #890, opened 2017-12-18
fix(sass): Release the gem with the standard Travis way

skateman:
It seems like Travis doesn't like us for [writing an ENV var to a file](https://travis-ci.org/patternfly/patternfly/builds/317710350#L5533). Travis has a standard way of deploying to Rubygems.org. I was not using this before as here we can't control when the deploy script will run. However, it seems like there's no other way...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,6 @@ after_success:
        sh -x ./node_modules/patternfly-eng-release/scripts/semantic-release/_bump.sh -p;
        sh -x ./node_modules/patternfly-eng-release/scripts/semantic-release/_publish-npm.sh || travis_terminate 0;
        npm run semantic-release-post;
-       sh -x ./scripts/gem_release.sh;
        sh -x ./node_modules/patternfly-eng-release/scripts/semantic-release/_publish-webjar.sh -p;
        sh -x ./node_modules/patternfly-eng-release/scripts/semantic-release/_release-all.sh -o;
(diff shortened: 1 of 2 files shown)


jeff-phillips-18 (approved):

.travis.yml
```diff
@@ -45,3 +44,15 @@ after_success:
...
   except:
     - /^v\d+\.\d+\.\d+$/
+
+# Automated RubyGems.org deployment
+before_deploy: "sh -x ./scripts/gem_release.sh"
+deploy:
+  skip_cleanup: true
+  provider: rubygems
+  api_key:
+    secure: RP2UsRAyz1+D5e/Mt4C8c3kfq9nzPunQ3QGSTFMkhCmH0dcyBNqGLaIWbrOtRPIeYXpcneoj4SfBvgdmixhi06NMkjw5Vcn213VUZAdafMthAkB3RaQexuzqyz8F8NVTdtXm15zBz/XDZtm42AkOlotoctNUeChFhyhzq3WG9uw=
```
dlabrecq, line 54: Should this value be moved to the secure envs of Travis CI?

dlabrecq (changes requested):

.travis.yml
skateman: No, it has to be like this, see the PR's description.

dlabrecq (approved):

skateman:
Yay, it [works](https://travis-ci.org/patternfly/patternfly/builds/318160101#L6303) :tada:

PR #906, opened 2018-01-11
fix(secondary-tabs): refactor the secondary tabs into nested structure

### LINKED ISSUE #905: secondary tabs are not nested in primary tabs
Please review https://github.com/patternfly/patternfly-react/pull/112 for details.

dabeng:
## Description
This PR is targeted to fix [issue](https://github.com/patternfly/patternfly/issues/905).

## Changes
Users can toggle different secondary tabs 

Examples: [Tabs](https://rawgit.com/dabeng/patternfly/sec-tabs-dist/dist/tests/tabs.html) under Widgets section

### DIFF
--- a/src/less/tabs.less
+++ b/src/less/tabs.less
@@ -26,7 +26,7 @@
       }
     }
   }
-  + .nav-tabs-pf {
+  + .nav-tabs-pf, &.nav-tabs-pf-secondary {
     font-size: @font-size-base;
     > li:first-child > a {
       padding-left: 15px;
--- a/src/sass/converted/patternfly/_tabs.scss
+++ b/src/sass/converted/patternfly/_tabs.scss
@@ -26,7 +26,7 @@
       }
     }
   }
-  + .nav-tabs-pf {
+  + .nav-tabs-pf, &.nav-tabs-pf-secondary {
     font-size: $font-size-base;
     > li:first-child > a {
       padding-left: 15px;
--- a/tests/pages/tabs.html
+++ b/tests/pages/tabs.html
@@ -5,175 +5,175 @@ title: Tabs
 resource: true
 ---
       <ul class="nav nav-tabs">
-        <li class="active"><a href="#">Home</a></li>
-        <li><a href="#">Profile</a></li>
-        <li><a href="#">Messages</a></li>
+        <li class="active"><a href="javascript:void(0);">Home</a></li>
+        <li><a href="javascript:void(0);">Profile</a></li>
+        <li><a href="javascript:void(0);">Messages</a></li>
       </ul>
       <br />
       <ul class="nav nav-tabs">
-        <li><a href="#">Home</a></li>
-        <li class="active"><a href="#">Profile</a></li>
-        <li><a href="#">Messages</a></li>
+        <li><a href="javascript:void(0);">Home</a></li>
+        <li class="active"><a href="javascript:void(0);">Profile</a></li>
+        <li><a href="javascript:void(0);">Messages</a></li>
       </ul>
       <br />
       <ul class="nav nav-tabs">
-        <li><a href="#">Home</a></li>
-        <li><a href="#">Profile</a></li>
-        <li class="active"><a href="#">Messages</a></li>
+        <li><a href="javascript:void(0);">Home</a></li>
+        <li><a href="javascript:void(0);">Profile</a></li>
+        <li class="active"><a href="javascript:void(0);">Messages</a></li>
       </ul>
       <h3>Justified</h3>
       <ul class="nav nav-tabs nav-justified">
-        <li class="active"><a href="#">Home</a></li>
-        <li><a href="#">Profile</a></li>
-        <li><a href="#">Messages</a></li>
+        <li class="active"><a href="javascript:void(0);">Home</a></li>
+        <li><a href="javascript:void(0);">Profile</a></li>
+        <li><a href="javascript:void(0);">Messages</a></li>
       </ul>
       <br />
       <ul class="nav nav-tabs nav-justified">
-        <li><a href="#">Home</a></li>
-        <li class="active"><a href="#">Profile</a></li>
-        <li><a href="#">Messages</a></li>
+        <li><a href="javascript:void(0);">Home</a></li>
+        <li class="active"><a href="javascript:void(0);">Profile</a></li>
+        <li><a href="javascript:void(0);">Messages</a></li>
       </ul>
       <br />
       <ul class="nav nav-tabs nav-justified">
-        <li><a href="#">Home</a></li>
-        <li><a href="#">Profile</a></li>
-        <li class="active"><a href="#">Messages</a></li>
+        <li><a href="javascript:void(0);">Home</a></li>
+        <li><a href="javascript:void(0);">Profile</a></li>
+        <li class="active"><a href="javascript:void(0);">Messages</a></li>
       </ul>
       <h3>Tabs with dropdowns</h3>
       <ul class="nav nav-tabs">
-        <li class="active"><a href="#">Home</a></li>
-        <li><a href="#">Help</a></li>
+        <li class="active"><a href="javascript:void(0);">Home</a></li>
+        <li><a href="javascript:void(0);">Help</a></li>
         <li class="dropdown">
           <a class="dropdown-toggle" data-toggle="dropdown" href="#">
             Dropdown <span class="caret"></span>
           </a>
           <ul class="dropdown-menu" role="menu">
-            <li><a href="#">Action</a></li>
-            <li><a href="#">Another action</a></li>
-            <li><a href="#">Something else here</a></li>
+            <li><a href="javascript:void(0);">Action</a></li>
+            <li><a href="javascript:void(0);">Another action</a></li>
+            <li><a href="javascript:void(0);">Something else here</a></li>
             <li class="divider"></li>
-            <li><a href="#">Separated link</a></li>
+            <li><a href="javascript:void(0);">Separated link</a></li>
           </ul>
         </li>
       </ul>
       <br />
       <ul class="nav nav-tabs">
-        <li><a href="#">Home</a></li>
-        <li class="active"><a href="#">Help</a></li>
+        <li><a href="javascript:void(0);">Home</a></li>
+        <li class="active"><a href="javascript:void(0);">Help</a></li>
         <li class="dropdown">
           <a class="dropdown-toggle" data-toggle="dropdown" href="#">
             Dropdown <span class="caret"></span>
           </a>
           <ul class="dropdown-menu pull-right" role="menu">
-            <li><a href="#">Action</a></li>
-            <li><a href="#">Another action</a></li>
-            <li><a href="#">Something else here</a></li>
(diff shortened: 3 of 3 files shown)


src/less/tabs.less
```diff
@@ -2,6 +2,18 @@
 // Tabs
 // --------------------------------------------------
 
+.tab-pane {
+  > .nav-tabs-pf {
+    font-size: @font-size-base;
```
jeff-phillips-18: This would break existing consumers. The `tab-pane` class would need to be added to the parent of all existing `nav-tabs-pf` instances.

jeff-phillips-18 (changes requested):

src/less/tabs.less
jgiardino: I agree with @jeff-phillips-18. We should only update the css files for this issue. The updates we make to core should not require changes to the html that is currently defined, because we don't want to break current implementations that use that html structure.

The main issue with the current css is this selector `.nav-tabs+.nav-tabs-pf` because it requires a very specific html structure that we can't reproduce in react.

To fix this, we should keep this selector `.nav-tabs+.nav-tabs-pf` so that we don't break current implementations, but also include an additional class-based selector that will apply the same styles `font-size: @font-size-base;`.  See another comment further down with specific suggestions on this change...

Then to fix this issue in the react Tabs component, add a class `nav-tabs-pf-secondary` to the react Tabs component so that the runtime html of [your storybook](https://rawgit.com/dabeng/patternfly-react/tabs-s/.out/index.html?knob-Secondary%20Tabs%20Justified=false&knob-Justified=false&knob-Tab%203%20Disabled=false&knob-Tab%202%20Disabled=false&knob-Tab%201%20Disabled=false&knob-Navigation%20Style=tabs&knob-Label=Danger%20Will%20Robinson%21&knob-Tab%203%20Tittle=Tab%203&knob-Tab%202%20Tittle=Tab%202&knob-Tab%201%20Tittle=Tab%201&selectedKind=Tabs&selectedStory=Secondary%20Tabs&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) resembles this:
```
...
<div id="secondary-tabs-1">
  <ul role="tablist" class="nav nav-tabs nav-tabs-pf nav-tabs-pf-secondary">
...
```

src/less/tabs.less
```diff
@@ -26,15 +38,6 @@
       }
     }
   }
-  + .nav-tabs-pf {
```
jgiardino, line 29: ... and then following up to my previous comment, you should be able to change the following selector:
`+ nav-tabs-pf`
to this:
`+ .nav-tabs-pf, .nav-tabs-pf-secondary`

But please make similar updates in the .scss file, and then please test that they actually work. 

Thanks for making this update!

dabeng:
Hi @jeff-phillips-18 , the following picture demonstrates what the PR has done.
screen shot 2018-01-12 at 1 31 16 pm

IMO, the top one confuses users because secondary tabs don't make any response when toggling primary tabs. However, I‘v seen the many application cases of the bottom one. I remember that there are lots of such nested secondary tabs in Errata when I contributed front-end code to it.

src/less/tabs.less
dabeng: Thanks a lot for your lighter solution. While, I found that this solution can not work in secondary tabs of patternfly-react because the less sentence `+ .nav-tabs-pf, .nav-tabs-pf-secondary` will produce the following css settings:
```css
.nav-tabs .nav-tabs-pf-secondary, .nav-tabs+.nav-tabs-pf {
    font-size: 12px;
}
```

dabeng: In patternfly-react secondary tabs, I need the following css settings:
```css
.nav-tabs-pf.nav-tabs-pf-secondary {
    font-size: 12px;
}
```
or
```css
.tab-pane .nav-tabs-pf-secondary {
    font-size: 12px;
}
```
or
```css
.tab-pane .nav-tabs-pf.nav-tabs-pf-secondary {
    font-size: 12px;
}
```
Sorry @jgiardino , I'm a js programmer and not good at Less, Sass, css style priority 😞. Could you provide another solution 😊

dabeng:
In patternfly-react secondary tabs, I need the following css settings:
```css
.nav-tabs-pf.nav-tabs-pf-secondary {
    font-size: 12px;
}
```
or
```css
.tab-pane .nav-tabs-pf-secondary {
    font-size: 12px;
}
```
or
```css
.tab-pane .nav-tabs-pf.nav-tabs-pf-secondary {
    font-size: 12px;
}
```
Sorry @jgiardino , I'm a JS programmer and not good at Less, Sass, css style priority 😞. Could you help me to choose a sensible solution ? Thanks in advance 😊

src/less/tabs.less
jgiardino: Thanks for pointing that out, @dabeng! You're right. 

I played around with the css locally, and this is what I had to do to get it to work (in the .scss version)...

In the following structure:
```
.nav-tabs {
  ...
  + nav-tabs-pf {
    ...
  }
}
```
Remove the selector `+ nav-tabs-pf` and it's styles so that it isn't nested inside `.nav-tabs`. Move this selector to a separate block. This will require that you change `+ nav-tabs-pf` to `.nav-tabs + nav-tabs-pf`, but you will also add the `.nav-tabs-pf-secondary` class to this selector, so that you end up with something like this: 
```
.nav-tabs {
...
}
.nav-tabs + .nav-tabs-pf, .nav-tabs-pf.nav-tabs-pf-secondary {
  font-size: $font-size-base;
  > li:first-child > a {
    padding-left: 15px;
    &:before {
      left: 15px !important;
    }
  }
}
```
This solution maintains both selectors—the original selector for supporting the original patternfly core solution and the new selector for supporting the patternfly-react solution. This will result in runtime css that looks like this:
```
.nav-tabs + .nav-tabs-pf, .nav-tabs-pf.nav-tabs-pf-secondary {
    font-size: 12px;
}
```

jgiardino:
Hi @dabeng,

I responded to your question with an inline comment. Hopefully this makes sense :-)

However, there has been discussion with @mcarrano on Slack regarding what the expected interaction should be with secondary tabs. We'd like to get input from @kybaker on this one before moving forward, so that we can confirm if our expectations for how this component works actually match the original intended design for the component.

src/less/tabs.less
dabeng: Awesome 👍 Appreciated 😊

dabeng:
Hi @jgiardino , I have updated css files according to new comments.

jeff-phillips-18:
@dabeng Could you add to the existing tabs test page or create a new secondary-tabs test page?

src/less/tabs.less
jgiardino: @dabeng, I'm so sorry. I made a mistake in my original comment. I often will pull selectors out of a nested structure to make sure I get the right selectors, and then put them back in the nested structure if possible. But this time I forgot to put the selectors back into the nested structure.

We still want the following structure:
```
.nav-tabs {
  ...
  + nav-tabs-pf {
    ...
  }
}
```
But to get the selectors we need for the secondary tabs, we would modify it like this: 
```
.nav-tabs {
  ...
  + nav-tabs-pf, &.nav-tabs-pf-secondary {
    ...
  }
}
```

jgiardino:
@dabeng I made a correction to my previous inline comments.

Also, please move forward with finalizing the updates as we have discussed. I met with @kybaker and @mcarrano to discuss the expected interaction of the examples labeled ".nav-tabs + .nav-tabs-pf" on the [Tabs test page](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/tabs.html). They agree that it is possible that designers would want to have the secondary tabs nested inside the tab panel for the primary nav. They have also requested that the updates we make to the css will continue to support the original html implementation. This is inline with the latest updates you have been making.

dabeng:
@jgiardino @jeff-phillips-18 , please review the “.nav-tabs.nav-tabs-pf-secondary” example on the [tabs test page](https://rawgit.com/dabeng/patternfly/sec-tabs-dist/dist/tests/tabs.html)

jeff-phillips-18 (dismissed):

jgiardino:
@dabeng I like that you include an example showing the alternate html structure that is now possible. However, I have a concern regarding the js and interaction for this example.

The interaction you have for the [react tabs component](https://rawgit.com/dabeng/patternfly-react/tabs-s/.out/index.html?knob-Secondary%20Tabs%20Justified=false&knob-Justified=false&knob-Tab%203%20Disabled=false&knob-Tab%202%20Disabled=false&knob-Tab%201%20Disabled=false&knob-Navigation%20Style=tabs&knob-Label=Danger%20Will%20Robinson%21&knob-Tab%203%20Tittle=Tab%203&knob-Tab%202%20Tittle=Tab%202&knob-Tab%201%20Tittle=Tab%201&selectedKind=Tabs&selectedStory=Secondary%20Tabs&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybooks%2Fstorybook-addon-knobs) works the way I expect it to.

The interaction of the new example labeled **.nav-tabs.nav-tabs-pf-secondary** is inconsistent in interaction—when I click the nth primary tab, the nth secondary tab is also selected. This example is potentially confusing to developers/designers who are only looking at this example, especially due to the lack of design documentation. 

~~My recommendation would be to remove the JS from this example, and just use it as an example of possible html structure. When I check the other tab examples, none of them have JS either.~~

@andresgalante @jeff-phillips-18 Do you have thoughts on this?

mcarrano:
I agree with this concern @jgiardino . When I click to a new primary tab, I would expect that the secondary tabs will change but the first one will always be selected by default.

jeff-phillips-18:
@dabeng I like having the interaction work correctly. Can you update such that the first of the secondary tabs is active by default?

Would also be nice to use `href="javascript:void(0);` rather than `href="#"` so the page does not jump to the top when click on the tabs.

jgiardino:
Ignore this comment from me above:
> ~~My recommendation would be to remove the JS from this example, and just use it as an example of possible html structure. When I check the other tab examples, none of them have JS either.~~

I'm fine with keeping the JS and interaction, as long as you fix the issues as @jeff-phillips-18 suggests.

dabeng:
yep, I updated it again : ) @jeff-phillips-18 @jgiardino @mcarrano

jeff-phillips-18:
@dabeng Looking better. Though I am now second guessing myself. I don't see a reason for this version of tabs to have interaction (making the primary tab active on click) whereas none of the others not the secondary tabs have the same interaction. I'm ok with leaving it though.

Could you update the rest of the tabs on this page to use `href="javascript:void(0);` rather than `href="#"` as well? It is really annoying when it scrolls to top on click.

dabeng:
OK , no problem. @jeff-phillips-18

jeff-phillips-18 (approved):

jgiardino:
The styling looks good to me. I'll defer to @mcarrano and @jeff-phillips-18 regarding the interaction.

mcarrano:
Yes, this looks right to me now, @jgiardino . Thanks, @dabeng

PR #908, opened 2018-01-11
feat(dependencies): Update older dependencies

### LINKED ISSUE #892: Why is moment.js pinned at minor level rather than major?
92cf2415923e499b4d2b7db05a08d5291774a11b added the dependency on moment but pins it at the minor level. Is there a reason not to pin at the major version? So ```^2.14.1``` rather than ```~2.14.1```.

jeff-phillips-18:
## Description
Updates out of date dependencies for:
- moment
- bootstrap-datepicker
- datatables.net-colreorder

Fixes #892 

Regression tests passed.
Test pages for table and time picker, forms with data picker.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -21,7 +21,7 @@
   ],
   "dependencies": {
     "bootstrap": "~3.3.7",
-    "bootstrap-datepicker": "~1.6.4",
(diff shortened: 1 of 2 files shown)


dlabrecq (approved):
LGTM

Issue #901, opened 2018-01-06
No default jsDelivr CDN file set


xeeljire:
[This package](https://www.jsdelivr.com/package/npm/patternfly) doesn't have a [default file](https://www.jsdelivr.com/features#publishing-packages) set. You can set it via `jsdelivr`, `browser`, or `main` field in `package.json`

cliffpyles:
@xeeljire Thanks for creating this ticket! What time are you typically online? I'd like to get a little clarification around the ticket?

MartinKolarik:
@cliffpyles hi - jsDelivr team member here. Basically what the message means is that we don't know which file from your package is the one users should use. When you set a default file, we can show it at the top of the list, like [here](https://www.jsdelivr.com/package/npm/jquery). You'll also be able to use a shorter URL https://cdn.jsdelivr.net/npm/patternfly@3.37.4 instead of https://cdn.jsdelivr.net/npm/patternfly@3.37.4/dist/js/patternfly.min.js

#913 takes care of this.

MartinKolarik:
Admittedly, this is not so useful when there are multiple files required to use the package.

cliffpyles:
@MartinKolarik Thanks for the input! I think I am in agreement about the usefulness of the change with the current organization of the JS. The library is primarily for CSS and the JS is typically handled in external projects (with a few exceptions like you mentioned).

@xeeljire Since there isn't a single main JavaScript file that should be imported, would you mind providing some more context about what you are wanting to do?

dgutride:
Closing due to lack of information - please reopen if there are any further details.

PR #913, opened 2018-01-16
fix(package.json): No default jsDelivr CDN file set

dabeng:
## Description
This PR is targeted to fix the [issue](https://github.com/patternfly/patternfly/issues/901)

## Changes
The following warning message will disappear:
screen shot 2018-01-16 at 2 19 46 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,7 @@
 {
   "name": "patternfly",
   "version": "0.0.0-semantically-released",
+  "main": "dist/js/patternfly.js",
   "author": "Red Hat",
   "license": "Apache-2.0",
   "dependencies": {


amarie401 (approved):

LHinson:
@xeeljire will you review this one as well?

cliffpyles:
Before we merge this one, I'd like to get a little more information about what problem the original issue was trying to solve. Since the project is primarily a CSS project (with a few exceptions), it seems like this change might not be appropriate for the goals of the project.

It could be a possibility in the future though. We will just need to put a little thought into the organization of the JS. We don't currently offer the JS assets as ES6 modules. @dgutride @jeff-phillips-18 @priley86 What are your thoughts on this?

priley86:
@cliffpyles i hope i am not conflating this issue... however one thing that has come up in the past is we'd love to be able to consume PF settings/globals as ES6 modules instead of globals. This came in pf-react recently here:

https://github.com/patternfly/patternfly-react/blob/master/src/common/patternfly.js

i.e. instead of saying:
```
import 'patternfly/dist/js/patternfly-settings';
import 'patternfly/dist/js/patternfly-settings-charts';

const patternfly = window.patternfly;
const c3ChartDefaults = patternfly.c3ChartDefaults();
```
it might be nice to also expose an ES6 Module in PF Core (so something like this would be supported) :
```
import {pfBreakpoints, c3ChartDefaults, ... } from 'patternfly';
```
Not sure if this is possible yet - have not looked into it. Icing on the cake here would also be adding the `mobile: 0` breakpoint:
https://github.com/patternfly/patternfly-react/pull/88#pullrequestreview-89568402

cc: @jeff-phillips-18 @mturley

cliffpyles:
@priley86 Yeah, I agree. ES6 modules would be awesome! I'm wondering about level of effort and scope of this PR though. @LHinson I'll add a ticket to the backlog to investigate. Any objections?

amarie401:
I agree with you both. I think introducing ES6 modules would be a great move. @cliffpyles should we hold off then on this merge?

LHinson:
that sounds good @cliffpyles - we can point it monday at grooming.

dabeng:
Hi @priley86 @cliffpyles . You are right. It's better to hold off this PR until we refactor the patternfly-core according to ES6 module specification. You know that current patternfly.js is a set of jquery plugin settings. It will be a huge task.

jeff-phillips-18:
@dabeng Should this be closed at this point?

LHinson:
Yes, I'll go ahead and close. Thanks.

Issue #902, opened 2018-01-08
Donut chart default color pattern has only 2 colors


amirfefer:
`blue` and `black-300` via [getDefaultDonutColors](https://github.com/patternfly/patternfly/blob/f3e97eed896f1cae0402a86ee526b68fc7abbb96/src/js/patternfly-settings-charts.js#L150).
Shouldn't we add a 6 colors (or more) pattern?

dabeng:
Hi @amirfefer , I don't think this is a issue.

According to the [design doc ](http://www.patternfly.org/pattern-library/data-visualization/donut-chart/#/design)of utilization donut chart, it only include two factors(used & available), so two colors ` patternfly.pfPaletteColors.blue` `patternfly.pfPaletteColors.black300` are just the right amount.

Of course, you can specify the colors for your utilization donut chart with following code snippets:
```js
  donutConfig.color = { pattern: [
      $.pfPaletteColors.red,
      $.pfPaletteColors.green,
    ]
  };
```

ohadlevy:
I would think it would be nice if we had a default colors (maybe the same
as in the design docs). As it stands now it's inconsistent as for one type
you have a default vs the other which doesn't.

jeff-phillips-18:
Agreed @ohadlevy. Based on https://github.com/theforeman/foreman/pull/5137#issuecomment-356959608, I think the default colors ought to be:

            patternfly.pfPaletteColors.blue,
            patternfly.pfPaletteColors.black300
            patternfly.pfPaletteColors.red100
            patternfly.pfPaletteColors.orange400
            patternfly.pfPaletteColors.green400
            patternfly.pfPaletteColors.cyan500
            patternfly.pfPaletteColors.gold200

The first 2 need to be what is already defined as to not break anyone currently using 2 default colors.

dabeng:
@jeff-phillips-18 , you mean we should remove the following settings
```js
    colors: {
      Cats: $.pfPaletteColors.blue,
      Hamsters: $.pfPaletteColors.green,
      Fish: $.pfPaletteColors.orange,
      Dogs: $.pfPaletteColors.red
    },
````
> https://github.com/patternfly/patternfly/blob/master/tests/pages/_includes/widgets/charts/donut-whole-relationship.html#L9~L14

> https://github.com/patternfly/patternfly/blob/master/tests/pages/_includes/widgets/charts/donut-mini.html#L8~L13

and replace the following settings
```js
    getDefaultDonutColors = function () {
[...]

PR #916, opened 2018-01-18
chore(donut chart): Donut chart default color pattern has only 2 colors

dabeng:
## Description
This PR is targeted to fix the issue #902

## Changes

Add more default colors for donut chart. 

Here's the [demo](https://rawgit.com/dabeng/patternfly/u-d-chart-dist/dist/tests/donut-charts.html)

### DIFF
--- a/src/js/patternfly-settings-charts.js
+++ b/src/js/patternfly-settings-charts.js
@@ -155,6 +155,18 @@
           ]
         };
       },
+      getDefaultRelationshipDonutColors = function () {
+        return {
+          pattern: [
+            patternfly.pfPaletteColors.blue,
+            patternfly.pfPaletteColors.red100,
+            patternfly.pfPaletteColors.orange400,
+            patternfly.pfPaletteColors.green400,
+            patternfly.pfPaletteColors.cyan500,
+            patternfly.pfPaletteColors.gold200,
+          ]
+        };
+      },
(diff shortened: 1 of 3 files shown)


src/js/patternfly-settings-charts.js
```diff
@@ -151,7 +151,11 @@
         return {
           pattern: [
             patternfly.pfPaletteColors.blue,
-            patternfly.pfPaletteColors.black300
```
jeff-phillips-18: As I mentioned in https://github.com/patternfly/patternfly/issues/902#issuecomment-358050813, we cannot change the first two colors returned by getDefaultDonutColors  or from a call to getDefaultDonutConfig because it would break existing consumers. 

We should, as the conversations in https://github.com/patternfly/patternfly/issues/902 concluded, have two donut color configs, one default (current used for utilization charts) and one for relationship donut charts.

I would suggest adding methods  getDefaultRelationshipDonutColors and getDefaultRelationshipDonutConfig to allow consumers to retrieve defaults for the relationship donut charts.

jeff-phillips-18 (changes requested):

dabeng:
OK, I get it.

jeff-phillips-18 (approved):

amarie401 (approved):

Issue #920, opened 2018-01-18
Issues with  navbar-brand-icon and navbar-brand-txt styling in vertical navigation


jeff-phillips-18:
The navbar-brand-icon does not set a height therefore if an image with > 35px in height is used the masthead height grows and does not work properly with the left vertical nav menus.

The navbar-brand-txt does not set a line-height and the text does not align vertically. Setting a line-height of 34px aligns it correctly.

I'm not sure if this effects horizontal nav as well.

matthewcarleton:
hey @jeff-phillips-18 so are you thinking a max-height:35px; on the image and a line-height of 34px on the text?

mturley:
This is now being resolved as part of #922

PR #922, opened 2018-01-21
fix(navbar-vertical): Set line-height for .navbar-brand-txt and max-height for .navbar-brand-icon

mturley:
## Description
As per a discussion in https://github.com/patternfly/patternfly-react/pull/88, I needed this bit of CSS for the example titles in VerticalNavigationMasthead to align correctly, and I was told that it should be moved to patternfly core.

## Changes

Title says it all. Applies these style rules using some new variables, to make sure images used in the masthead are sized correctly and text positioned correctly.

Here's what the text looks like before:

screen shot 2018-01-21 at 5 58 20 pm

And after:

screen shot 2018-01-21 at 5 59 17 pm

### DIFF
--- a/src/sass/converted/patternfly/_navbar-vertical.scss
+++ b/src/sass/converted/patternfly/_navbar-vertical.scss
@@ -114,6 +114,11 @@
     .navbar-brand-icon {
       display: inline-block;
       margin: $navbar-pf-vertical-navbar-brand-icon-margin;
+      max-height: $navbar-pf-vertical-navbar-brand-max-height;
+    }
+
+    .navbar-brand-txt {
+      line-height: $navbar-pf-vertical-navbar-brand-line-height;
     }
   }
 
--- a/src/less/navbar-vertical.less
+++ b/src/less/navbar-vertical.less
@@ -114,6 +114,11 @@
     .navbar-brand-icon {
       display: inline-block;
       margin: @navbar-pf-vertical-navbar-brand-icon-margin;
+      max-height: @navbar-pf-vertical-navbar-brand-max-height;
+    }
+
+    .navbar-brand-txt {
+      line-height: @navbar-pf-vertical-navbar-brand-line-height;
     }
   }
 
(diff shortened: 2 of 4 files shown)


mturley:
@jeff-phillips-18 @jgiardino @priley86 here's that bit of CSS from my PR. Removing it from the React PR, which if we merge first will just have some misaligned text in storybook, no big deal.

mturley:
@patternfly/patternfly-core-maintainers not sure who I should have review this individually. Would be nice to get it merged before the next patternfly-react community meeting so the masthead looks good in that demo.

jeff-phillips-18:
@mturley Can you make the adjustment necessary for the navbar brand icon as well? See issue #920 

If you do that, this PR will resolve #920

mturley:
Sure, @jeff-phillips-18, is it sufficient to add `max-height: @navbar-pf-vertical-navbar-brand-min-height;` on the image? using that same variable, I mean?

matthewcarleton:
@mturley you'd have to create a new var that uses max instead of min. so `@navbar-pf-vertical-navbar-brand-max-height`

src/sass/converted/patternfly/_navbar-vertical.scss
```diff
@@ -115,6 +115,10 @@
       display: inline-block;
       margin: $navbar-pf-vertical-navbar-brand-icon-margin;
     }
+
+    .navbar-brand-txt {
+      line-height: $navbar-pf-vertical-navbar-brand-min-height;
+    }
   }
```
matthewcarleton: Hey Mike, I would suggest creating a new variable for the line height. The line height shouldn't rely on the min height variable.

mturley:
@matthewcarleton I was thinking maybe, why not just change it to `navbar-pf-vertical-navbar-brand-height` instead of having two `min-height` and `max-height` variables that are the same value? Happy to do either.

mturley:
Also, I had used the min-height variable for the line-height because it was close enough, but it's 35px not 34px. Feels a little silly over a difference of 1px, but I'm going to correct it with a new variable `navbar-pf-vertical-navbar-brand-line-height` at 34px, if that's ok.

mturley:
@jeff-phillips-18, @matthewcarleton, I pushed the max-height fix and the new line-height variable, without consolidating the max-height and min-height variables. I could go either way on whether we should. If you guys don't think we should bother, we can go ahead and merge this.

matthewcarleton (approved):
This looks good Mike, thanks for taking care of this. I'm happy to merge as is.

jeff-phillips-18 (approved):

PR #933, opened 2018-01-24
fix(kebab): Add sass file updates from #835

jeff-phillips-18:
#835  was merged without including the Sass updates.

### DIFF
--- a/src/sass/converted/patternfly/_dropdowns.scss
+++ b/src/sass/converted/patternfly/_dropdowns.scss
@@ -170,6 +170,7 @@
 
 // Kebab dropmenu
 .dropdown-kebab-pf {
+  &.btn-group > .btn:first-child,
   .btn-link {
     color: $gray-darker;
     font-size: ($font-size-base + 4);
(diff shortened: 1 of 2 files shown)


matthewcarleton (approved):
nice catch 👍

jeff-phillips-18:
Better late than never

jgiardino (commented):
These updates look good. Thanks @jeff-phillips-18!

Also, according to 
the [less to sass conversion](https://github.com/patternfly/patternfly#less-to-sass-conversion) section in the readme, my expectation was that the sass conversion happens automatically as part of the PR merge process. Is that correct? And is this just an odd case where the sass needed to be added manually? And if that's true, how did you know? (@matthewcarleton and I were curious how to know in case this happens again)

jeff-phillips-18:
@jgiardino No you have to run the build sass commands `npm start -- --sass` or `grunt build --sass`

I should have seen that there were no sass files added and flagged it on the PR.

jgiardino:
Ah, ok. So that command isn't just for me to test the sass conversion locally, but I would actually include those files in my commit? I think it was this statement that made me think that whole process was automated:
> The Less to Sass Conversion step will be accomplished and managed as a part of any Pull Request which includes Less file changes.

I'll try to think of a suggestion for updating it so that it's clearer.

PR #938, opened 2018-01-25
fix(alerts): updates aria info

### LINKED ISSUE #891: Alerts dismiss button are not screenreader accessible
In the [alerts test page](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/alerts.html), for the alerts that have the X dismiss button, the current html results in the button not being screen reader accessible.

**Current HTML:**
```
<button type="button" class="close" data-dismiss="alert" aria-hidden="true">
  <span class="pficon pficon-close"></span>
</button>
```
Issues:

- Using `aria-hidden="true"` tells the screen reader to ignore this element.
- There is no text string associated with this button that can be read aloud by the screen reader

**Suggested HTML:**
```
<button type="button" class="close" data-dismiss="alert" aria-label="Close">
  <span class="pficon pficon-close"></span>
</button>
```
Suggestions include:

- Remove `aria-hidden="true"`
- Add `aria-label="Close"`

matthewcarleton:
This updates the aria info on the alerts as per this issue  #891 

Closes #891

### DIFF
--- a/tests/pages/_includes/widgets/communication/alert-danger-dismissable.html
+++ b/tests/pages/_includes/widgets/communication/alert-danger-dismissable.html
@@ -1,5 +1,5 @@
 <div class="alert alert-danger alert-dismissable">
-  <button type="button" class="close" data-dismiss="alert" aria-hidden="true">
+  <button type="button" class="close" data-dismiss="alert" aria-label="Close">
     <span class="pficon pficon-close"></span>
   </button>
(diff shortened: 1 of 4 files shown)


jeff-phillips-18 (approved):

jgiardino (approved):
Looks good! I'm only able to test on mac voiceover, but this is pretty basic.

PR #940, opened 2018-01-26
fix(wizard): updates focus state for sub items

### LINKED ISSUE #869: Wizard: show keyboard focus in the steps
When I use the keyboard to navigate the steps in the [wizard](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/wizard.html), it is difficult to know which step has focus.

Note: In the [core test page](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/wizard.html), I'm not able to place focus on a primary step, but in the [pf-react implementation](https://github.com/patternfly/patternfly-react/pull/69), I can and in this example focus is visible.

matthewcarleton:
## Description
Added styling for focus and hover on sub nav text for better accessibility. 

I discussed this with @jgiardino and we concluded it made more sense to include just the underlined text for focus rather than reinstating the outline. In order to reinstate the currently disabled outline the css would have to be refactored and this seems beyond scope at this point. 

closes #869 

https://rawgit.com/matthewcarleton/patternfly/wizard-sub-nav-focus-dist/dist/tests/wizard.html

### DIFF
--- a/src/less/wizard.less
+++ b/src/less/wizard.less
@@ -52,6 +52,14 @@
           width: 14em;
           &:hover {
             text-decoration: none;
+            background-color: @color-pf-black-200;
+          }
+          &:focus {
+            //corrects odd behavior when hover and focus are combined.
+            text-decoration: none;
+            span {
+              text-decoration: underline;
+            }
           }
         }
         &.active {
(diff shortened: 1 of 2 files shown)


jgiardino:
Thanks for adding this! I agree that we should go with a minimal solution for this.

jeff-phillips-18 (approved):

srambach:
Seems like a good solution to keep it simple for now.

srambach (approved):

mcarrano (approved):
I'm good!

Issue #641, opened 2017-05-10
Examples should use ids to limit JS scope


michpetrov:
Since some examples may end up being included together their event listeners could interfere. This is currently causing patternfly/patternfly-org#408. Having an id will make sure the event listeners are not applied outside of the example. I'll prepare a PR.

jeff-phillips-18:
Fixed by #952

PR #952, opened 2018-02-08
fix(list view): add ids to limit JS scope

michpetrov:
add ids to examples to avoid interference between JS scripts in case multiple examples are included
on the same page

re #641


LHinson:
Thanks for your contribution @michpetrov! @cliffpyles will you review this one?

jeff-phillips-18 (approved):

PR #953, opened 2018-02-12
fix(sidebar): support for sidebar within vertical navigation

### LINKED ISSUE #512: Sidebar doesn't work properly with tertiary-nav
The `$(document).sidebar()` function isn't implemented for `navbar-pf-vertical` only `navbar-pf`:

screenshot from 2016-10-17 10-55-26

jeff-phillips-18:
## Description
Updated sidebar functions to correctly size when used within vertical navigation. 

Fixes #512 

http://rawgit.com/jeff-phillips-18/patternfly/sidebar-dist/dist/tests/sidebar-vertical-navigation.html

@dabeng

### DIFF
--- a/src/js/patternfly-functions-sidebar.js
+++ b/src/js/patternfly-functions-sidebar.js
@@ -11,7 +11,12 @@
       documentHeight = $(document).height();
       navbarpfHeight = $('.navbar-pf').outerHeight();
       colHeight = documentHeight - navbarpfHeight;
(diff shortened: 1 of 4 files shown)


jeff-phillips-18:
Note that this only fixes the sidebar when setup the way the original sidebar (for horizontal navigation). Would be better to style it such that the sidebar was set to be the full height of its parent and its parent's height was set correctly based on the application's usage.

jeff-phillips-18:
Killing this fix. Should be fixed via styiing.

PR #955, opened 2018-02-13
chore(tests): Upgrade backstop and add visual regression tests for navigation

cliffpyles:
## Description
The vertical and horizontal navigations were missing regressions tests. This PR adds the regression test scenarios, and makes the necessary code changes for the tests to work.

https://patternfly.atlassian.net/browse/PTNFLY-2676

Replaces https://github.com/patternfly/patternfly/pull/889/

## Changes

- add regression test scenario for horizontal navigation
- add regression test scenario for vertical navigation
- upgrades backstop
- adds a script to the test pages for applying appropriate class names (for reproducing the state)
- adds a query param to the end of all regression test urls
- adds reference images test against


jeff-phillips-18:
Why are most of the references removed?

jeff-phillips-18:
Is there a way to name the reference image files without the `backstop_default_` prefix and possibly lose the `_<n>_document_<n>_` as well? 

Better if we just had 

`table-view-columns_large-device.png`

instead of 

`backstop_default_table-view-columns_0_document_0_large-device.png`

cliffpyles:
The references are the ones we discussed the other day. They were removed a while back, but they made it back in. It could have been a force push or something. I'm not sure.

The naming is the default convention that backstop uses. I'm not sure the level of effort to change it's naming conventions, but I can look into it. To keep the scope limited to the work in the Jira ticket, are you good with me doing that in a separate PR?

jeff-phillips-18:
Can you confirm there there are references for each test page then? Seems all of them for some pages are removed. I thought the ..._..._..._..._..._...._...._large-device.png's were supposed to be kept.

jeff-phillips-18:
those names make it difficult to review PRs like this.

cliffpyles:
I looked at the Backstop docs, but I'm not seeing any hooks to modify the display format. So we would need to contribute the functionality upstream if that is something we want to do.

I looked through the files and I couldn't find any that are missing and shouldn't be. Was there a specific scenario in question?

jeff-phillips-18:
I see, they say "Diff not rendered" others say "Deleted file not rendered." I was confused.

cliffpyles:
@jeff-phillips-18 Is there anything else needed on this PR? Also – is there another person you want me to assign as the second reviewer, since Dana is on PTO?

jeff-phillips-18 (approved):

PR #969, opened 2018-02-28
Move test page code to include files

cliffpyles:
## Description
In order to add code snippets on patternfly.org, we need to make sure the test pages on patternfly-core aren't writing code directly on the test page. They should be including the code from a file in the widgets directory. This allows us to reuse the test pages across environments, without running into conflicts with the frontmatter.

These changes don't make any changes to the code that was originally written, other than moving it to a new file.

https://patternfly.atlassian.net/browse/PTNFLY-2734

## Changes

* Moves modal-overlay test page code to a new file
* Moves pagination test page code to a new file
* Moves filter test page code to a new file
* Moves bootstrap-slider test page code to a new file


matthewcarleton (approved):
LGTM 👍 thanks Cliff!

PR #973, opened 2018-03-01
Close btn listview

### LINKED ISSUE #919: ListView close button in expansion area interferes with expansion content
The close button is overlaid on the expansion content which causes issues when there is content in that space.

See https://github.com/patternfly/patternfly-react/issues/153 for original details.

andresgalante:
Fixes #919 by changing the close btn to have an absolute position instead of a float


matthewcarleton:
rawgit link :) ?

matthewcarleton (approved):

PR #976, opened 2018-03-05
fix(pagination): disabled buttons display not-allowed cursor

### LINKED ISSUE #975: Disabled pager does not show `not-allowed` cursor
screen shot 2018-03-05 at 2 24 02 pm

☝️ look, see? Its not a disabled cursor 😭 on drats, doesn't show the cursor, well... just know... it doesn't show `not-allowed`

Offending line is here:
https://github.com/patternfly/patternfly/blob/master/src/less/pager.less#L35

AllenBW:
fixes #975

well fixes most of #975, still gotta do that other part

#but what's it look like? 
pager

### DIFF
--- a/src/less/pager.less
+++ b/src/less/pager.less
@@ -32,7 +32,7 @@
       background: @pagination-bg;
       .box-shadow(none);
       color: @pager-disabled-color;
-      cursor: default;
+      cursor: not-allowed;
     }
   }
   .next {
--- a/src/less/pagination.less
+++ b/src/less/pagination.less
@@ -43,7 +43,7 @@
     > a:hover,
     > a:focus {
       .box-shadow(none);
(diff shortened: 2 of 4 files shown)


jeff-phillips-18 (approved):

matthewcarleton (approved):
thanks for taking care of this!

matthewcarleton:
@jeff-phillips-18 do you want to open an issue for the `Would be nice to have styling to allow the use of a button rather than an anchor as well.` ?

jeff-phillips-18:
Yes please @matthewcarleton

AllenBW:
If someone assigns me to that issue once its open... 🌋 **I'LL MAKE IT HAPPEN** 😜

PR #979, opened 2018-03-07
feat(inline edit): add inline-edit in a form

### LINKED ISSUE #967: Inline edit example missing from core
Related to pf-react issue: https://github.com/patternfly/patternfly-react/issues/187 which states: 

From @dabeng 
> I created a jquery version of [inline-edit in a form](https://rawgit.com/dabeng/patternfly/inline-edit2-dist/dist/tests/inline-edit.html) according to the [design doc](http://www.patternfly.org/pattern-library/forms-and-controls/inline-edit/#design) on pattern.org.

dabeng:
## Description
Fixes #967 

All the input fields are existing widgets in patternfly-core repo.

Maybe we can regard it as a starting point(**[inline edit in a form](https://rawgit.com/dabeng/patternfly/inline-edit2-dist/dist/tests/inline-edit-for-form.html)**) and go on with other inline-edit forms(inline-edit in a list view, table view) described in design doc.

@mcarrano @jgiardino @cliffpyles , looking forward to your advices.

Here's a [demo page](https://rawgit.com/dabeng/patternfly/inline-edit2-dist/dist/tests/inline-edit-for-form.html).

## Additional Information
1. All the components used in "inline-edit in a form" are existing components listed in patrernfly library section of patternfly.org;

2. Currently, "inline-edit in a form" pattern includes 6 types of input field - enhanced text box, enhanced textarea, simple  single select of bootstrap select, multiple select of bootstrap select, bootstrap combobox, bootstrap datepicker;

3. screen shot 2018-02-26 at 10 37 46 am spinner is used to mock async operations in common scenarios, e.g. submitting updated results via ajax after users complete their inline-edit;

4. IMO, "inline-edit in a form" shouldn't be entirely responsible for all sorts of form validation. In fact, form validation is complex enough to be a separate js library([Parsley](http://parsleyjs.org/) is a good example). Moreover, we have ["Errors and Validation"](http://www.patternfly.org/pattern-library/forms-and-controls/errors-and-validation/) design already. Although not comprehensive and detailed, "Errors and Validation" is indispensable part of front-end development;

5. In "inline-edit in a form", there is only one field validation - screen shot 2018-02-26 at 12 00 06 pm save button will be enabled when user inputs the new value for a field;

6. Due to the following [code snippets](https://github.com/danielfarrell/bootstrap-combobox/blob/master/js/bootstrap-combobox.js#L424) of bootstrap combobox:
```js
, click: function (e) {
      e.stopPropagation();
      e.preventDefault();
      this.select();
      this.$element.focus();
    }
```
I can't capture the "click" event outside the combobox and then the save button can not be enabled when user selects the different dropdown menu item by click. Using arrow keys + Tab/Enter is OK;

7.  "inline-edit in a list view" is easy to implement, but "inline-edit in a table view" seems like a difficult task. For now, Datatables doesn't support inline row edit for now. There are some valuable references:
- [Inline row edit not working](https://datatables.net/forums/discussion/34765)
- [Inline table editing of a child row](https://datatables.net/forums/discussion/comment/125596/#Comment_125596)

8. We know that where is input, where is XSS. It will be a big bonus point if adding some anti-xss logic to input components or forms. The choice of anti-xss js libraries is not small. If you are interested in this field, I recommend you [jsxss](http://jsxss.com/en/index.html) 😊.


mcarrano:
This looks good @dabeng except there is a bug with the way the combo box exit works. Seems I need to add an extra key press to enable the Accept button. Also, why is there an extra "x" button after the combo box?

dabeng:
Hi @mcarrano. Thanks a lot for your review 😊 

1. there are three ways to enable "✔" button : 

- input key words(optionally use ⬇ ⬆arrows) + press Tab key

- input key words(optionally use ⬇ ⬆arrows) + press Enter key

- directly select dropdown menu item by clicking([this feature will work well after this PR is merged](https://github.com/patternfly/patternfly-bootstrap-combobox/pull/3))

I wonder which operation confuses you.

2. This PR adheres to the [design doc](http://www.patternfly.org/pattern-library/forms-and-controls/inline-edit/#design) of inline edit on patternfly.org.
form_edit_patterns

I think the design intention of adding extra "x" button is to keep the uniform inline-edit style. Additionally, "x" button within bootstrap-combobox is used to clear up input box, however "x" button adjacent to "✔" button is used to close the whole edit status of bootstrap-combobox. It is used like "cancel" button.

matthewcarleton:
@dabeng this looks good, thanks for working on it 👍. Have you tested this in Safari? I think there are some bugs to work out there. When you unselect an item the background stays a solid blue which is a bit confusing. Also I can't seem to get the check to work at all. here's a link to a recording of it. http://g.recordit.co/5ayNfbmqfE.gif

mcarrano:
@dabeng I am still seeing a different behavior of the combo box than what you describe. The only way I can make the accept button enable is to press the Enter key. I am testing this on Firefox.

jgiardino:
Hi @dabeng! These look great! I have a few observations below:

1. The Edit buttons are not keyboard accessible. I should be able to hit the Tab key to place focus on the Edit button, but I can't. The way I recommend to make it keyboard accessible is by using a button like `<button class"btn btn-link" type="button">[edit icon]</button>`. 
    - However that would render the icon as blue, and I recently heard that they are not supposed to be blue. @matthewcarleton Do you know what the expectation is for icons that are buttons? I don't see where in the css this styling is provided. I would assume that wrapping an icon with a button with the class `btn-link` would provide the expected color.
2. I think I'm seeing the same issues that @mcarrano mentioned for the combo box. I'm testing in Chrome. I listed the issues I noticed and the steps to reproduce below.

**Ability to clear a combo box value** 
I'm assuming that I should be able to clear the value in a combo box and save the update, but I can't.
1. Click the "UXD" value for "Department" to show the combo box
2. Click the X button inside the combo box to clear the value
**Expected result:** The OK button is enabled.
**Actual result** The OK button is still disabled.

Additionally, complete the steps above then:

3. Click away from the combo box. The value "UXD" is still visible. I think this is as expected.
4. Click "UXD" to show the combo box
  **Expected result:** "UXD" is the current selected value in the combo box, and the X displays inside the combo box? OR the combo box is empty and the dropdown arrow displays inside the combo box?
  **Actual result** "UXD" displays as the current selected value in the combo box, BUT the X does not display inside the combo box. Instead, the dropdown arrow displays. When I click this button to show the menu, "UXD" is no longer visible. 

A question about this issue for @mcarrano is what the expectation is when a user changes a value, but doesn't confirm or cancel the value and exits the field. Do you expect to see the updated value the next time the user tries to edit the field? I kind of think that's the right thing in case the user accidentally clicks off the field while trying to make a selection or edit. 

**Ability to confirm an updated value in a combo box**

1. Click the "UXD" value for "Department" to show the combo box
2. Click the X button inside the combo box to clear the value
3. Click the dropdown arrow button.
4. Select a new value
**Expected result:** The OK button is enabled.
**Actual result** The OK button is still disabled. The button only becomes enabled if I hit the Enter key.

mcarrano:
In answer to your question, @jgiardino , I would expect clicking away from the field to be like an implicit cancel (i.e. just return to the last saved value). True that a use could click away unexpectedly and need to repeat their edit, but I don't think were talking about loosing very much. To work as you are suggesting would mean to somehow persist a temporary value that has not yet been committed, is that right? On the surface that seems like a good thing. But how long to you persist this? If I come back hours later, would that value still show up? I think that might be unexpected.

jgiardino:
I was thinking the data would only persist as long as the page was loaded.

mcarrano:
@dabeng @jgiardino I guess that if we can do that (persist the last value without committing to the backend while the page is loaded) that would be a good thing. But from my perspective is not an absolute requirement.

dabeng:
1. @mcarrano @jgiardino I agree that clicking away from the field should be like clicking cancel button, because "inline-edit in a form" applies to the scenarios that users just want to modify a few fields (e.g. edit one field and then save the updated field.That's why we provide standalone "✔" button for every field). However, if users want to edit many fields and then submit the whole form at once, we should provide them the following form view:

37219602-eae37adc-23fe-11e8-91d0-750e19adb1c9

 2. @matthewcarleton now the bootstrap multiple-select can work well on Safari
mcnljgs7gf

3. @jgiardino now the "✔" button adjacent to bootstrap combobox can be enabled after users clear up current input value.

4. Users will enable "✔" button by clicking different dropdown menu item of bootstrap combobox ([this feature will work well after this PR is merged](https://github.com/patternfly/patternfly-bootstrap-combobox/pull/3))

5. @jgiardino  ` The Edit buttons are not keyboard accessible.` I think this is not specific for inline-edit page. [empty state](http://www.patternfly.org/pattern-library/communication/empty-state/#code) page also includes similar action buttons. Additionally, chrome and opera provide extra highlight effect(blue halo) when action buttons are focused, but firefox and safari don't do that.

src/less/inline-edit.less
```diff
@@ -0,0 +1,65 @@
...
+      border-color: @color-pf-black-400;
+    }
+    &:after {
+      content: "\f040";
+      font-family: FontAwesome;
+      margin-left: 10px;
+    }
+    + :not(textarea) {
+      vertical-align: bottom;
+      width: ~"calc(100% - 60px)";
```
jeff-phillips-18: Any thoughts on using flexbox rather than hard-coded widths?

dabeng: Hi Jeff. I have refactored css code based on flexbox and removed the hard-code snippets - `width: ~"calc(100% - 60px)";`

jeff-phillips-18:
Looks good @dabeng .  Can you add the regression test file for this new component?

dabeng:
Hi @jeff-phillips-18 . I have added css regressions testing code for inline edit in a form.

backstop_data/engine_scripts/chromy/onReady.js
```diff
@@ -1,3 +1,4 @@
-module.exports = function (engine, scenario, vp) {
-
+module.exports = function (chromy, scenario, vp) {
+  console.log('SCENARIO > ' + scenario.label);
```
jeff-phillips-18: Left over or do we want this console statement?

dabeng: got it. I'll remove this line.

jeff-phillips-18 (dismissed):

matthewcarleton:
@jgiardino Are you ok with this PR? I haven't been tracking it as closely as you. Looks like it's ready to merge but want to confirm with you first :)

jgiardino:
Hi @dabeng, regarding my previous comment above:

> The Edit buttons are not keyboard accessible. I should be able to hit the Tab key to place focus on the Edit button, but I can't. The way I recommend to make it keyboard accessible is by using a button like `<button class"btn btn-link" type="button">[edit icon]</button>`.

and your reply to that comment:

> I think this is not specific for inline-edit page. empty state page also includes similar action buttons. Additionally, chrome and opera provide extra highlight effect(blue halo) when action buttons are focused, but firefox and safari don't do that.

I'm not referring to how the buttons look when they have focus. I'm referring instead to whether the buttons can receive focus and be triggered by the Enter key. Currently, they cannot, and the UI is not operable by a user using only a keyboard. Ideally, any UI that we implement can be operated using only a keyboard, for users who are unable to use a mouse. 

By default, `<span>` elements cannot receive keyboard focus. Elements that can receive focus by default are `<a>` and `<button>` elements. I'd suggest using a `<button>` in this case since `<a>` elements are more appropriate for navigation.

After you replace the `<span>` with a `<button>`, then try following these steps to test if the UI is keyboard accessible:

1. Open the demo page
2. Hit the Tab key to shift focus within the page until focus is on the Edit icon button
3. Hit the Enter key
4. You should now see the input field display for the field. I would also focus to automatically be placed within the input field (like it is now when you click the Edit icon).

jgiardino:
Related to my previous comment, I created the following issue to figure out how to define the color of the icon when using a `<button>`: https://github.com/patternfly/patternfly/issues/992

dabeng:
Hi @jgiardino . I have updated [demo](https://rawgit.com/dabeng/patternfly/inline-edit2-dist/dist/tests/inline-edit-for-form.html) page. Now, you can use arrows, tab and enter keys to finish all edit operations. In Safari, you need option + tab to switch focused item.


peek 2018-03-22 23-01


1. I added attribute setting `tabindex="0"` to ervery `<span class="value">`

2. You mentioned we could replace `<span class="value">` with `<button class="btn btn-link">`. IMO. this solution can not deal with textarea field

3. According to the design idea of inline-edit in a form, users can trigger edit status of one field when users click current field value. The clickable region are composed of value + pencil icon.

4. On the other hand, if we change a pseudo element pencil to a button including pencil icon, focus effect will be only applied on pencil icon. It's less intuitive than focus on current field value

jgiardino:
Thanks for making the updates! I do see that the keyboard accessibility for sighted users is much better. But I suspect that it's not very usable for the screen reader user, because there's no indication to the screen reader user that the values are editable. For example, there's no text that tells the user that they can edit the value, and there's no element being used that tells the user they can act on the element (such as `<a>` or `<button>`).

But I also don't know what the requirements are for this first implementation of the component, or what our goals are for accessibility in this implementation. If we were trying to get something that were usable for screen readers, I would suggest doing something like this:

```
<span class="value" tabindex="0">
  Mike 
  <button class"btn btn-link" type="button" aria-label="Edit Name">[edit icon]</button>
</span>
```
If the `<button>` is inside the `<span>`, then I think this would solve the issues you were seeing with the `<textarea>` element. 

I had also failed to include the `aria-label` on the button in my previous comment, which is important to include if the button label is an icon. 

The result of using a button that has a aria-label would provide the ability for the screen reader user to get a list of form elements he can act on, and this list would be similar to the following:

- Button Edit Name
- Button Edit Gender
- Button Edit Department

But in the current implementation, the user would not be able to get a list of elements, because even though I can act on the `<span>` element by click or Enter, this element isn't associated with a role that has any meaning to the user, like a `<button>` element would. And in the event that the user did have focus on the editable text, there would be nothing telling the user that they can edit the value.

Another thing that I failed to notice in a previous review is that `for` is missing from `<label>` and `id` is missing from `<input>` as shown in the following. 
  `<label class="col-sm-2 control-label">Name</label>`
  `<input type="text" value="Mike" class="form-control editor" autocomplete="off">`
The inline edit in a form is kind of a weird case for a form, because the input isn't always visible, so I'm not really sure what would happen when the `<label>` is associated with an `<input>` that isn't visible. But I do feel like it's worth having for the case where the `<input>` is visible so that the screen reader user can know what the input label is.

dabeng:
@jgiardino Great idea. Thanks a lot. 

I have the following concerns:
1. Could we add "role" and "aria-label" attributes to the tag span:
`<span class="value" tabindex="0" role="button" aria-label="Edit Name">Mike</span>`

2. As inline-edit input is hidden by default, so if we add "for" attribute to tag label and then click the label, nothing will happen

3. IMO, in common form, editor elements could be enabled by clicking corresponding labels. However, editor elements are enabled only by clicking the current value region. If users click the outside area of (even the corresponding labels) editor element of inline-edit field, the field value will revert to readonly status

jgiardino:
The demo page isn't loading at the moment, so I can't check the latest updates. But here are a few comments:

> 1. Could we add "role" and "aria-label" attributes to the tag span:
`<span class="value" tabindex="0" role="button" aria-label="Edit Name">Mike</span>`

Yes, it is possible to add `role="button"` to a generic element so that it's recognized as a button. In general, I would advise against doing this for the following reasons:

- Some assistive devices might not fully support this (without testing, we don't know for sure). 
- This method requires additional JS to fully implement all of the button-like behaviors (e.g. capturing Space and Enter key events) 
- The example we implement in core is what the JS repos will follow. My preference would be to make the patterns as simple as possible for them to duplicate, and one way we can do that is by relying on semantic html elements that already exist. 

With the current implementation, though, the way text strings are defined would most likely not work as expected. Without testing this, I'm not sure what would be read by a screen reader for the following element: 
`<span class="value" tabindex="0" role="button" aria-label="Edit Name">Mike</span>`

It could be: 
- just "Edit Name"
- just "Mike"
- or maybe in some screen readers "Edit Name Mike"

Here's some [helpful information about `aria-label`](https://developer.paciellogroup.com/blog/2017/07/short-note-on-aria-label-aria-labelledby-and-aria-describedby/)

An alternative solution that might work better is to wrap the text with "sr-only":
```
<span class="value" tabindex="0" role="button">
  <span class="sr-only">Edit Name: </span>
  Mike
</span>
```


> As inline-edit input is hidden by default, so if we add "for" attribute to tag label and then click the label, nothing will happen

I know :-( The only work-arounds I can think of are:
1. Dynamically add the `for` attribute to `<label>` whenever the `<input>` is visible
2.  Use `aria-label` for `<input>` and set the value to the text inside `<label>`, e.g. `aria-label="name"`
3. Add `id` to `<label>` and then `aria-labelledby="<label id>"` to the `<input>`

Options 2 and 3 are probably the simplest options since they don't require additional JS.

> IMO, in common form, editor elements could be enabled by clicking corresponding labels. However, editor elements are enabled only by clicking the current value region. If users click the outside area of (even the corresponding labels) editor element of inline-edit field, the field value will revert to readonly status

One of the benefits of associating `<label>` elements with `<input>` elements is so that screen reader users know what the name of the input is. Without providing some label for the input, screen reader users have no way of knowing what value they're being asked for.

dabeng:
Great solutions @jgiardino I have added "sr-only.span" and "aria-label" 😊

jgiardino:
Hi @dabeng,

Thanks for adding the sr-only text and the aria-label attributes to the inputs! I checked in Voiceover on my mac, and this text displays with the value of the field when I browse the list of form elements on the page. 

I did notice that the Save and Cancel buttons do not have text labels. You can fix this by adding the `aria-label` attribute to these buttons as well, like this: 
```
<button type="button" class="btn btn-primary save" aria-label="Save">
  <i class="glyphicon glyphicon-ok"></i>
</button>
<button type="button" class="btn btn-default cancel" aria-label="Cancel">
  <i class="glyphicon glyphicon-remove"></i>
</button>
```

I checked the keyboard accessibility, and everything related to inline edit is keyboard accessible. The menu toggle in the combobox component is not keyboard accessible, but that's more of an issue with the combobox component and not with the updates in PR.

I did notice that with the Save and Cancel buttons are in reverse order in the DOM due to the `float: right` property. This causes inconsistency with the other input types. You could wrap them with a parent `<div>` to address this.

We have been discussing adopting the [5 rules of aria](https://www.w3.org/TR/using-aria/#notes2) for patternfly 3. If that's the case, then according to the [1st rule](https://www.w3.org/TR/using-aria/#rule1), we should be using `<button>` elements for buttons, instead or recreating a button using aria attributes, like `<span role="button">`. I think whatever css issues you might see when you replace the `<span>` elements with `<button>` elements as a result can be addressed. The only issues I noticed when I checked was the text alignment with the text area and the extra padding.

jgiardino:
One more comment... I just noticed that the following button doesn't have an accessible name:
```
<button type="button" class="empty-btn">
  <span class="fa fa-times-circle fa-lg"></span>
</button>
```
You can add an aria-label attribute to this button as well, for example:
`<button type="button" class="empty-btn" aria-label="Clear Value">...</button>`

dabeng:
Hi @jgiardino I have refactored PR according to your advice:

1. add setting `aria-label=Save` and `aria-label=Cancel` to every group of save and cancel buttons
2. fix "tab index" issue for action buttons of textarea field
3.  add `aria-label="Clear Value"` to text box field
4.  replace the following code snippets 
```html
<span class="value" tabindex="0" role="button"><span class="sr-only">Edit Name: </span>Mike</span>
```
with
```html
<span class="value" tabindex="0">
  <span>Mike</span> 
  <button class="btn btn-link" type="button" aria-label="Edit Name"><i class="pficon pficon-edit"></i></button>
</span>
```

jgiardino:
Hi Xuebin,
These look good! Thanks for making the updates!

Regarding #4, when I tested the previous version with a screen reader, one thing I noticed about the following html was that when I opened the list of form elements, the screen reader announced the buttons as "Edit Name: Mike button". With the updated html, the buttons are listed without the value, for example "Edit Name". This isn't the only way to navigate using the screen reader, but I thought that in this case, it was really helpful that the value was announced along with the button label and field label.
```
<span class="value" tabindex="0" role="button">
  <span class="sr-only">Edit Name: </span>
  Mike
  <i class="pficon pficon-edit"></i>
</span>
```
Is it possible to use a similar markup, but like this instead?:
```
<button class="value" tabindex="0" type="button">
  <span class="sr-only">Edit Name: </span>
  Mike
  <i class="pficon pficon-edit"></i>
</button>
```
Another benefit of this method is that for the sighted keyboard-only user, it's only one tab stop for the field instead of two, since it's all one button. 

Also, I just wanted to double-check: For the classes that you included in this, are all of the patternfly classes identified with "pf-" in the class name? I'm not too familiar with all the bootstrap classes, so it's not obvious to me which ones are bootstrap vs patternfly classes. But we usually differentiate the patternfly classes with a "pf".

jgiardino:
I followed up with @andresgalante about the class names. The feedback I got about class names was:

- all pf related classes should have "pf" somewhere
- "pf" should be included in the name syntax like this in pf 3: `block-pf-element-mod`
  - for example: 
    - `list-view-pf` "list-view" is the block
    - `list-view-pf-checkbox` "checkbox" is an element of "list-view"
    - `list-view-pf-icon-sm` "sm" is a modifier of the element "icon" in the block "list-view"


He also commented that the block name "inline-edit-form-field" is really long for the block name. After chatting about it, we came up with `form-control-pf-editable` where "editable" is a modifier for the block "form-control".

Can you replace `pf-inline-edit-form-field` with `form-control-pf-editable`?

Can you update the class names for the other pf-related classes? Following the list view pattern, I think for most cases you could just include "form-control-pf" in front of the other class names, like:

- `form-control-pf-value`
- `form-control-pf-save`

@matthewcarleton Do you think these class names work?

matthewcarleton:
@jgiardino yup I think this is the right direction, great work!

jgiardino (approved):
These updates look good to me. Great job, @dabeng!

jeff-phillips-18 (approved):

Issue #950, opened 2018-02-05
Tree List View with Muti-Child Expansion


beanh66:
We have a request for the [Tree List View with Multi-Child Expansion](http://www.patternfly.org/pattern-library/content-views/tree-list-view/#design) in Angular. Currently there is no implementation for PF Core, so if possible, we would be looking to get this into core and then follow up with the Angular implementation.  The ADS Team in IT is looking for this component. See related Angular issue here: https://github.com/patternfly/angular-patternfly/issues/702

@dtaylor113 @matthewcarleton @LHinson  FYI

LHinson:
@beanh66 The current pf-list implementation [1] does support the addition of nested rows. That said, after reviewing the implementation, we found that there is a discrepancy between the design and the implementation that we must resolve. I will be leaving for PTO and back the week of 2/19 and have scheduled a meeting for that week to discuss best next steps. Apologies for the delay here but at least wanted to make it known we are working to get this resolved in the coming weeks. 

Let me know if there is any concerns. 

[1] https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/list-pf.html

LHinson:
After our meeting today, we have determined that the design documentation with nested rows will remain until a mobile strategy is defined for PatternFly. You should be all set with the core implementation therefore I would like to close this issue @beanh66. Please let me know if there are any concerns with this.

beanh66:
@LHinson I created this issue originally because I don't see anything on the PF site as far as a core implementation goes: http://www.patternfly.org/pattern-library/content-views/tree-list-view/#code. If there is already an implementation, can we get it up on the main code page?

LHinson:
I've added this to the PF Scrum backlog to contribute to the core code https://patternfly.atlassian.net/browse/PTNFLY-2780

LHinson:
@dabeng is currently working on contributing this

beanh66:
Thanks @LHinson!

dabeng:
Hi @beanh66 @mindreeper2420 @michael-coker , please review demo page:

- [Tree List View - Single Child Expansion](https://rawgit.com/dabeng/patternfly/tree-list-view-dist/dist/tests/tree-list-view-single-child-expansion.html)
- [Tree List View - Multiple Child Expansion](https://rawgit.com/dabeng/patternfly/tree-list-view-dist/dist/tests/tree-list-view-multiple-child-expansion.html)

matthewcarleton:
[...]

PR #980, opened 2018-03-07
feat(tree-list-view): add tree list view

dabeng:
## Description
This PR is targeted to fix issue #950.

@cliffpyles @beanh66 @matthewcarleton  @michael-coker  Please review this PR.

- [Tree List View - Single Child Expansion](https://rawgit.com/dabeng/patternfly/tree-list-view-dist/dist/tests/tree-list-view-single-child-expansion.html)
- [Tree List View - Multiple Child Expansion](https://rawgit.com/dabeng/patternfly/tree-list-view-dist/dist/tests/tree-list-view-multiple-child-expansion.html)


mcoker (approved):
LGTM!

src/less/tree-list-view.less
```diff
@@ -0,0 +1,62 @@
...
+  background-color: @list-group-top-border;
+  border: none;
+  margin-top: 30px;
+  > .list-group-item:first-child {
+    border-top: 1px solid transparent;
+  }
+  .list-group-item {
+    &.active {
+      color: currentColor !important;
+      z-index: auto;
```
matthewcarleton: Why are declaring z-index:auto; here? there is a z-index:2; on the .list-group-item.active class. Am I missing something?

src/less/tree-list-view.less
```diff
@@ -0,0 +1,62 @@
...
+.tree-list-view-pf {
+  background-color: @list-group-top-border;
+  border: none;
+  margin-top: 30px;
+  > .list-group-item:first-child {
+    border-top: 1px solid transparent;
+  }
+  .list-group-item {
+    &.active {
+      color: currentColor !important;
```
jeff-phillips-18: why !important?

beanh66:
@dabeng LGTM!

tests/pages/_includes/widgets/tree-list-view/tree-list-view-rows.html
```diff
@@ -0,0 +1,163 @@
...
+      </div>
+    </div>
+  </div>
+</div>
+<script>
+  $(document).ready(function () {
+    // set up identation for child rows
+    $('.tree-list-view-pf').children().find('.list-group-item').each(function() {
+      var indent = $(this).parentsUntil('.tree-list-view-pf', '.list-group-item').length * 85;
+      $(this).children('.list-group-item-header').css('padding-left', indent);
```
jeff-phillips-18: This works for the test page, but all consumers would need to do the same. Seems to me it would be beneficial to use some html or css to indent the children a specified amount (and 85px seems a bit much, likely horrid at mobile).

jeff-phillips-18 (changes requested):

src/less/tree-list-view.less
dabeng: Hi @matthewcarleton ， `z-index: 2;` will cause that active list-group-item covers the bottom division line.
screen shot 2018-03-09 at 10 14 56 am

src/less/tree-list-view.less
dabeng: yes. you are right. It's not necessary.

src/less/tree-list-view.less
matthewcarleton: 👍

src/less/tree-list-view.less
```diff
@@ -0,0 +1,62 @@
...
+        color: @list-group-link-color;
+        &:hover {
+          background-color: @list-view-hover-bg;
+        }
+        .list-group-item-text {
+          color: @list-group-link-color;
+        }
+      }
+      .list-group-item-text {
+        color: currentColor !important;
```
matthewcarleton: can this !important be removed as well?

dabeng: yes. you are right. it could be removed too.

tests/pages/_includes/widgets/tree-list-view/tree-list-view-rows.html
dabeng: Yes @jeff-phillips-18 You are right. CSS solution should be the priority solution for setting indentation of child rows of tree-list-view. In fact, I also did the **[test](https://rawgit.com/dabeng/patternfly/tree-list-view-css-dist/dist/tests/tree-list-view-multiple-child-expansion.html)**, but finally I found this demo page can not fulfill the needs described on the design doc.

dabeng: Finally, I followed the idea of [TreeGrid Table](https://github.com/patternfly/patternfly/blob/master/src/js/patternfly-functions-tree-grid.js#L68)  to set indentation according to nested depth with js code.

dabeng:
@jeff-phillips-18 CSS solution should be the priority solution for setting indentation of child rows of tree-list-view. In fact, I also did the **[test](https://rawgit.com/dabeng/patternfly/tree-list-view-css-dist/dist/tests/tree-list-view-multiple-child-expansion.html)**, but I found this demo page can not fulfill the needs described on the design doc.
Finally, I followed the idea of [TreeGrid Table](https://github.com/patternfly/patternfly/blob/master/src/js/patternfly-functions-tree-grid.js#L68)  to set indentation according to nested depth with js code.

jeff-phillips-18:
The JS solution does not help adopters. They will not get this JS and would need to dig to find it. I believe there is an HTML/CSS solution that could be done here. Maybe there's an issue trying to have the border between child rows extend to the far left edge but I think it would be better to not extend and have a usable solution.

dabeng:
@jeff-phillips-18 I have removed the hard coding of setting indentation with js code. Please review the updated demo page:
- [Tree List View - Single Child Expansion](https://rawgit.com/dabeng/patternfly/tree-list-view-dist/dist/tests/tree-list-view-single-child-expansion.html)
- [Tree List View - Multiple Child Expansion](https://rawgit.com/dabeng/patternfly/tree-list-view-dist/dist/tests/tree-list-view-multiple-child-expansion.html)

jeff-phillips-18 (approved):

PR #990, opened 2018-03-15
feat(contextswitcher): Change from dropdown to Bootstrap-select

### LINKED ISSUE #1006: Context Selector - can we remove the vertical separator?
In reference to #990 the vertical separator in the context selector is causing some problems. (FYI @tlwu2013 @beanh66)

38058890-52775fd0-3299-11e8-999f-694ccdc59556

@mcarrano @kybaker @matthewcarleton why was a vertical separator introduced as part of the context selector pattern?  Can we get that removed asap?  This is needed for the OpenShift/Tectonic Summit demo

rhamilto:
Fixes #1006

The contextswitcher was originally written using a dropdown, which doesn't provide the correct
functionality.  This PR preserves the existing styles for the dropdown for backward compatibility.

## Changes

* Replace dropdown with Bootstrap-select

https://rawgit.com/rhamilto/patternfly/contextselector-selectpicker-dist/dist/tests/context-selector.html
https://rawgit.com/rhamilto/patternfly/contextselector-selectpicker-dist/dist/tests/context-selector-vertical-nav.html

## PR checklist (if relevant)
Note:  I am not currently able to cross-browser test IE as I don't have full access to BrowserStack at the moment.

- [X] **Preview the PR**: An image or a URL for designer to preview this PR is provided.

### DIFF
--- a/src/less/context-selector.less
+++ b/src/less/context-selector.less
@@ -1,67 +1,119 @@
 //
 // Context Selector
 // --------------------------------------------------
+
+.contextselector-pf {
+  float: left;
+  .bootstrap-select {
+    width: @contextselector-pf-bootstrap-select-width-mobile !important;
+    @media (min-width: @screen-xs-min) {
+      width: @contextselector-pf-bootstrap-select-width-desktop !important;
+    }
+    &.open {
+      background: @navbar-pf-navbar-primary-active-bg-color-stop;
+      > .dropdown-toggle {
+        color: @navbar-pf-vertical-active-color;
+        &:focus {
+          outline: 0 !important; // don't show restored focus ring when open
+        }
+      }
+    }
+    > .dropdown-toggle {
+      background: none;
+      border: 0;
+      box-shadow: none !important;
+      color: @navbar-pf-vertical-color;
+      font-weight: normal;
+      padding-bottom: @contextselector-pf-dropdown-toggle-padding-bottom;
+      padding-left: @contextselector-pf-dropdown-toggle-padding-left;
+      padding-top: @contextselector-pf-dropdown-toggle-padding-top;
+      &:focus {
+        color: @navbar-pf-vertical-active-color;
+        // restore the focus ring
+        // Default
+        outline: thin dotted !important;
+        // WebKit
+        outline: 5px auto -webkit-focus-ring-color !important;
+        outline-offset: -2px !important;
+      }
+      &:hover {
+        background: @navbar-pf-navbar-primary-active-bg-color-stop;
+        color: @navbar-pf-vertical-active-color;
+        outline: 0 !important; // don't show restored focus ring when open
+      }
+      .filter-option {
+        text-overflow: ellipsis;
+      }
+    }
+    .dropdown-menu li a span.text {
+      display: block;
+      overflow-x: hidden;
+      text-overflow: ellipsis;
+      width: 100%;
+    }
+  }
+}
 .layout-pf-fixed .navbar-pf-vertical.navbar-pf-contextselector {
   z-index: @zindex-modal-background;
 }
 .navbar-pf-vertical .nav.contextselector-pf {
-    @media (min-width: @screen-sm-min) {
-      margin-left:@contextselector-pf-margin-left;
-    }
-    border-left:1px solid @color-pf-black-700;
-      .nav-item-iconic {
-        padding:@contextselector-pf-nav-item-iconic-padding;
-        display: flex;
-        align-items: center;
-      }
+  @media (min-width: @screen-md-min) {
+    margin-left: @contextselector-pf-margin-left;
   }
+}
 
-  .contextselector-pf {
-    float:left;
-    &-title {
-      width: @contextselector-title-width-mobile;
-      @media (min-width: @screen-xs-min) {
-        width: @contextselector-title-width-desktop;
-      }
-      white-space: nowrap;
-      display:inline-block;
-      overflow: hidden;
-      text-overflow: ellipsis;
-      line-height: normal;
+// the following rules are deprecated
+.contextselector-pf {
+  float:left;
+  &-title {
+    display: inline-block;
+    line-height: normal;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    width: @contextselector-title-width-mobile;
+    @media (min-width: @screen-xs-min) {
+      width: @contextselector-title-width-desktop;
     }
-    .dropdown {
-      &.open, &:hover {
-        background-color: @navbar-pf-navbar-primary-active-bg-color-stop;
-      }
+  }
+  .contextselector-pf-list {
+    @media (min-width: @screen-sm-min) {
+      max-height: @contextselector-pf-list-max-height;
+      overflow-y: auto;
     }
-    .dropdown-menu {
-      width: 100%;
-      margin-top:0;
+    a {
+      color: @color-pf-black-800;
+      display: block;
     }
-    .form-group {
-      margin: @contextselector-pf-form-group-margin;
-    }
-    .contextselector-pf-list {
-      @media (min-width: @screen-sm-min) {
-        max-height: @contextselector-pf-list-max-height;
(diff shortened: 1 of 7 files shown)


jeff-phillips-18:
@rhamilto Can you run the regression tests for this? I assume there will be some updates.

rhamilto:
> @rhamilto Can you run the regression tests for this? I assume there will be some updates.

Ran against my branch and got 21 failures, none of which I caused (all appear to be related to missing primary nav items from horizontal nav).  Ran against master, got the same result.

rhamilto:
Per @dgutride, the test failures are a known issue and will be fixed with https://github.com/patternfly/patternfly/pull/991.

dabeng:
ipeiahgfhl

There are two highlighted items at the same time. My browser is Chrome Version 66.0.3359.33.

dabeng:
screen shot 2018-03-21 at 10 41 05 am

Maybe this blank space is not expected.

matthewcarleton:
LGTM, other than what @dabeng pointed out, the only thing I saw was the padding on the  .dropown-toggle not using variables. Thanks for taking care of this @rhamilto !

rhamilto:
Thanks, @dabeng and @matthewcarleton.

> There are two highlighted items at the same time.

This is the default functionality of Bootstrap-select.  I'm not sure we'd want to change that for this specific context?

bwnyc0wfj5

> Maybe this blank space is not expected.

Indeed it is not.  Fixed.

screen shot 2018-03-21 at 11 08 20 am

> the padding on the .dropown-toggle not using variables. 

Modified to use vars.

matthewcarleton:
I agree with @rhamilto we should keep the default bootstrap select functionality. I wonder if we need to keep all the deprecated css that I wrote for the old one. It would be nice to just remove it, What am I missing, is there a good reason to keep it? @LHinson what do you think?

rhamilto:
> is there a good reason to keep it?

Semantic versioning says you should keep it since removing it would break anyone using the dropdown version that already shipped.

dgutride:
@rhamilto @matthewcarleton  - I'd agree that we should keep it to prevent releasing a breaking change

matthewcarleton:
ok that makes sense. I wasn't sure if it was being used at all yet, I suppose that doesn't matter though. Thanks for the clarification @rhamilto @dgutride

jeff-phillips-18 (dismissed):

jeff-phillips-18:
@rhamilto Is this still WIP? Just due to cross-browser testing?

rhamilto:
> @rhamilto Is this still WIP? Just due to cross-browser testing?

Yes and yes.  Y'all have a BrowserStack account I can make use of?  @dtaylor113?

tlwu2013:
Hey @rhamilto , I was just chatting with Serena and Colleen to see if we can remove the vertical separator of the context switcher? 

We ran into a tricky alignment issue between the vertical separator and the sidebar:
tricky-misalignment-case

It's tricky because the logo length could be different, so if we can take out the left border of the switcher could help.  Thanks!

serenamarie125 (changes requested):
@rhamilto in the example here:  https://rawgit.com/rhamilto/patternfly/contextselector-selectpicker-dist/dist/tests/context-selector-vertical-nav.html can we change the vertical alignment so that the bottom of the product name and the selected item in the context selector ?  I'm unsure if this is possible. If it's not, can we have the name vertically centered in the top banner?

serenamarie125:
> We ran into a tricky alignment issue between the vertical separator and the sidebar:

@tlwu2013 agreed, the vertical separator will cause an issue for both the cluster & app consoles.  I'm not sure that we should make this fix as part of this PR, because it may hold it up.

I've created a separate issue to track this : #1006

rhamilto:
Like this?

screen shot 2018-03-29 at 8 25 32 am
screen shot 2018-03-29 at 8 25 52 am

Note the utility nav items do not sit on the same baseline as the logo (existing bug), so now that the context picker does, there is a difference between the context picker and the utility nav items

screen shot 2018-03-29 at 8 28 43 am

rhamilto:
(

rhamilto:
> @tlwu2013 agreed, the vertical separator will cause an issue for both the cluster & app consoles. I'm not sure that we should make this fix as part of this PR, because it may hold it up.

I think it's better to get this the way we want it before merging so there is one less deprecated version of the context switcher in PatternFly.

rhamilto:
PR updated with requested changes.  Once https://travis-ci.org/patternfly/patternfly/builds/359839029?utm_source=github_status&utm_medium=notification finishes, the test page will be updated.

beanh66:
Thanks @rhamilto, that looks great!

beanh66 (approved):
Thanks @rhamilto, that looks great!

serenamarie125 (approved):
great thanks @rhamilto LGTM!

src/sass/converted/patternfly/_vertical-nav.scss
```diff
@@ -129,6 +129,8 @@
       line-height: 25px;
       max-width: 120px;
       // If flexbox is supported, do not set max-width, take all space with just some right padding
+      // This generates a known issue on IE11:
+      // https://github.com/patternfly/patternfly/pull/810
```
jeff-phillips-18: this change can be removed as it was done in #1008

rhamilto:
Cross-browser testing complete.  Ready for final review.  [Build is failing](https://travis-ci.org/patternfly/patternfly/builds/361163857?utm_source=github_status&utm_medium=notification), but it's not clear as to why.

matthewcarleton:
@rhamilto this is looking great. I have two minor things to point out. 
1. As per the [design doc](http://www.patternfly.org/pattern-library/navigation/context-selector/#design) there is meant to be an scroll on the list once it reaches a certain number of items (or a fixed height).
2. Do we want the longer strings to be truncated rather than breaking to a second line? Is that default for Bootstrap-select? 
@mcarrano what are your thoughts on these two items?

rhamilto:
> As per the design doc there is meant to be an scroll on the list once it reaches a certain number of items (or a fixed height).

That's possible, but would be an override to existing Bootstrap-select functionality. 

> Do we want the longer strings to be truncated rather than breaking to a second line? Is that default for Bootstrap-select?

That is also the default Bootstrap-select behavior, but I did add `text-overflow: ellipsis` so the text is ellipsized rather than just lopped off.

jeff-phillips-18 (dismissed):

matthewcarleton:
@rhamilto I talked to @mcarrano about the truncated text and he is good with it so that can stay :) 

For the scrolling list I am going by the design doc so if we keep it this way we should update the doc. I'll leave that up to @mcarrano though.

jeff-phillips-18:
@rhamilto Can you rebase please? I'd like to see if #1012 helps determine the build failure.

rhamilto:
Rebased.

jeff-phillips-18 (approved):

jeff-phillips-18:
@matthewcarleton I'd say approve this and then file an issue against either PF design or PF core to update whichever way we go. I think its ok to iterate on this as any changes in order to limit and scroll would not be breaking.

matthewcarleton (approved):
👍 thanks @rhamilto!

mcarrano:
@matthewcarleton So just to make sure that I understand the scrolling behavior, are we saying that the list would never produce a scroll bar? So what if there were 100 items to select from? Would the list just grow indefinitely? I don't think we would want that. Can file a separate issue, if necessary.

rhamilto:
> @matthewcarleton So just to make sure that I understand the scrolling behavior, are we saying that the list would never produce a scroll bar? So what if there were 100 items to select from? Would the list just grow indefinitely? I don't think we would want that. Can file a separate issue, if necessary.

Bootstrap-select will only make the dropdown as big as the available space (the dropdown stays within the viewport and a scrollbar is added to the dropdown).

screen shot 2018-04-03 at 3 34 11 pm

matthewcarleton:
@mcarrano @rhamilto I think this is fine. We should update the design doc to reflect this though.

PR #994, opened 2018-03-21
chore(tests): Setup initial prep for the about-modal regression test

dabeng:
## Description
The about-modal regression test should be watching the about-modal and not the trigger. 

related Jira story - https://patternfly.atlassian.net/browse/PTNFLY-2807.

@cliffpyles @dgutride I used backstopjs option "clickSelector" to trigger about modal. It seems that this solution is lighter than adding js code to the [bottom](https://github.com/patternfly/patternfly/blob/master/tests/pages/vertical-navigation-with-badges.html#L17) of test page. Does this make sense ?


cliffpyles:
That's a great idea @dabeng! Reviewing it now.

cliffpyles (dismissed):
👍

jeff-phillips-18 (approved):

Issue #997, opened 2018-03-22
fix issues with style linting failures


cliffpyles:
Style linting sporadically fails, and as a result the builds fail.

Example: https://travis-ci.org/patternfly/patternfly/builds/356754165

dgutride:
This appears to have been a temporary issue within a stylelint dependency - we can no longer reproduce it, moving to closed.

PR #1012, opened 2018-04-03
fix(build): add stylelint formatter and fix regression entry

dgutride:
## Description
Regression test build isn't working correctly locally - this addresses that and provides a change for stylelinting also

https://github.com/patternfly/patternfly/issues/997


jeff-phillips-18 (approved):
LGTM, let's change this to chore rather than fix since no need for this to create a release.

PR #1033, opened 2018-04-22
fix(listview): update List View active state per design specs

### LINKED ISSUE #1020: Update List View styling to remove "selected" highlight color
Per design issue https://github.com/patternfly/patternfly-design/issues/633 we want to remove the blue background color for selected rows and use only the checkbox state to indicate selection. All other styling should remain as currently implemented.

AdamJ:
## Description
Update the List View active state to match the updated designs specs.

fixes https://github.com/patternfly/patternfly/issues/1020

## Changes

Set the active state of the `list-group-item` to have a transparent background, rather than the blue highlight.

* Remove `background-color: @list-view-active-bg;` from `.list-view-pf:active` and replace with `background-color: transparent;`
* Remove `background: @list-group-top-border;` from `.list-view-pf-view` and replace with `background-color: transparent;`

**Old**
screen shot 2018-04-22 at 1 51 52 pm
screen shot 2018-04-22 at 1 52 10 pm

**New**
screen shot 2018-04-22 at 1 51 28 pm
screen shot 2018-04-22 at 1 51 04 pm

https://rawgit.com/mindreeper2420/patternfly/issue-1020-dist/dist/tests/list-view-rows.html

## PR checklist (if relevant)

- [X] **Cross browser**: works in IE9
- [X] **Cross browser**: works in IE10
- [X] **Cross browser**: works in IE11
- [X] **Cross browser**: works in Edge
- [X] **Cross browser**: works in Chrome
- [X] **Cross browser**: works in Firefox
- [X] **Cross browser**: works in Safari
- [X] **Cross browser**: works in Opera
- [X] **Preview the PR**: An image or a URL for designer to preview this PR is provided.

### DIFF
--- a/src/less/list-view.less
+++ b/src/less/list-view.less
@@ -22,7 +22,7 @@
     }
     &.active {
       color: @list-group-link-color;
-      background-color: @list-view-active-bg;
+      background-color: transparent;
       background-clip: border-box;
       border-color: @list-view-active-border transparent transparent;
       z-index:auto;
@@ -258,7 +258,7 @@
   }
 }
 .list-view-pf-view {
-  background: @list-group-top-border;
+  background: transparent;
   border: none;
   margin-top: 30px;
 }
--- a/src/sass/converted/patternfly/_list-view.scss
+++ b/src/sass/converted/patternfly/_list-view.scss
@@ -22,7 +22,7 @@
     }
     &.active {
       color: $list-group-link-color;
-      background-color: $list-view-active-bg;
+      background-color: transparent;
       background-clip: border-box;
       border-color: $list-view-active-border transparent transparent;
       z-index:auto;
@@ -258,7 +258,7 @@
   }
 }
 .list-view-pf-view {
-  background: $list-group-top-border;
+  background: transparent;
   border: none;
   margin-top: 30px;
 }


matthewcarleton:
@mindreeper2420 can you provide the rawgit for this? Thanks!

AdamJ:
Do Not Merge until final review from the Design Team.

AdamJ:
@matthewcarleton Rawgit link: https://rawgit.com/mindreeper2420/patternfly/issue-1020-dist/dist/tests/list-view-rows.html

matthewcarleton:
@mindreeper2420 @mcarrano  has the dust settled on this one or are we still finalizing what it should be?

mcarrano:
Yes, we can proceed forward with this @mindreeper2420 and @matthewcarleton . My only question is that the demo I see is for List View Rows. Will these changes propagate through all of the different list view pages? There are several in Layouts.

src/less/list-view.less
```diff
@@ -22,7 +22,7 @@
     }
     &.active {
       color: @list-group-link-color;
-      background-color: @list-view-active-bg;
+      background-color: transparent;
       background-clip: border-box;
```
jeff-phillips-18, line 26: For list views without checkboxes that use row selection, how will the selection be indicated?

mcarrano:
@jeff-phillips-18 Can you give me an example of where that's needed? The assumption is that if rows are selectable, we should always use a checkbox.

matthewcarleton:
>Yes, we can proceed forward with this @mindreeper2420 and @matthewcarleton . My only question >is that the demo I see is for List View Rows. Will these changes propagate through all of the different >list view pages? There are several in Layouts.

good question @mcarrano. @mindreeper2420 I believe there is still more work to be here see : https://rawgit.com/mindreeper2420/patternfly/issue-1020-dist/dist/tests/tree-list-view-single-child-expansion.html

jeff-phillips-18:
@mcarrano What about single select only mode? Checkboxes are not a good fit for that. I don't know of any product uses at the moment, but angular-patternfly allows for row selection: http://www.patternfly.org/angular-patternfly/#/api/patternfly.views.component:pfListView

mcarrano:
@jeff-phillips-18 looks like Angular and Angular-ng support this behavior but React does not. The design document does not call for this explicitly, but I agree we need to consider. @jennyhaines, can you let me know what you think? It may be that we need to stick with the blue highlight for selected rows as Angular does now. Does that cause a problem? Can't remember why we decided move move away from that other than it allowed us to represent selected without affecting the background.

AdamJ:
@mcarrano @jeff-phillips-18 would it make sense to close this PR until it is decided what direction to go? IMO, the design docs would need to be updated and then issues also created in the angular repos to match the new behavior as well.

jeff-phillips-18:
@mindreeper2420 makes sense to me

mcarrano:
I'm also fine with that @mindreeper2420 . I will pull together a discussion with the visual team to figure out what we want to do here from a design standpoint.

AdamJ:
I'll close this then and create a new PR with work based off of the upcoming discussions.

PR #1037, opened 2018-04-25
Update Context Selector Example to reflect lastest Masthead design

### LINKED ISSUE #1029: Update Context Selector Example to reflect lastest Masthead design
The Context Selector example ( https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/context-selector.html) need to be updated to reflect the updated Masthead design. Design details can be found here: http://www.patternfly.org/pattern-library/application-framework/masthead/#design

Specific changes required include:
* Remove "v" from Help icon.
* Help and User menu items should display a selected state with the menu is open.

mcoker:
fixes #1029

## Description
Update context selector snippet to reflect new masthead design.

## Changes
* Update navbar item hover/open state CSS
* remove caret from masthead in context selector snippet

https://rawgit.com/michael-coker/patternfly/update-context-selector-masthead-dist/dist/tests/

cc @matthewcarleton

### DIFF
--- a/src/less/variables.less
+++ b/src/less/variables.less
@@ -112,11 +112,13 @@
 @navbar-pf-color:                                                   @color-pf-black-300;
 @navbar-pf-height:                                                  60px;
 @navbar-pf-icon-bar-bg-color:                                       @color-pf-white;
+@navbar-pf-item-hover-bg-color:                                     rgba(255, 255, 255, .14);
+@navbar-pf-item-open-bg-color:                                      rgba(255, 255, 255, .24);
+@navbar-pf-item-active-color:                                       @color-pf-white;
 @navbar-pf-navbar-navbar-brand-min-width:                           270px;
 @navbar-pf-navbar-navbar-brand-padding:                             8px 0 7px;
 @navbar-pf-navbar-navbar-persistent-bg-color:                       @color-pf-black-150;
 @navbar-pf-navbar-navbar-persistent-border-color:                   @color-pf-black-300;
-@navbar-pf-vertical-active-color:                                   @color-pf-white;
 @navbar-pf-vertical-bg-color:                                       #1d1d1d;
 @navbar-pf-vertical-bg-repeat:                                      no-repeat;
 @navbar-pf-vertical-bg-size:                                        auto 100%;
@@ -140,10 +142,10 @@
 @navbar-pf-vertical-nav-item-iconic-padding:                        21px 12px;
 @nav-pf-container-transition-period:                                250ms;
 @nav-pf-menu-transition-period:                                     100ms;
-@nav-pf-vertical-active-color:                                      @color-pf-white;
 @nav-pf-vertical-active-before-color:                               @color-pf-blue-300;
 @nav-pf-vertical-active-bg-color:                                   @color-pf-black-800;
 @nav-pf-vertical-active-border-color:                               @color-pf-black-900;
+@nav-pf-vertical-active-color:                                      @color-pf-white;
 @nav-pf-vertical-active-font-weight:                                600;
 @nav-pf-vertical-active-icon-color:                                 @color-pf-blue-300;
 @nav-pf-vertical-badge-color:                                       @color-pf-white;
--- a/src/less/variables.less
+++ b/src/less/variables.less
@@ -119,6 +119,7 @@
 @navbar-pf-navbar-navbar-brand-padding:                             8px 0 7px;
 @navbar-pf-navbar-navbar-persistent-bg-color:                       @color-pf-black-150;
 @navbar-pf-navbar-navbar-persistent-border-color:                   @color-pf-black-300;
+@navbar-pf-vertical-active-color:                                   @navbar-pf-item-active-color;
 @navbar-pf-vertical-bg-color:                                       #1d1d1d;
 @navbar-pf-vertical-bg-repeat:                                      no-repeat;
 @navbar-pf-vertical-bg-size:                                        auto 100%;
--- a/src/less/context-selector.less
+++ b/src/less/context-selector.less
@@ -10,9 +10,13 @@
       width: @contextselector-pf-bootstrap-select-width-desktop !important;
     }
     &.open {
-      background: @navbar-pf-navbar-primary-active-bg-color-stop;
       > .dropdown-toggle {
-        color: @navbar-pf-vertical-active-color;
+        &,
+        &:hover,
+        &:focus {
+          background: @navbar-pf-item-open-bg-color;
+          color: @navbar-pf-item-active-color;
+        }
         &:focus {
           outline: 0 !important; // don't show restored focus ring when open
         }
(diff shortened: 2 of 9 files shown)


mcoker:
@matthewcarleton added the rawgit test URL.

mcarrano:
Looks great @michael-coker !

matthewcarleton:
@michael-coker looks great! The only things I'm seeing is this is inheriting bootstrap selects behaviour of putting focus on the search when the dropdown is active. In our other dropdowns the dropdown remains in focus when active. Should we consider making this more consistent @mcarrano ?

matthewcarleton (changes requested):
see my comment  ⬆️

mcarrano:
@matthewcarleton I guess this becomes a question of what's the right thing to do from an accessibility standpoint. I don't see a problem with what's implemented here, but would be interested in hearing thoughts from @jgiardino .

matthewcarleton:
ya I agree @mcarrano

mcoker:
@matthewcarleton @mcarrano another discrepancy I just found is the `:focus` state of the context selector has a blue border/outline, and the menus on the right do not. 

The `:focus` state is consistent between all of the masthead dropdown items when you toggle `:focus` in dev tools, however if you do this, you'll see the difference:

1. Click the context selector dropdown twice (open and close it) and move your mouse away - you should see the blue outline/border.
1. Click either the help or user menu the same way (twice to open/close it) and when you move your mouse away, you should not see a blue outline.

However, as mentioned above, toggling the `:focus` state in dev tools creates a blue border/outline for all of those menus.

Not sure why that's happening, but would you like me to look at that as well? If so, those menu links are visually indicated with a light background on `:focus`, which is appearing fine for all of the links. Should they also have a blue border/outline, or should I remove that and just leave the light background as indication that the item has `:focus`?

mcarrano:
All great questions @michael-coker .  To complicate this further, it behaves differently on different browsers. I use Firefox by default and didn't see this, but I can reproduce on Chrome. In fact for Firefox, I don't really seem to get any visual indication of focus like I do on Chrome. 

I guess that if we can suppress the blue focus rectangle on the context selector, that were be preference as it looks odd in the masthead.

mcoker:
@mcarrano ah, good catch! I was only testing in Chrome. The behavior is quite a bit different in Firefox. You can't tab to the help/user menu links in FF, either. @matthewcarleton gave me access to Browserstack and the behavior in IE10/11 and edge is consistent and matches Chrome for the most part (you can tab to all of the menus and the focus state is inicated, though focusing on the context menu has a border/outline and the other links in the masthead do not)

jgiardino:
@michael-coker For keyboard navigation in FF, I think it follows settings you have defined for your Mac System Preferences > Accessibility > Keyboard. I'm happy to compare what I have with yours if it helps, but that might be what's preventing you from navigating with a keyboard in Firefox.

For showing focus, I do think having some visual indication is important. I don't have a strong preference regarding whether we use our own visual styling or use the method that the browser provides. I think arguments can be made in support for both. 

For keyboard behavior when using the context selector, I was reviewing the different variations on the [bootstrap select test page](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/bootstrap-select.html) and there are differences among those variations. For example, the only difference between the Single-Select Dropdown on that test page and the [Dropdown](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/dropdowns.html) (like what we use for the other menus in the masthead) is that focus immediately shifts to the first menu item in the Select menu and focus stays on the toggle in the Dropdown. Other than that, Tab and Arrow keys seem to behave the same. 

But the context selector appears to be based on the Data Live Search variation on the [bootstrap select test page](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/bootstrap-select.html). In this example, I can only shift focus to another menu item using Arrow keys. If I hit the Tab key, the menu will collapse and focus will shift to the next item in the masthead.

Since the inconsistencies in keyboard navigation are not based on this implementation of the bootstrap-select component, but are part of the bootstrap-select component itself, I'm hesitant to suggest any changes. If we changed this implementation, would we also update the bootstrap-select component itself? Since this is pf 3, I'm inclined to keep the bootstrap and 3rd party implementations of bootstrap-like components as they currently are, and revisit this interaction pattern as part of pf next.

mcoker:
Thanks @jgiardino! Other than the inconsistencies mentioned above, everything looks good cross-browser.

matthewcarleton (dismissed):

mcoker:
@mcarrano @jeff-phillips-18 when one of you have a sec, would you mind reviewing this PR?

mcoker:
@jgiardino @srambach @dgutride would one of you mind giving this a review?

mcarrano:
What did we decide about the issue with the focus state @michael-coker . Are we going to leave as is? I'm OK with that unless there is an easy fix.

mcoker:
@mcarrano I was fine with leaving it as is, but I took another look at it and changed the `<a>` links that act as dropdown toggles in the navbar to `<button>`'s instead. It's a quick and easy change, and more semantic. Now the menus are all keyboard accessible (except in Safari, but it never worked in Safari), and the focus/outline behavior is the same on chrome/opera, and although different from chrome/opera, the same on firefox and all versions of IE. I think that's an improvement. Rebuilt the demo link (https://rawgit.com/michael-coker/patternfly/update-context-selector-masthead-dist/dist/tests/)

src/less/variables.less
```diff
@@ -112,11 +112,13 @@
...
 @navbar-pf-height:                                                  60px;
 @navbar-pf-icon-bar-bg-color:                                       @color-pf-white;
+@navbar-pf-item-hover-bg-color:                                     rgba(255, 255, 255, .14);
+@navbar-pf-item-open-bg-color:                                      rgba(255, 255, 255, .24);
+@navbar-pf-item-active-color:                                       @color-pf-white;
 @navbar-pf-navbar-navbar-brand-min-width:                           270px;
 @navbar-pf-navbar-navbar-brand-padding:                             8px 0 7px;
 @navbar-pf-navbar-navbar-persistent-bg-color:                       @color-pf-black-150;
 @navbar-pf-navbar-navbar-persistent-border-color:                   @color-pf-black-300;
-@navbar-pf-vertical-active-color:                                   @color-pf-white;
```
jeff-phillips-18, line 119: Removing this _could_ be considered a breaking change.

mcoker: Ah! Thanks @jeff-phillips-18 this is my first contribution to PF, I appreciate the feedback. Can you clarify what a breaking change would be? I grepped the codebase for that var before committing, but I suppose someone could be using that variable in their own CSS? If so, would it be a better practice to add it back as

```
@navbar-pf-vertical-active-color: @navbar-pf-item-active-color;
```

jeff-phillips-18: Correct. If an adopter is referencing this variable it would break them. Better to leave it and set it as you have above.

mcarrano:
Looks good @michael-coker . I can see the consistency when tabbing around.

src/less/variables.less
mcoker: @jeff-phillips-18 awesome. Added back and updated the branch.

jeff-phillips-18 (approved):

PR #1052, opened 2018-05-14
fix(patternfly-settings.js): Make compatible with NodeJS environment

mturley:
## Description
Don't reference `window` when it isn't present, use `global` instead. This change makes this repo
compatible with being imported in a NodeJS script, which I want to use for some snapshot testing
experiments.

## Changes

* In the three files which compile down to patternfly-settings.js, I used a ternary expression in the self-executing function expressions, where there was `(window)` it now passes `(typeof window !== 'undefined' ? window : global).
* In the top of each of those same closures, I added `const patternfly = window.patternfly` to make sure the passed-in window argument was actually being used, in all the following `patternfly.*` assignments (so far this just happens to work because in a browser `patternfly` is an alias for `window.patternfly`).

I don't really have a link, but I can show that my build is still passing.

### DIFF
--- a/src/js/patternfly-settings-base.js
+++ b/src/js/patternfly-settings-base.js
@@ -13,4 +13,4 @@
 
   window.patternfly = patternfly;
 
-})(window);
+})(typeof window !== 'undefined' ? window : global);
--- a/src/js/patternfly-settings-charts.js
+++ b/src/js/patternfly-settings-charts.js
@@ -1,6 +1,9 @@
 (function (window) {
   'use strict';
 
+  // Ensure we are assigning these to the patternfly property of the window argument, and not the implicit global patternfly
+  var patternfly = window.patternfly;
+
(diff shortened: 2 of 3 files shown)


jeff-phillips-18:
Why the sass file changes?

mturley:
Whoa, that is weird... I must have fat fingered my commit, sorry about that @jeff-phillips-18 . Will fix in the morning

jeff-phillips-18:
@mturley could be that the sass was not generated in a previous PR

jeff-phillips-18:
@mturley I was correct, those files were not updated. A rebase should remove them from this PR.

jeff-phillips-18:
@mturley Can you rebase please?

jeff-phillips-18 (dismissed):

jeff-phillips-18 (approved):

dgutride (approved):

Issue #947, opened 2018-02-02
TreeGrid table is slow with large datasets 


jamesnetherton:
I'm working with a TreeGrid table which has around 600 items. It's quite slow to do the initial tree view render and when subsequently expanding or collapsing any of the tree nodes.

After a brief bit of profiling, it seems much of the time is being spent [here](https://github.com/patternfly/patternfly/blob/master/src/js/patternfly-functions-tree-grid.js#L10).

Does it need to use `closest`? Parent is the id attribute of some `<tr>` element that should be considered the parent tree node. Isn't this unique? If so, instead of `closest`, maybe we use `parent = $(parent)`. It performs much better.

dabeng:
Hi @jamesnetherton , I agree with you. `rows.closest(parent)` will bring performance impact. It's better to replace `rows.closest(parent)` with `$(parent)`

@dlabrecq @jeff-phillips-18 @cliffpyles  Any thoughts ?

dlabrecq:
Not certain $(parent) is the correct syntax here, but .parent() should return the direct parent of a node. You can also do something more like $(“child”).parent();

jamesnetherton:
Not sure `$(parent)` or `$("child").parent() ` is correct either. Parent is set like this:

```javascript
var parent = node.attr('data-parent');
```
A typical value will look like `#some-parent-id`. We just need to do an element look up against that value.   So `$(parent)` should be ok (I think). I hacked my local copy of patternfly.js to use this and it seems to be working fine.

dabeng:
@LHinson @cliffpyles I did a [testing](https://codepen.io/dabeng/pen/xWRXyX) with Chrome DevTools. The Demo TreeGrid Table has more than 400 rows. I found that it will take 1000 ms to collapse the child rows of "Parent 1".

![treegrid table testing](https://user-images.githubusercontent.com/5689075/37580640-d534cc58-2b7f-11e8-9dbf-9738db21ece6.gif)

<img width="671" alt="screen shot 2018-03-19 at 2 07 14 pm" src="https://user-images.githubusercontent.com/5689075/37580730-3fbcbb58-2b80-11e8-9a77-b2f74b35e93c.png">

dabeng:
[TreeGrid Table Testing2](https://codepen.io/dabeng/pen/wmopwd) works well as expected.

[...]

PR #1054, opened 2018-05-15
fix(treegrid): Refactor the treegrid component

dabeng:
## Description
This PR is targeted to fix issue #947

## Changes

Just replace the low efficient selector with a higher one

Feel free to experience the performance difference between the following demos:
- [TreeGrid Table Testing](https://codepen.io/dabeng/pen/xWRXyX)
-  [TreeGrid Table Testing2](https://codepen.io/dabeng/pen/wmopwd)

Hi @jeff-phillips-18 , please review this PR whenever you get the chance 😊

### DIFF
--- a/src/js/patternfly-functions-tree-grid.js
+++ b/src/js/patternfly-functions-tree-grid.js
@@ -7,7 +7,10 @@
 
     if (typeof parent === "string") {
       if (isNaN(parent)) {
-        parent = rows.closest(parent);
+        parent = $(parent);
+        if (parent.length > 1) {
+          parent = rows.closest(parent);
+        }
       } else {
         parent = $(rows[parseInt(parent, 10)]);
       }


dgutride (dismissed):

src/js/patternfly-functions-tree-grid.js
```diff
@@ -7,7 +7,7 @@
 
     if (typeof parent === "string") {
       if (isNaN(parent)) {
-        parent = rows.closest(parent);
+        parent = $(parent);
```
jeff-phillips-18, line 10: I fear applications may be counting on the use of `closest` and not be using unique `data-parent` selectors. Can we add a check for multiple results and if there are multiples revert to using `closest`?

dabeng: Hi @jeff-phillips-18 . In this sentence, "parent" is just "#id".
`parent = rows.closest(parent);`

jeff-phillips-18: not necessarily, it is whatever the `data-parent` attribute is set to. Could be an id, maybe should be an id, but it doesn't _have_ to be an id. There may be applications out there using a selector other than id and that selector might not be unique.

dabeng: Yes. You are right Jeff :blush:  variable "parent" could be an id but it doesn't have to be an id.

If we review the source code of treegrid table example, we will find that there are two types of “data-parent” attribute as shown below. One is row index(int), the other is row id(string) .

screenshot from 2018-05-18 07-43-00

So the author of treegrid plugin use the following code block to handle the scenario of row index:

screenshot from 2018-05-18 08-07-06


and use the following code block to handle the use case of row id:

screenshot from 2018-05-18 08-08-34

Additional info: original expression here is `rows.closest(parent);` instead of `$(parent)`

jeff-phillips-18: I'm just saying... it used to use `closest`, your change removes that and _expects_ the selector to be unique. _IF_ there are _any_ applications out there already using this but use a selector that is _not_ unique, then this is a breaking change.

If we add a check to ensure the selector is unique (length of 1) and if not revert to using _closest_, then this is _not_ a breaking change.

jeff-phillips-18 (changes requested):
Convinced myself that we should add the check :wink

src/js/patternfly-functions-tree-grid.js
skateman: I agree with @jeff-phillips-18

dabeng:
Hi @jeff-phillips-18 I have added check according to your advice.

src/js/patternfly-functions-tree-grid.js
```diff
@@ -7,7 +7,7 @@
 
     if (typeof parent === "string") {
       if (isNaN(parent)) {
-        parent = rows.closest(parent);
+        parent = parent.charAt(0) === '#' ? $(parent) : rows.closest(parent);
```
jeff-phillips-18: non-id selectors can also have just a single result which would not require the `closest` call. I suggest we only use the `closest` call if the selector returns multiple elements:

```
parent = $(parent);
if (parent.length > 1) {
  parent = rows.closest(parent);
}
```

May cause an extra lookup but only if the application did not use a unique selector.

jeff-phillips-18 (changes requested):

dabeng:
Hi @jeff-phillips-18 I have updated PR according to your advice.

jeff-phillips-18 (approved):

dgutride (approved):

PR #1049, opened 2018-05-07
Update applauncher masthead

mcoker:
## Description
Update masthead in app launcher examples to reflect new masthead design http://www.patternfly.org/pattern-library/application-framework/masthead/#design

fixes https://github.com/patternfly/patternfly/issues/1031

## Changes

Write a list of changes the PR introduces
* remove caret from help menu
* change app launcher dropdown menu from infotip to standard dropdown
* update app launcher's hover/open states to match the help and user links

## Link to rawgit and/or image

https://rawgit.com/michael-coker/patternfly/update-applauncher-masthead-dist/dist/tests/application-launcher.html

## PR checklist (if relevant)

- [X] **Cross browser**: works in IE9
- [x] **Cross browser**: works in IE10
- [x] **Cross browser**: works in IE11
- [x] **Cross browser**: works in Edge
- [x] **Cross browser**: works in Chrome
- [x] **Cross browser**: works in Firefox
- [x] **Cross browser**: works in Safari
- [x] **Cross browser**: works in Opera
- [x] **Responsive**: works in extra small, small, medium and large view ports.
- [x] **Preview the PR**: An image or a URL for designer to preview this PR is provided.

PR #1056, opened 2018-05-18
fix(rcue): Adds missed changes to RCUE sass files from #1049

jeff-phillips-18:
Some RCUE sass files were not generated due to the timing of the rcue addition and modifications
made in #1049


mcoker (approved):
Looks good

AdamJ (approved):

PR #1072, opened 2018-06-06
feat(release): remove tests directory from package.json

### LINKED ISSUE #1067: Remove unnecessary test files from npm package
The test directory is also being included in our npm package - it should be ignored when creating the package (but left as part of the dist so test pages will continue to work).

dgutride:
## Description
Remove the tests directory from the npm package bundle

Fixes #1067

## Changes

* Test directories are no longer included in the npm package

### DIFF
--- a/package.json
+++ b/package.json
@@ -136,8 +136,12 @@
     }
   },
   "files": [
-    "dist/",
-    "tests/",
+    "dist/css/",
+    "dist/fonts/",
+    "dist/img/",
+    "dist/js/",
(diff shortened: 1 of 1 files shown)


dlabrecq (dismissed):

jeff-phillips-18:
I think this should be a feat. Just in case some application requires the tests directory. At least they can pin to a version that contains the tests until they are ready to move forward.

jeff-phillips-18 (approved):

Issue #1064, opened 2018-05-29
Component : Message Dialog


serenamarie125:
Create a message dialog component.
See requirements in the PF React Issue : https://github.com/patternfly/patternfly-react/issues/360

Corresponding PatternFly design issue: https://github.com/patternfly/patternfly-design/issues/632

serenamarie125:
@michaelkro please take this on!
@mcarrano FYI - Mike it taking this before the PF React contribution

PR #1074, opened 2018-06-13
feat(message-dialog): Add the message-dialog pattern

michaelkro:
## Description
The message dialog pattern is built on `modal`, and consequently, almost all of the introduced CSS are overrides to `modal` styles dictated by [the specs outlined here](https://github.com/patternfly/patternfly-design/issues/632#issuecomment-396402810)

**Notes**
1. SASS/LESS nesting is used here to scope all styles/overrides to `.message-dialog-pf`
2. Test examples based on [original samples](https://github.com/patternfly/patternfly-react/issues/360#issuecomment-392789901)
3. Closes https://github.com/patternfly/patternfly/issues/1064

## Link to rawgit

[rawgit](https://rawgit.com/michaelkro/patternfly/message-dialog-rawgit/message-dialog-dist/dist/tests/message-dialogs.html)

## PR checklist (if relevant)


michaelkro:
@serenamarie125

serenamarie125:
@michaelkro a few requests:

1. Clicking the primary action button on all of the examples should close the dialog
1. Autofocus on the primary button for Non Destructive, Message, Success & Min Height Dialog examples
1. Autofocus on the secondary button for Error, Delete Confirmation, Warning, Question examples
1. Once a button has focus, tab should traverse between buttons.  But disregard the "x"

mcarrano:
@matthewcarleton @deernano I tagged you guys to also take a look at this for PF Core contribution. We don't yet have an approved design, but just want to make sure there are no disconnects with current direction. My only question is around the recommended button lables, i.e. "Close" vs. "OK" for acknowledgement. @deernano please verify.

deernano:
@mcarrano  I would prefer ok than close. Ok is more like the action after the user is acknowledged.

deernano (dismissed):
Looks good to me

jeff-phillips-18:
Something other that the buttons is in the tab order, but I'm not sure what.

src/less/modals.less
```diff
@@ -36,3 +36,75 @@
...
+.message-dialog-pf {
+  // message-dialog header
+  .modal-header {
+    align-items: center;
+    display: flex;
+    height: 40px;
+    justify-content: space-between;
+    padding-bottom: 0;
+    padding-left: 20px;
+    padding-right: 18px;
```
jeff-phillips-18: padding-right should already be 18px.

src/less/modals.less
```diff
@@ -36,3 +36,75 @@
...
+    &:before {
+      content: none;
+    }
+  }
+
+  // message-dialog title
+  .modal-title {
+    display: inline-block;
+    font-weight: bold;
+  }
```
jeff-phillips-18: if we set a line-height of 20px here, we could remove the height, align-items, and padding top/bottom from the .modal-header settings.

src/less/modals.less
```diff
@@ -36,3 +36,75 @@
...
+
+  // message-dialog close button
+  .close {
+    color: #616161;
+    &:hover {
+      color: #363636;
+    }
```
jeff-phillips-18: The colors here should us colors from the palette and use variables. Is there a reason why message-dialog is different from modal here?

src/less/modals.less
```diff
@@ -36,3 +36,75 @@
...
+  .message-dialog-pf-icon {
+    display: flex;
+    flex-direction: column;
+    font-size: 30px;
+    margin-right: 15px;
+  }
+
+  // Color override for info icon
+  .pficon-info {
+    color: @color-pf-black-500;
```
jeff-phillips-18: We use @color-pf-black-700 for the alert, should this be the same?

jeff-phillips-18 (commented):
In general, I question why the message dialog has different margins/colors from modal. I see the spec you are working from specifies these settings but I'm not sure they need to vary from the existing modal settings. @jennyhaines ?

src/less/modals.less
matthewcarleton: There are alot of things here that I don't think are necessary. We should not apply height here. I'm curious why we are reapply any of this when the modal already has it? Shouldn't the styles be consistent with what we already have?

src/less/modals.less
```diff
@@ -36,3 +36,75 @@
...
+  .modal-footer {
+    margin-top: 0;
+    padding-right: 15px;
+    padding-top: 20px;
+  }
+
+  // message-dialog icon
+  .message-dialog-pf-icon {
+    display: flex;
+    flex-direction: column;
```
matthewcarleton: I would remove flex from here and add `line-height:0;`

src/sass/converted/patternfly/_modals.scss
```diff
@@ -36,3 +36,75 @@
...
+    display: flex;
+    height: 40px;
+    justify-content: space-between;
+    padding-bottom: 0;
+    padding-left: 20px;
+    padding-right: 18px;
+    padding-top: 0;
+    &:after,
+    &:before {
+      content: none;
```
matthewcarleton: what is the purpose of this?

src/sass/converted/patternfly/_modals.scss
```diff
@@ -36,3 +36,75 @@
...
+    }
+  }
+
+  // message-dialog title
+  .modal-title {
```
matthewcarleton: I don't believe either of these are necessary.

tests/pages/_includes/widgets/communication/message-dialog-delete-confirmation.html
```diff
@@ -0,0 +1,44 @@
+<button class="btn btn-default" data-toggle="modal" data-target="#{{ include.messageDialogId }}">Delete Confirmation Dialog</button>
+<div class="modal fade" id="{{ include.messageDialogId }}" tabindex="-1" role="dialog" aria-labelledby="myDeleteDialogLabel" aria-hidden="true">
+  <div class="modal-dialog">
+    <div class="modal-content message-dialog-pf">
```
matthewcarleton: `.message-dialog-pf` should be with the `.message-dialog`

tests/pages/_includes/widgets/communication/message-dialog-delete-confirmation.html
```diff
@@ -0,0 +1,44 @@
...
+      <div class="modal-header">
+        <h4 class="modal-title" id="myDeleteDialogLabel">Modal Title Here</h4>
+        <button type="button" class="close" data-dismiss="modal" aria-hidden="true" tabindex="-1">
```
matthewcarleton: I'm not sure about the `tabindex="-1"` here.

tests/pages/_includes/widgets/communication/message-dialog-delete-confirmation.html
```diff
@@ -0,0 +1,44 @@
...
+      <div class="modal-body message-dialog-pf-body">
+        <div class="message-dialog-pf-icon">
+          <span class="pficon pficon-error-circle-o"></span>
+        </div>
+        <div class="message-dialog-pf-text">
+          <div class="message-dialog-pf-text-primary">
+            Main dialog text
+          </div>
+          <div class="message-dialog-pf-text-secondary">
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do
```
matthewcarleton: This should be wrapped in a`<p>`

tests/pages/_includes/widgets/communication/message-dialog-delete-confirmation.html
```diff
@@ -0,0 +1,44 @@
...
+          <span class="pficon pficon-close"></span>
+        </button>
+      </div>
+      <div class="modal-body message-dialog-pf-body">
+        <div class="message-dialog-pf-icon">
+          <span class="pficon pficon-error-circle-o"></span>
+        </div>
+        <div class="message-dialog-pf-text">
+          <div class="message-dialog-pf-text-primary">
+            Main dialog text
```
matthewcarleton: this should use a heading

matthewcarleton:
@michaelkro thanks for doing this! I've left some comments. 

One overall question I have is why we aren't using the styles that we have for the modal already? Are they different for the message dialog? If so, I think we should consider realigning the design to make it more consistent.

@serenamarie125 can you explain why the desired outcome is not being able to tab to the "x"? I think that's what you requested Serena unless I misinterpreted it.

serenamarie125:
@matthewcarleton that was my request (not tabbing to the "X") - I thought we did the same with the wizard but I could be wrong.  If the "X" is a short cut for the "Cancel"/"Close/OK" button, I'm not sure why it would be included in the tab order, but happy to defer on this !

serenamarie125:
@matthewcarleton did you see the specs which are referenced in the initial description?  Jenny produced these a few days ago, as part of the new pattern designs.

src/sass/converted/patternfly/_modals.scss
michaelkro: The pseudo-elements were interfering with my `justify-content: space-between`

I was definitely a little iffy about this, but not sure how to push the header contents to the ends of the div.  Any suggestions to do this without removing the elements?  I assume they are there for a reason lol

michaelkro:
I've updated the following
* CSS/HTML cleanup
* Put the header close button back into tab sequence

Will wait for further direction regarding `modal` spacing/color overrides

michaelkro:
btw, thank you all for the detailed reviews, I appreciate it so much!!!

matthewcarleton:
@michaelkro I have spoken to @mcarrano and @jgiardino this morning about this and we are all on the same page that the styles that are over writing the default styles for the modal should be removed. I can comment further on the code if you'd like or if you want to chat quickly just let me know. 
@serenamarie125 I also talked to @jgiardino about the "X" and she agrees that it should receive focus.

jeff-phillips-18:
Can you update the rawgit link?

michaelkro:
I just did @jeff-phillips-18 :)

jeff-phillips-18:
I liked the message title and body layout before. It kept things centered with the icon when there was little text or top justified if there was enough text.

Seems to me the `Message dialog title` text should be bold.

src/sass/converted/rcue/_modals.scss
```diff
@@ -36,3 +36,9 @@
     }
   }
 }
+
+// message-dialog icon
+.message-dialog-pf-icon {
+  font-size: 30px;
+  line-height: 0;
```
matthewcarleton: you should be able to remove this now. It is using the line-height from the .pf-icon class.

src/less/modals.less
```diff
@@ -36,3 +36,9 @@
     }
   }
 }
+
+// message-dialog icon
+.message-dialog-pf-icon {
+  font-size: 30px;
+  line-height: 0;
```
matthewcarleton: you should be able to remove this now. It is using the line-height from the .pf-icon class.

michaelkro:
@jeff-phillips-18 are you referring to the primary text, or the modal header title? (for boldness)

jeff-phillips-18:
The primary text

jgiardino:
I'm not sure that text should be bold.

Here are two variations, one with bold, one without, of how I imagine text could be defined in these dialgos. I am making an assumption that the bulk of the message is in the first paragraph with larger text, and that the smaller text only really displays if supplemental text is necessary. But these assumptions could be totally wrong. 

image

image

mcarrano:
The intended design is outlined here https://github.com/patternfly/patternfly-design/issues/632 Let's try to remain consistent with that (other than following the default modal styling as discussed above). The summary text is not bold.

tests/pages/_includes/widgets/communication/message-dialog-delete-confirmation.html
```diff
@@ -0,0 +1,42 @@
...
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-hidden="true">
+          <span class="pficon pficon-close"></span>
+        </button>
+        <h4 class="modal-title" id="myDeleteDialogLabel">Modal Title Here</h4>
+      </div>
+      <div class="modal-body message-dialog-pf-body">
+        <span aria-hidden="true" class="pficon pficon-error-circle-o message-dialog-pf-icon"></span>
+        <div class="message-dialog-pf-text">
```
jeff-phillips-18: `message-dialog-pf-text` is no longer defined. Do we need this div at all?

michaelkro: It's for the `flex` layout of `message-dialog-pf-body`, to keep the primary and secondary text nodes grouped as a single flex child

jeff-phillips-18: OK, right. But we can remove that class

michaelkro: Ok, yes, sorry, force of habit.

Taking that out now!

jeff-phillips-18 (dismissed):

tests/pages/_includes/widgets/communication/message-dialog-delete-confirmation.html
```diff
@@ -0,0 +1,42 @@
+<button class="btn btn-default" data-toggle="modal" data-target="#{{ include.messageDialogId }}">Delete Confirmation Dialog</button>
+<div class="modal fade" id="{{ include.messageDialogId }}" tabindex="-1" role="dialog" aria-labelledby="myDeleteDialogLabel" aria-hidden="true">
+  <div class="modal-dialog">
```
matthewcarleton: I wonder if it makes more sense to have the `.message-dialog-pf` here and use that to target the `.modal-body` and the `.pficon` rather than adding classes for each. I think from a users perspective this will be easier to implement. @jeff-phillips-18 @jgiardino what do you think?

jeff-phillips-18: Not sure we can specify just .pficon and assume it is the dialog icon. But I could agree with this for the `.modal-body`.

matthewcarleton: true. what about `.message-dialog-pf .modal-body .pficon` ? I like the idea of reducing the amount of unique markup for the modal.

jeff-phillips-18: My concern is folks adding icons in their message text. Might be unlikely. May also need to account for `.fa` as well though.

matthewcarleton: ya that's a good point. I'd be ok with leaving the icon class. But also adding it to the main wrap and removing it from the modal-body. @michaelkro are you ok with this approach?

michaelkro: Sure!  I think I am picking up what you guys are putting down...  I'll push a commit and see if I get it right lol

src/less/modals.less
```diff
@@ -36,3 +36,15 @@
...
   }
 }
+
+// message-dialog pattern
+.message-dialog-pf {
+  .modal-body {
+    display: flex;
+  }
+
+  &-icon {
```
jeff-phillips-18: I think this should be: 
```
  .fa,
  .pficon {
      font-size: 30px;
      margin-right: 15px;
  }
```

tests/pages/_includes/widgets/communication/message-dialog-delete-confirmation.html
```diff
@@ -0,0 +1,42 @@
...
+  <div class="modal-dialog message-dialog-pf">
+    <div class="modal-content">
+      <div class="modal-header">
+        <button type="button" class="close" data-dismiss="modal" aria-hidden="true">
+          <span class="pficon pficon-close"></span>
+        </button>
+        <h4 class="modal-title" id="myDeleteDialogLabel">Modal Title Here</h4>
+      </div>
+      <div class="modal-body">
+        <span aria-hidden="true" class="pficon pficon-error-circle-o message-dialog-pf-icon"></span>
```
jeff-phillips-18: No longer need `message-dialog-pf-icon`

jeff-phillips-18 (changes requested):

src/less/modals.less
michaelkro: I think that will interfere with the header close icon though, right?

jeff-phillips-18: yeah, it should be under `.modal-body`

jeff-phillips-18 (approved):

serenamarie125 (approved):
@michaelkro LGTM thanks for doing this!

Issue #1082, opened 2018-06-22
Correct RGB colors of list view doesn't take effect


dabeng:
It seems that the css code of  “States and Row Styling” section has not been merged into the master branch.

<img width="1114" alt="screen shot 2018-06-21 at 11 35 33 am" src="https://user-images.githubusercontent.com/5689075/41754301-6d65da16-7604-11e8-8e6c-5f38db63c627.png">

<img width="1063" alt="screen shot 2018-06-22 at 10 10 30 am" src="https://user-images.githubusercontent.com/5689075/41754347-9b51bddc-7604-11e8-8883-a91bdce7b671.png">

mcarrano:
@matthewcarleton I thought we were all set with this but perhaps not. I know there was an earlier PR that @mindreeper2420 was working on that we closed. If this is something we need to fix, can you pick this up @dabeng ?

dabeng:
@mcarrano I made a [commit](https://github.com/dabeng/patternfly/commit/cf071ff0ced4b8998f46bdcd96b49a755d047b79) to fix this issue. You can find the expected colors on the [demo pages](https://rawgit.com/dabeng/patternfly/list-view-colors-dist/dist/tests/index.html) with keyword "list" or "list view"

<img width="1124" alt="screen shot 2018-06-25 at 3 07 51 pm" src="https://user-images.githubusercontent.com/5689075/41835307-aeab4254-7889-11e8-8d68-282badebb518.png">

mcarrano:
Hmmm... This does not look right. I thought that the hover state was intended to be the lighter blue. It's possible that we got these reversed in the design doc. @jennyhaines can you confirm this? Please double check the color values entered here: http://www.patternfly.org/pattern-library/content-views/list-view/#design

jennyhaines:
@mcarrano - they aren't looking right to me either. Currently, the color values in the design documentation say:

_"Hover - when the user hovers over a row (pf-blue-50, #def3ff)
Selected - when the user selects a row via single click or selecting the checkbox (#edf8ff)
Open - when a row is expanded (##edf8ff)
Edit - when a row is in edit mode via inline edit (##edf8ff)
Disabled - when a row is disabled (pf-black-100, #fafafa)"_

[...]

PR #1085, opened 2018-06-27
fix(list view): Adjust bg colors of list view

dabeng:
## Description
This PR is targeted to fix issue #1082.

## Changes

The final background color solution of list view is shown below:

- Hover - when the user hovers over a row (#edf8ff)
- Selected - when the user selects a row via single click or selecting the checkbox (pf-blue-50, #def3ff)
- Open - when a row is expanded (pf-blue-50, #def3ff)
- Edit - when a row is in edit mode via inline edit (pf-blue-50, #def3ff)
- Disabled - when a row is disabled (pf-black-100, #fafafa)

You can find the expected colors on the [demo](https://rawgit.com/dabeng/patternfly/list-view-colors-dist/dist/tests/index.html) pages with keyword "list view"

### DIFF
--- a/src/less/list-group.less
+++ b/src/less/list-group.less
@@ -11,6 +11,11 @@
 .list-group-item {
   border-left: 0;
   border-right: 0;
+  &.disabled,
+  &.disabled:hover,
+  &.disabled:focus {
+    background-color: @list-group-disabled-bg;
+  }
 }
 
 .list-group-item-heading {
--- a/src/less/list-view.less
+++ b/src/less/list-view.less
@@ -16,7 +16,7 @@
     padding-bottom: 0;
     padding-top: 0;
     &.list-view-pf-expand-active {
-      background-color: @list-view-hover-bg;
+      background-color: @list-view-active-bg;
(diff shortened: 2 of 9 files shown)


jeff-phillips-18 (approved):
Is it possible to add a backstop test for active/disabled rows? If one existed, this change should have caused a failure.

dabeng:
@mcarrano @jennyhaines Could you give the color solution when one list-group-item includes ".active" and "diabled" classes ?

mcarrano (approved):
Looks good to me @dabeng

dabeng:
Hi @jeff-phillips-18 . I dug into this testing demand and research the related repos(BackstopJS, node-resemble.js and resemble.js) and found that BackstopJS can not identify subtle changes of colors due to this [line](https://github.com/lksv/node-resemble.js/blob/master/resemble.js#L51) of node-resemble.js

screen shot 2018-06-28 at 3 03 27 pm

The background color of left side is rgb(237, 248, 255) and background color of right side is rgb(222, 243, 255). The two images are reported as being 100% identical.

jeff-phillips-18:
OK, thanks for looking into it @dabeng

PR #1097, opened 2018-07-18
fix(forms): Remove hyphens from inclusion variables

### LINKED ISSUE #924: examples not working with hyphenated variables
Same problem as #738, I haven't properly checked other pages. This affects expand/collapse functionality and error feedback (both in forms.html) but only in a local build using liquid.

There is also another problem with [navbar-vertical.html](https://github.com/patternfly/patternfly/blob/74cdae7f189969478663d88b5801efc2a57e51cd/tests/pages/_includes/widgets/layouts/navbar-vertical.html#L42-L44). Even with the variable fixed the notification drawer stays hidden because the page doesn't include the bell button, e.g. [here](http://rawgit.com/patternfly/patternfly/master-dist/dist/tests/vertical-navigation-with-tertiary-no-icons.html). I'm not sure what the purpose of [this code](https://github.com/patternfly/patternfly/blob/74cdae7f189969478663d88b5801efc2a57e51cd/tests/pages/_includes/widgets/navigation/vertical-navigation.html#L2-L6) is, since the drawer gets included either way.

This should be all, I haven't found any other hyphenated variables. (Interestingly things like `page.notification-drawer` work fine with hyphens.)

michpetrov:
Removed hyphens from inclusion variables as they cause build errors when using liquid.

closes #924

### DIFF
--- a/tests/pages/_includes/widgets/forms/expand-collapse-section.html
+++ b/tests/pages/_includes/widgets/forms/expand-collapse-section.html
@@ -1,52 +1,52 @@
 <form class="form-horizontal">
   <div class="form-group">
-    <label class="col-sm-{{include.col-size-label}} control-label" for="{{include.id-blueprintname}}">Blueprint Name</label>
-    <div class="col-sm-{{include.col-size-field}}">
-      <input type="text" id="{{include.id-blueprintname}}" class="form-control">
+    <label class="col-sm-{{include.col_size_label}} control-label" for="{{include.id_blueprintname}}">Blueprint Name</label>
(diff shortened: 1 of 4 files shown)


jeff-phillips-18:
Can we use underscores or camel case? I find it hard to read with no word differentiation.

LHinson:
@michpetrov any concerns with the the requested changes?

michpetrov:
Sorry for the delay, looks like underscored are fine, I've changed the PR.

jeff-phillips-18 (approved):

dgutride (approved):

Issue #876, opened 2017-12-08
Release notes include broken HTML markup


rhwood:
Since 3.26.3, the release notes have included a broken HTML markup fragment ```<a name"VERSION">``` where VERSION is the current version. This probably just needs to be removed.

LHinson:
this was a bug with the semantic release https://github.com/semantic-release/semantic-release/issues/363

we should fix globally

AdamJ:
PatternFly-ng already uses the latest major release, so there will be no corresponding PR for that repo.

LHinson:
Thanks @mindreeper2420 I'm going to close this.

PR #1098, opened 2018-07-18
fix(dependencies): bump minimum version for semantic release

AdamJ:
## Description
This PR bumps the minimum version of the Semantic Release package. This is due to the error with incorrect HTML rendering in the automated release messages.

This issue was fixed in the package update for Semantic Release from September 2017.

The PR addresses issue https://github.com/patternfly/patternfly/issues/876

A matching PR has been filed to the PatternFly Release Repo: https://github.com/patternfly/patternfly-eng-release/pull/87

## Changes

* Update Semantic Release package from a minimum of 6.3.6 to 15.0.0.


package.json
```diff
@@ -104,7 +104,7 @@
     "start-patternfly": "grunt build-patternfly serve",
     "start-rcue": "grunt build-rcue serve",
     "grunt": "grunt",
-    "semantic-release": "semantic-release pre && npm publish && semantic-release post",
+    "semantic-release": "semantic-release pre && npm prepare && semantic-release post",
```
dlabrecq: I don't believe these commands will publish a package. 

1. The `pre/post` commands don't appear to be valid with semantic-release 15.x
2. `npm prepare` isn't a valid option for `npm`
3. The latest semantic version requires node 8+

dlabrecq (changes requested):

package.json
AdamJ: @dlabrecq I have another PR (https://github.com/patternfly/patternfly/pull/1100), which you commented on, that upgrades to node 8+. Would it make more sense to combine the two, or just have the node update merged first?

Regarding the `pre/post` commands, I've changed them over to `prepare/publish`, per the semantic release docs (https://github.com/semantic-release/npm#prepare).

dlabrecq:
Unless the node 8 PR is committed first, this PR will not build as is. 

Where are you referencing the prepare/publish commands from?

AdamJ:
@dlabrecq I updated the node 8 PR per your comments. Regarding the commands, the `prepare/publish` came from the Semantic Release documentation and, I believe, are included commands in the package.

dlabrecq:
Both prepare/publish are not valid via the command line. (I say that because I just implemented the latest semantic release for patternfly-ng.) However, there are prepare/publish plugins defined in package.json. 

For reference, you can look at the package.json of patternfly-ng and patternfly-next. ~~We'll also need to modify travis.yml to call semantic release directly instead of using the patternfly-eng-release script because we no longer call npm publish directly.~~

~~You might also consider testing this with a dummy npm package, run via your own fork? As with patternfly-ng and patternfly-next, it took a few iterations to nail down.~~

dlabrecq:
~~One more thing to note is that the Travis build also updates the package.json version number committed to master-dist, which this is required for bower installs.~~

~~Currently, we use the old semantic release pre/post commands to determine the package.json version. Then, that change is committed to master-dist for bower installs.~~

~~If the semantic release pre/post commands are no longer available, we need a new way to do that. The trick is that all files must be committed before the semantic release publishes and tags the build.~~

package.json
```diff
@@ -104,9 +104,9 @@
     "start-patternfly": "grunt build-patternfly serve",
     "start-rcue": "grunt build-rcue serve",
     "grunt": "grunt",
-    "semantic-release": "semantic-release pre && npm publish && semantic-release post",
-    "semantic-release-post": "semantic-release post",
-    "semantic-release-pre": "semantic-release pre",
+    "semantic-release": "semantic-release prepare && npm publish && semantic-release publish",
```
dlabrecq: I did some testing with the available semantic-release commands. The prepare/publish args are not valid (as is), but --prepare and --publish are.

package.json
```diff
@@ -104,9 +104,9 @@
...
+    "semantic-release-publish": "semantic-release publish",
+    "semantic-release-prepare": "semantic-release prepare",
```
dlabrecq: We can delete these two lines as they are unused. Or, change the semantic-release target to use them (i.e., using npm-run-all)?

dlabrecq (changes requested):

package.json
AdamJ: I updated the script to use `semantic-release --prepare` and `semantic-release --publish`, but I can't run them locally without errors. Is that expected?

dlabrecq (approved):
Looks good to me.

jeff-phillips-18 (approved):

Issue #727, opened 2017-08-28
Consider forms of configuration for app specific behavior


benjaminapetersen:
Original discussion on [this PR](https://github.com/patternfly/patternfly/pull/717#discussion_r135292347), regarding features implemented to support specific apps.  

`src/js/patternfly-functions-vertical-nav.js` is using local storage in a specific way to save state across browser restarts.  This behavior would likely be strange or unexpected in a different app without the same needs.  

Two items worth addressing:
- [ ] provide the ability to set `localStorage` or `sessionStorage` for this component so that consumers have the option to opt in to long term persistence.
- [ ] Consider adding a block to the `Contributing` section for features that may be app specific, but ensure they are sufficiently configurable so that other consumers can opt in/opt out of unique features.

andresgalante:
@jeff-phillips-18 Can you please review this one when you have time please?

jeff-phillips-18:
Yeah, I think this is worth pursuing.

dabeng:
@dgutride  @jeff-phillips-18 I tried to append a [new option](https://github.com/dabeng/patternfly/commit/50e8a83238aa220e01d259182217e58013e816bf#diff-3a08537993f8159a012ac07ee723e26cR11) for method setupVerticalNavigation(). Here's the [demo page](https://rawgit.com/dabeng/patternfly/ls-option-dist/dist/tests/vertical-navigation-with-tertiary-pins-disable-localstorage.html). In this page, developers can use the following sentence to disable localStorage cache:
```js
$().setupVerticalNavigation(true, undefined, { rememberOpenState: false });
```

dgutride:
@dabeng  - I think the request was to make this configurable to be able to use a different form of storage (in addition to disabling storage altogether) - can you take another stab at it?

dabeng:
Thanks a lot @dgutride I have updated the [demo](https://rawgit.com/dabeng/patternfly/ls-option-dist/dist/tests/) page according to your advice. You can access the following links to check modifications:
<img width="559" alt="screen shot 2018-07-23 at 4 20 50 pm" src="https://user-images.githubusercontent.com/5689075/43065516-acebc8be-8e94-11e8-8166-8100dc9af694.png">

dgutride:
@dabeng  - this looks like the right direction - one change, please use full words for the option instead of L or S (so localStorage or sessionStorage) and then put up a PR.

dabeng:
[...]

PR #1109, opened 2018-07-26
feat(vertical-nav): Provide new option to PF vertical navigation

dabeng:
## Description
This PR is targeted to issue #727.

In this PR, we expose new options - rememberOpenState and storage to fulfill the demands mentioned on the issue #727.

## Changes
Setting rememberOpenState to false in the current vertical nav test pages.

Here's the [demo page](https://rawgit.com/dabeng/patternfly/ls-option-dist/dist/tests/). 

@dgutride @amarie401 Could you review this PR when you have time :blush:

### DIFF
--- a/src/js/patternfly-functions-vertical-nav.js
+++ b/src/js/patternfly-functions-vertical-nav.js
@@ -8,6 +8,8 @@
     var options = $.extend({
         hoverDelay: 500,
         hideDelay: 700,
+        rememberOpenState: true,
+        storage: 'localStorage',
       }, userOptions || {}),
 
       navElement = $('.nav-pf-vertical'),
@@ -16,6 +18,7 @@
       handleResize = true,
       explicitCollapse = false,
       subDesktop = false,
+      storageLocation = options.storage === 'sessionStorage' ? 'sessionStorage' : 'localStorage',
 
       inMobileState = function () {
         return bodyContentElement.hasClass('hidden-nav');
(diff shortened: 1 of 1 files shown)


jeff-phillips-18:
I believe we can leave the set of test pages as is as it only serves to confuse with so many options that are not really visually different. I would actually prefer all the existing vertical nav test pages use the no storage option so that they always start in the initial (non-collapsed) state.

dabeng:
Hi @jeff-phillips-18 . Do you mean that we should abandon `vertical-navigation-primary-only-no-cache.html` and `vertical-navigation-primary-only-sessionstorage.html`

jeff-phillips-18:
Yes. I don't feel those are necessary as test pages. However, it would be good to set `rememberOpenState` to false in the current vertical nav test pages.

jeff-phillips-18:
This should be a `feat` as it adds a feature.

dabeng:
Thanks a lot for your advice. @jeff-phillips-18  I have refactored the PR.

src/js/patternfly-functions-vertical-nav.js
```diff
@@ -8,6 +8,8 @@
     var options = $.extend({
         hoverDelay: 500,
         hideDelay: 700,
+        rememberOpenState: false,
```
jeff-phillips-18: This needs to be initialized to true so that it doesn't break existing uses.

jeff-phillips-18 (changes requested):

src/js/patternfly-functions-vertical-nav.js
dabeng: ok. done

jeff-phillips-18 (approved):
I would still like to see the test pages set the rememberOpenState option to false, but I'm OK with keeping them as is as well.

dgutride (approved):

PR #1128, opened 2018-08-16
feat(flexbox layout): Add flexbox layout w/ example pages

### LINKED ISSUE #512: Sidebar doesn't work properly with tertiary-nav

### LINKED ISSUE #1118: Create a test page that includes sidebar working w/ the vertical nav
Create a test page that includes sidebar working w/ the vertical nav. Related to #512. 

Side bar can be seen in this test page here: https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/form.html

Create a new test page using the above from page with vertical nav vs horizonal nav.

jeff-phillips-18:
## Description
Adds a flexbox layout. Adds test pages to use the flexbox layout with scrolled content and inner scrolled content. 
Fixes sidebar styling when used with vertical navigation, adds a test page for the same.

Fixes #1118
Fixes #512 

[vertical-navigation-flexbox-inner-content.html](https://rawgit.com/jeff-phillips-18/patternfly/flexbox-layout-dist/dist/tests/vertical-navigation-flexbox-inner-content.html)
[vertical-navigation-flexbox.html](https://rawgit.com/jeff-phillips-18/patternfly/flexbox-layout-dist/dist/tests/vertical-navigation-flexbox.html)
[vertical-navigation-sidebar.html](https://rawgit.com/jeff-phillips-18/patternfly/flexbox-layout-dist/dist/tests/vertical-navigation-with-sidebar.html)


skateman:
This is beautiful! So the thing is, on some pages our sidebar is resizable I think between `col-md-2` and `col-md-6`. We have the code that changes the sizes, but I don't see any `col` classes on the sidebar/inner divs. Is it somehow possible to combine the two things?

UPDATE: below `col-md-2` the sidebar just becomes hidden.

epwinchell:
@jeff-phillips-18 It looks great. The footer is under the whole area, which actually will free up some much needed footer real estate on some screens. 

@terezanovotna  See any issues?

epwinchell:
@jeff-phillips-18 It looks like the vert nav is conflicting with the content

screen shot 2018-08-16 at 1 13 59 pm

terezanovotna:
LGTM! Thank you again @jeff-phillips-18  for taking time and fixing our MIQ issue!

jeff-phillips-18:
@skateman is it possible for you to set the width rather than changing the class?

skateman:
@jeff-phillips-18 well, yes, but then we would have to implement a drag and drop resizer which is a lot of work. @epwinchell @dclarizio how bad would be to drop the resizable sidebar from the explorer screens?

epwinchell:
> @epwinchell @dclarizio how bad would be to drop the resizable sidebar from the explorer screens?

@skateman I have no objections to dropping it. The idea was to mimic the slider in our old DHTMLX layouts - and alleviate the wrapping in some trees, but it's created other problems - like not being properly responsive/tablet-friendly, relying on hacky code, and not being inline with Patternfly.

skateman:
I'm still a little worried about this being responsive, the content of the sidebar on different page width can look differently. @jeff-phillips-18 can at least the flexbox somehow mimic the one width as it would be a col-md-3 for example?

mcoker:
@jeff-phillips-18 the CSS looks great to me. I noticed a few visual things, not sure if they're within the scope of this PR or should be in a separate issue.

on [https://rawgit.com/jeff-phillips-18/patternfly/flexbox-layout-dist/dist/tests/vertical-navigation-flexbox-inner-content.html](https://rawgit.com/jeff-phillips-18/patternfly/flexbox-layout-dist/dist/tests/vertical-navigation-flexbox-inner-content.html)
* there is a background color on `.sidebar-pf` @ `min-width: 768px` that creates a `#fafafa` background to the expanded content in the accordion. There is no background < 768px, so the background matches the background of the container it's in.
* the top/bottom borders of the accordion would probably look better removed, so that all you see there is the top toolbar and bottom footer borders there.
   * same thing on [https://rawgit.com/jeff-phillips-18/patternfly/flexbox-layout-dist/dist/tests/vertical-navigation-with-sidebar.html](https://rawgit.com/jeff-phillips-18/patternfly/flexbox-layout-dist/dist/tests/vertical-navigation-with-sidebar.html)
* < 768px, the pagination is misplaced
screen shot 2018-08-17 at 9 40 55 am

jeff-phillips-18:
@skateman You could set the width to percentages to mimic bootstrap column widths (ie 25% at min-width: @screen-md-min)

jeff-phillips-18:
@michael-coker Updated per your comments.

dclarizio:
> @epwinchell @dclarizio how bad would be to drop the resizable sidebar from the explorer screens?

I think that would require some input from @bascar and perhaps @terezanovotna as this will affect the user experience.  Without the ability to see deeper tree nodes (I know we've tried to remove most of these, but think "Automate") and/or using ellipses (...) to truncate text, this could cause some usability issues.

Perhaps some example screenshots of where removing the adjuster will negatively affect the trees are in order for us to truly judge the impact.

jeff-phillips-18:
The implementation for the resizable slider and that feature is out of scope for this PR. I don't believe this PR hinders the future implementation of the resizable slider, so I'd like to see us proceed with this as is and add any enhancements/features in follow up / future PRs.

LHinson:
I agree with this approach @jeff-phillips-18

tests/pages/_layouts/layout-flexbox.html
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE html>
+<!--[if IE 9]><html lang="en-us" class="ie9 layout-pf layout-pf-fixed"><![endif]-->
+<!--[if gt IE 9]><!-->
+<html lang="en-us" class="layout-flexbox-pf">
```
skateman, line 4: Is it really necessary to add CSS classes to the top HTML tag? We're already struggling with this as we need to mantain multiple top-level template files :disappointed: cc @himdel

jeff-phillips-18: which layout are you currently using? what classes are set on `html`, `body`, down to your inner content container?

skateman: I just simply don't understand why do you need a CSS class on the `<html>` tag, it makes top-level template handling complicated.

himdel: This is just a test, right? So... this doesn't mean anything for us.

himdel: That said, if the layout requires the class on html, then we probably can't use it at all @jeff-phillips-18 .

So please make sure no classes are *required* on html or body.

skateman: I could make peace with `<body>` if really necessary, but I don't see the reason for being so super-global to have classes on `<html>`.

jeff-phillips-18: In this implementation, `<html>` needs to have `height: 100%`. I'll look for another way.

skateman: Not the body? I think we have the `100%` globally btw...

jeff-phillips-18: I have a fix so it doesn't require the class on `html`

jeff-phillips-18:
@himdel @skateman Have you taken a look at the latest update? Does this meet your needs?

skateman:
@jeff-phillips-18 we're still trying to integrate this into our codebase, let's leave this PR open until we're able to use it without issues.

jeff-phillips-18:
Pretty stale and no further requests for this functionality, closing.

PR #1129, opened 2018-08-16
Login Basic Alert page has been created and aligns with styles from design specs  

### LINKED ISSUE #1125: questions about Login screen design and implementation
A couple of questions came up during implementation of the [Login screen for pf-react](https://github.com/patternfly/patternfly-react/pull/362).

1. The [design documentation](https://www.patternfly.org/pattern-library/application-framework/login-page/#design) shows a divider during the responsive state after the Log In button, but this isn't present in the [core example](https://rawgit.com/patternfly/patternfly/master-dist/dist/tests/login-social-account-two-column.html). If it isn't intended to be there, then an issue should be opened for both patternfly-design and patternfly-react to remove the divider.
2. The [design documentation](https://www.patternfly.org/pattern-library/application-framework/login-page/#design) shows a warning message at the top, with spacing above the message, but there isn't an example of this in the core implementation. Guidance is needed on how to handle that spacing (e.g. do we want to add classes and css in patternfly core for that, or address it only in pf-react). This also will require an issue in pf-react based on how we want to move forward. 
3. The css defines a background image that seems very specific to PatternFly and not useful for any project that would use this pattern. Should the background image be removed? The react implementation was altered to allow the consumer to specify their own background image. 
4. The test pages in core include 100px padding at the bottom of the page. If this spacing is expected in the react implementation, then it should be added to the core css.
5. The design documentation includes inline messages directly above the form. Examples for these messages are needed in core so that we can later add them to the react implementation. 

When these items are resolved, any changes needed in the react implementation will need a new issue to capture those updates.

christiemolloy:
fix #1125

## Description

Needed to fix the log in screen implementation (styles to align with the login alert page). 

## Changes

* Created login-basic-alert.html
* added alert component 
* 100px padding-bottom added to the footer
* text-danger added below form inputs with red color 
*padding added to the top of the page

## PR checklist (if relevant)


matthewcarleton:
Thanks @christiemolloy this is looking good. We can take a look at the error. It's likely from your json file.

tests/pages/_includes/widgets/framework/login-basic-alert.html
```diff
@@ -0,0 +1,71 @@
+<div class="login-pf-page">
+
+<div class="container-fluid">
+    <div class="row">
+        <div class="col-sm-8 col-sm-offset-2 col-md-8 col-md-offset-2 col-lg-8 col-lg-offset-2">
+            {% include widgets/communication/alert-danger.html %}
+        </div>  
+    </div>
+</div>
+
```
jeff-phillips-18: Might be better to add an variable to login-basic.html so as not to repeat all that content. Something like
```
<div class="login-pf-page">

{% if page.login-alert %}
<div class="container-fluid">
    <div class="row">
        <div class="col-sm-8 col-sm-offset-2 col-md-8 col-md-offset-2 col-lg-8 col-lg-offset-2">
            {% include widgets/communication/alert-danger.html %}
        </div>  
    </div>
</div>
{% endif %}

...
```

Then in the header of login-basic-alert.html:
```
login-alert: true
```

src/less/login.less
```diff
@@ -105,6 +105,9 @@
 // The following styles are for the new login
 // --------------------------------------------------
 .login-pf-page {
+  padding-top: @login-pf-page-padding-top;
+  // padding-bottom: @login-pf-page-padding-bottom;
```
jeff-phillips-18: remove commented out code

src/less/login.less
```diff
@@ -158,7 +161,7 @@
     }
     display:flex;
     flex-direction: column;
-    h1, p {
+    h1 {
```
jeff-phillips-18, line 163: why remove the `p` ? Any risk of breaking existing uses?

jeff-phillips-18 (changes requested):

src/less/login.less
matthewcarleton: hey @jeff-phillips-18 having all the paragraphs centre aligned isn't flexible enough based on the design. There is a helper class in Bootstrap that can be used for text that needs to be centre aligned.

jeff-phillips-18: OK, thanks.

tests/pages/_includes/widgets/framework/login-basic-alert.html
christiemolloy: okay sounds good. will update

src/less/login.less
christiemolloy: yes

christiemolloy:
changes fixed, @jeff-phillips-18 @matthewcarleton

matthewcarleton (dismissed):
This looks good, thanks!

tests/pages/login-basic.html
```diff
@@ -7,4 +7,5 @@
 full-page: true
 url-js-extra: '//cdnjs.cloudflare.com/ajax/libs/bootstrap-select/1.12.2/js/bootstrap-select.min.js'
 ---
+
```
jeff-phillips-18: no need to modify this file

jeff-phillips-18 (dismissed):
LGTM, just the one nit. Thanks for updating!

jeff-phillips-18 (approved):

PR #1133, opened 2018-08-22
feat(blank slate): Add content variant for Blank Slate

### LINKED ISSUE #1079: Remove Background Color & Border from Empty State
Per a discussion in the uxd-visual Slack channel with @serenamarie125 and @jennyhaines, the empty state pattern should not have a set background color or border. As such, the background should be set as transparent and the border removed.

http://www.patternfly.org/pattern-library/communication/empty-state/#code

The design document needs to be updated as well, as that shows a background with color and a border. Issue created here for that update: https://github.com/patternfly/patternfly-design/issues/682

jeff-phillips-18:
This variant:
- makes the background transparent
- removes the padding
- sets a 20px margin around the component
- removes the borders.

Fixes #1079

https://rawgit.com/jeff-phillips-18/patternfly/empty-state2-dist/dist/tests/blank-slate.html

### DIFF
--- a/tests/pages/blank-slate.html
+++ b/tests/pages/blank-slate.html
@@ -5,7 +5,7 @@ title: Blank Slate
 resource: true
 ---
 
-<h2>Empty Blank Slate</h2>
+<h2>Blank Slate</h2>
 <div class="blank-slate-pf">
   <h1>Empty State Title</h1>
   <p>Lorem ipsum dolor sit amet, consectetur adipisicing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.</p>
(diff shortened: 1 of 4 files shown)


tests/pages/blank-slate.html
```diff
@@ -5,7 +5,7 @@
 resource: true
 ---
 
-<h2>Empty Blank Slate</h2>
+<h2>Blank Slate</h2>
```
AdamJ, line 8: The pattern has this defined as "Empty State". Should we leave it that for consistency, or change the name of the pattern & navigation on PatternFly.org?

jeff-phillips-18: I'm ok either way. It was just inconsistent with it's page name so I made it consistent with itself.

AdamJ: Oy - it looks like nothing matched, so I'm good with updating it as you have proposed. :-)

AdamJ (approved):

dgutride (approved):

PR #1152, opened 2018-11-07
feat(loading-state): Add the loading-state component

jeff-phillips-18:
Adds the loading-state component as defined on [patternfly.org]( https://www.patternfly.org/pattern-library/communication/loading-state)

[rawgit examples link](https://rawgit.com/jeff-phillips-18/patternfly/loading-state-dist/dist/tests/loading-state.html)


src/sass/converted/patternfly/_loading-state.scss
```diff
@@ -0,0 +1,24 @@
+.loading-state-pf {
+  font-size: ($spinner-size * .5);
+  height: 100%;
+
+  .loading-state-pf-spacer {
```
mcoker: I think you can get rid of this element all together and do

```
.loading-state-pf&::before {
    content: "";
    height: 25%;
    display: block;
}
```

src/sass/converted/patternfly/_loading-state.scss
```diff
@@ -0,0 +1,24 @@
+.loading-state-pf {
+  font-size: ($spinner-size * .5);
+  height: 100%;
```
mcoker, line 6: if you use `display: flex; flex-direction: column; align-items: center;` you can get rid of the `.loading-state-pf-text {` block.

mcoker (dismissed):
Hey Jeff, looks great 👍I left a couple of comments if you want to review them, but what you have works fine, too.

jeff-phillips-18:
Thanks @mcoker, updated to your suggestions. Much better.

PR #1151, opened 2018-10-30
feat(icon): Add support for fontawesome v5

skateman:
## Description
Allow the usage of fontawesome 5 icons in PF3.

## Changes
Added the new FA prefixes into the codebase.
https://fontawesome.com/how-to-use/on-the-web/setup/upgrading-from-version-4

PR #1155, opened 2018-11-13
fix(Icon): Add missing converted files for FA v5 support

jeff-phillips-18:
Adding the converted SASS files from changes made in #1151


mcoker (approved):
👍

PR #1174, opened 2019-05-15
build(package.json): Upgrade jquery to 3.4.0 to address downstream CVE alert

mturley:
## Description

In ManageIQ, we are seeing an alert in GitHub for a CVE describing a vulnerability in jquery versions below 3.4.0: https://access.redhat.com/security/cve/cve-2019-11358

Because we depend on patternfly-react, which depends on patternfly, we're depending indirectly on jquery@3.2.1 which is causing this alert. This PR upgrades the jquery version in patternfly, so that we can upgrade it in patternfly-react by upgrading patternfly, so that we can upgrade it in ManageIQ repos by upgrading patternfly-react. ⛓ 

According to https://blog.jquery.com/2019/04/10/jquery-3-4-0-released/, this is a safe upgrade:
> There should be no compatibility issues if upgrading from jQuery 3.0+.

## Changes

* Upgrades the `jquery` dependency in package.json from `~3.2.1` to `~3.4.0`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "dependencies": {
     "bootstrap": "~3.4.0",
     "font-awesome": "^4.7.0",
-    "jquery": "~3.2.1"
+    "jquery": "~3.4.0"
   },
   "devDependencies": {
     "@commitlint/cli": "^3.2.0",


tuxgasy (approved):

droideck:
Looks like the issue is [here](https://github.com/patternfly/patternfly/blob/73bfd1083657148ef38db5ca3b82b13814f4e1a9/tests/unit/bootstrap-switch.spec.js#L31)

[jQuery 3.4.0 changelog](https://blog.jquery.com/2019/04/10/jquery-3-4-0-released/) has information that `:first` was depricated. `.first();` should be used now.
Also, the issue described in `Radio elements: expected state in event handlers` can be related to the failure too...
Hope it helps.

ddelabru:
> [jQuery 3.4.0 changelog](https://blog.jquery.com/2019/04/10/jquery-3-4-0-released/) has information that `:first` was depricated. `.first();` should be used now.

Seems like changing that alone isn't enough to get the unit test to pass; I tried tweaking that with: https://github.com/patternfly/patternfly/pull/1178

Same test is failing on my own unrelated PR: https://github.com/patternfly/patternfly/pull/1177

I'd like to work on fixing this unit test but when I run `npm test` locally *all* of the unit tests fail, and that makes it a more daunting problem; it might be easier if I could use a container to replicate the environment Travis CI is using and isolate what's making this particular test fail there.

droideck:
The issue is fixed in #1180
Probably, this one can be closed.

mturley:
Thanks @droideck. Closing.

PR #1178, opened 2019-06-12
Fix deprecated use of :first in unit test

ddelabru:
## Description
Recent pull requests on Patternfly are failing because of the same unit test, and a comment suggests this may be due to a deprecated jQuery selector in one of the tests: https://github.com/patternfly/patternfly/pull/1174#issuecomment-497004617

When I run unit tests locally, they *all* fail, so I'm opening a pull request to see removing the `:first` selector from that test makes it pass in Travis CI.

## Changes

* Replace uses of the deprecated `:first` selector in a specific unit test


ddelabru:
Alright, so it looks like the answer is no, this does not fix the affected unit test. Closing my pull request.

PR #1184, opened 2019-08-14
chore(build): add package-lock.json

### LINKED ISSUE #1021: Lock Down Dependencies
In order to lock down the dependencies of PatternFly, we should be using a lock file of sorts. The recommended method for those that publish packages is to use `npm-shrinkwrap.json`.

This should prevent the issues where dependent dependencies cause build failures, even when packages were not updated by PatternFly.

https://docs.npmjs.com/files/shrinkwrap.json

### LINKED ISSUE #1185: Update PatternFly dependencies
There are several security alerts showing up on PatternFly right now - we need to update insecure dependencies and perform a new release.  Can be done in conjunction with adding the lock files #1021

redallen:
Closes #1021
Closes #1185


dgutride (approved):

mcoker (approved):
👍

PR #1188, opened 2019-09-17
Add main field in package.json for easier import

fbastien:
Added the missing "main" field in package.json for easier import/require by other packages.

## Description
When trying to import patternfly in my project using Webpack Encore, the `require('patternfly')` instruction gave me a dependency not found error.
I figured out this is because patternfly's package.json has no "main" field, so I had to dig into the package folders to get the full path and then write `require('patternfly/dist/js/patternfly.js')` instead.

This PR is about fixing this, so that it's possible to import/require patternfly the first and better way.

## Changes

* Added "main" field in package.json, pointing to `dist/js/patternfly.js`

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "0.0.0-semantically-released",
   "author": "Red Hat",
   "license": "MIT",
+  "main": "dist/js/patternfly.js",
   "dependencies": {
     "bootstrap": "~3.4.1",
     "font-awesome": "^4.7.0",


fbastien:
Sorry, I should have opened an issue first, I realized pointing `main` to `js/patternfly.js` is not sufficient, I had to import the following files as well (as in `bower.json`):

```javascript
require('patternfly/dist/js/patternfly.js');
require('patternfly/dist/css/patternfly.css');
require('patternfly/dist/css/patternfly-additions.css');
require('patternfly/dist/css/rcue.css');
require('patternfly/dist/css/rcue-additions.css');
```

I don't know how to fix this since, contrary to Bower, [NPM allow only a single file in the "main" field](https://docs.npmjs.com/files/package.json#main). Is it possible to create a wrapper (e.g. `dist/js/main.js`) that would import other JS and CSS files?

PR #1189, opened 2019-09-18
Added package.json to list of the file in the gemspec explicitly

majioa:
## Description
add package.json to list of the files in the gemspec explicitly, because it is required by version.rb


jeff-phillips-18 (dismissed):

jeff-phillips-18:
@majioa Could you please re-word your commit to the semantic release format: 
```
fix(patternfly-sass): Added package.json to list of the file in the gemspec explicitly
```

majioa:
done

